PR #58, opened 2020-06-11
Fix Channel Editing permissions

AMorporkian:
Simple change to make edit_channel() work properly.


Skillz4Killz:
@AMorporkian Can you tell me what was wrong making it not edit properly? IMO the Overwrite instead of RawOverwrite is a lot more user friendler and makes it easier to edit channels by passing strings instead of calculating the permission bit

AMorporkian:
The use case that prompted the change was copying the permission overrides from one channel and applying it to another. 
```ts
    const mentionedChannel = guild?.channels.get(channelId)
    const perm = guild?.channels.get("#############")?.permission_overwrites
    if (mentionedChannel && perm?.length) {
      const newOptions = {
        permission_overwrites: perm,
        parent_id: "#############",
      }
      editChannel(mentionedChannel, newOptions)
```

Editing a channel's permissions requires a permission bitset for `allow` and `deny`, necessitating the `Raw_Overwrite[]` in `ChannelEditOptions`.  

We could modify `edit_channel()` to calculate the bitmasks from the `Overwrite[]` in `ChannelEditOptions`, but I am fairly new to TypeScript and couldn't figure out a good way to reverse the Enum and get a bit set back.

Skillz4Killz:
I just deployed a small change to the master branch. Can you try switching to master on your test bot and see if this solution works. I basically just made it accept an array of Permission[] and convert it to a bit when sending. Note that you may need to change
`    const perm = guild?.channels.get("#############")?.permission_overwrites` to 
`    const perm = guild?.channels.get("#############")?.permissions`

PR #68, opened 2020-07-16
Alternative sharding solution for smaller bots

Skillz4Killz:
This PR is a major feature enhancement without implementing any breaking changes but due to it's massive effect I want to give it some time for other devs to review and test.

**Understanding**
Until now Discordeno had only 1 sharding solution that was designed for large bots. This solution used the Deno Workers(which are still unstable) but it allowed insanely huge bots to work as well. However, for smaller bots, this increased a huge amount of complexity. Dealing with workers and different threads is not an easy concept so to solve this I created a second alternative sharding for now being called `basicShard`. 

**Effects**

If your bot has less than 25 shards aka less than 25,000 servers you will be automatically included in the basic shard system. This means you no longer use Deno workers which gives added benefits. 
a. Removed complexity
b. No more need to have `--unstable` flag when running the bot
c. Better errors. Deno workers hide and remove a majority of the errors so this means if there are errors, it will be a lot cleaner and helpful. As an example, some of you may have noticed these silly errors that are impossible to debug:
```ts
error: Uncaught Error: Uncaught ConnectionReset: Socket has already been closed
    at WorkerImpl.#poll ($deno$/web/workers.ts:210:17)
```
Once Deno workers are removed out of the equation it will no longer swallow the errors allowing us a much better dev experience.
d. Memory improvements

**TODO:**

This needs to be heavily tested before merging. I expect quite a few functionalities may have been lost as I may have forgotten to bring it over but after testing is done, I am certain this will work as intended with full functionality.

Update docs about this.

### DIFF
--- a/module/basicShard.ts
+++ b/module/basicShard.ts
@@ -1,0 +1,203 @@
+import {
+  DiscordBotGatewayData,
+  GatewayOpcode,
+  ReadyPayload,
+} from "../types/discord.ts";
+import {
+  eventHandlers,
+  botGatewayData,
+  IdentifyPayload,
+} from "./client.ts";
+import { delay } from "https://deno.land/std@0.61.0/async/delay.ts";
+import {
+  connectWebSocket,
+  isWebSocketCloseEvent,
+  WebSocket,
+} from "https://deno.land/std@0.61.0/ws/mod.ts";
+import { DiscordHeartbeatPayload } from "../types/discord.ts";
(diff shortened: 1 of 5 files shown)


Skillz4Killz:
Just tested this locally with two different bots. 1 small bot with 17 servers and one with 7K servers. Even fetched 2 million members. Looks good to merge, will update docs in the next batch of docs updates.

PR #79, opened 2020-07-30
Add oldPresence to presenceUpdate event

Androz2091:
This is useful to compare the presence payload before the update and the new presence payload. Here is a simple use-case below:
```ts
import Client, { botID } from "../Discordeno/module/client.ts";
import { sendMessage } from "../Discordeno/handlers/channel.ts";
import { Intents } from "../Discordeno/types/options.ts";
import config from "./config.ts";

Client({
    token: config.token,
    intents: [Intents.GUILD_MESSAGES, Intents.GUILDS, Intents.GUILD_PRESENCES],
    eventHandlers: {
        ready: () => {
            console.log(`Logged!`);
        },
        messageCreate: (message) => {
            if (message.content === "!ping") {
                sendMessage(message.channel, "Pong");
            }
        },
        presenceUpdate: (oldPresence, newPresence) => {
            if (!oldPresence) return;
            if (oldPresence.status !== 'offline' && newPresence.status === 'offline') {
                console.log(`User ${newPresence.user.id} is now offline!`);
            } else if(oldPresence.status === 'offline' && newPresence.status !== 'offline') {
                console.log(`User ${newPresence.user.id} is now online!`);
            }
        }
    }
});
```

It seems me logical to have the `oldPresence` parameter before the `newPresence` one ([as Discord.js does](https://github.com/discordjs/discord.js/blob/2338594030ff8839e36406ee979cc5f65253d6ca/src/client/actions/PresenceUpdate.js#L39)), but it would mean that this PR includes a breaking change, so maybe it would be better to keep the new payload before, and add the `oldPresence` after.

### DIFF
--- a/module/shardingManager.ts
+++ b/module/shardingManager.ts
@@ -566,7 +566,7 @@ export async function handleDiscordPayload(
         const payload = data.d as PresenceUpdatePayload;
         const oldPresence = cache.presences.get(payload.user.id);
         cache.presences.set(payload.user.id, payload);
-        return eventHandlers.presenceUpdate?.(oldPresence, payload);
+        return eventHandlers.presenceUpdate?.(payload, oldPresence);
       }
 
       if (data.t === "TYPING_START") {
--- a/module/shardingManager.ts
+++ b/module/shardingManager.ts
@@ -563,7 +563,10 @@ export async function handleDiscordPayload(
       }
 
       if (data.t === "PRESENCE_UPDATE") {
-        return eventHandlers.presenceUpdate?.(data.d as PresenceUpdatePayload);
+        const payload = data.d as PresenceUpdatePayload;
+        const oldPresence = cache.presences.get(payload.user.id) ?? null;
+        cache.presences.set(payload.user.id, payload);
+        return eventHandlers.presenceUpdate?.(oldPresence, payload);
       }
 
       if (data.t === "TYPING_START") {
--- a/module/shardingManager.ts
+++ b/module/shardingManager.ts
@@ -564,7 +564,7 @@ export async function handleDiscordPayload(
 
       if (data.t === "PRESENCE_UPDATE") {
         const payload = data.d as PresenceUpdatePayload;
-        const oldPresence = cache.presences.get(payload.user.id) ?? null;
+        const oldPresence = cache.presences.get(payload.user.id);
         cache.presences.set(payload.user.id, payload);
         return eventHandlers.presenceUpdate?.(oldPresence, payload);
       }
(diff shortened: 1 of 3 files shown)


module/shardingManager.ts
```diff
@@ -563,7 +563,10 @@ export async function handleDiscordPayload(
       }
 
       if (data.t === "PRESENCE_UPDATE") {
-        return eventHandlers.presenceUpdate?.(data.d as PresenceUpdatePayload);
+        const payload = data.d as PresenceUpdatePayload;
+        const oldPresence = cache.presences.get(payload.user.id) ?? null;
```
Skillz4Killz, line 567:
```suggestion
        const oldPresence = cache.presences.get(payload.user.id);
```
Forcing a `null` here is unnecessary and TS can handle undefined just as nicely and it'll make typings easier as well. Also as consistency measure the rest of the lib avoids null whenever possible as well for `undefined` except the in the cases where `null` is necessary.

module/shardingManager.ts
```diff
@@ -563,7 +563,10 @@ export async function handleDiscordPayload(
...
+        cache.presences.set(payload.user.id, payload);
+        return eventHandlers.presenceUpdate?.(oldPresence, payload);
```
Skillz4Killz, line 569:
```suggestion
        return eventHandlers.presenceUpdate?.(payload, oldPresence);
```
Can we reverse this and make payload first. Consistency wise with the rest of the lib like memberUpdate, roleUpdate etc all the update events have the latest data first and the cached data as last.

Example:
`guildMemberUpdate?.(guild, member, cachedMember);`

types/options.ts
```diff
@@ -105,7 +105,10 @@ export interface EventHandlers {
     nickname: string,
     oldNickname?: string,
   ) => unknown;
-  presenceUpdate?: (data: PresenceUpdatePayload) => unknown;
+  presenceUpdate?: (
+    oldPresence: PresenceUpdatePayload | null,
```
Skillz4Killz, line 109:
```suggestion
    presence: PresenceUpdatePayload,
```

Based on my suggestion above to be consistant with the rest of the lib its best to swap these around please.

types/options.ts
```diff
@@ -105,7 +105,10 @@ export interface EventHandlers {
...
+    newPresence: PresenceUpdatePayload,
```
Skillz4Killz, line 110:
```suggestion
    oldPresence?: PresenceUpdatePayload,
```

Skillz4Killz:
I've gone ahead and made the small nitpick changes I requested and merged this in. Once you reload the cache for v7 you will be able to use this! Thank you so much for the PR <3

PR #103, opened 2020-09-21
Use ClientOptions#properties option

chroventer:
The option ClientOptions#properties was completely disregarded previously. However, it is now taken into account:
```ts
createClient({
    ...,
    properties: {
        $os: "...",
        $browser: "...",
        $device: "..."
    }
});
```

*Did not format the entire source folder since that would be irrelevant to this PR.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,5 @@
 {
   "deno.enable": true,
-  "editor.formatOnSave": true
+  "editor.formatOnSave": true,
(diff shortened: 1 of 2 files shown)


Skillz4Killz:
@chroventer Can u give me a use case for this? When would you need this? TBH I don't think this belongs in the client options.

chroventer:
> @chroventer Can u give me a use case for this? When would you need this? TBH I don't think this belongs in the client options.

We can have a nested option in the ClientOptions named "websocketOptions".

PR #114, opened 2020-09-29
refactor: unnecessary Collection#get()

chroventer:
Since `Collection` extends `Map` and `Map` is an iterable, `Collection#get()` to get the value from the collection is unnecessary.

### DIFF
--- a/src/utils/collection.ts
+++ b/src/utils/collection.ts
@@ -47,16 +47,15 @@ export class Collection<K, V> extends Map<K, V> {
 
   map<T>(callback: (value: V, key: K) => T) {
     const results = [];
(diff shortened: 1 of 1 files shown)


Skillz4Killz:
Although this PR can clean up the libraries code greatly, it comes at other costs. The memory optimizations are removed by creating all these extra arrays `[key, value]`. Furthermore the performance of entries vs keys is vastly different meaning the loops will be a lot slower. 

`.get()` is extremely fast at about 90million ops per second. So there isn't much concern of performance. Although the code is a bit extra using this method it is much better in terms of performance and memory usage to use .keys and .get

chroventer:
Lol

PR #124, opened 2020-10-01
improve perf of zlib inflating

devcat:
## Description
replaced javascript implementation of zlib with webassembly

image


Skillz4Killz:
@devcat LGTM! Is there any reason why your package is stil pre-1.0

Ideally, it would be nice if this was stable for production use.

PR #135, opened 2020-10-11
Update discord.ts

Suyashtnt:
very minor

Update a comment in `discord.ts`
**Status**

  - [x] This PR includes breaking changes (methods removed or renamed, parameters moved or removed)
- [] This PR **only** includes non-code changes, like changes to documentation, README, etc.(I guess it counts?)

### DIFF
--- a/src/constants/discord.ts
+++ b/src/constants/discord.ts
@@ -1,5 +1,5 @@
 export const baseEndpoints = {
-  BASE_URL: `https://discord.com/api/V8`,
+  BASE_URL: `https://discord.com/api/v8`,
   CDN_URL: "https://cdn.discordapp.com",
 };
 
--- a/src/constants/discord.ts
+++ b/src/constants/discord.ts
@@ -1,5 +1,5 @@
 export const baseEndpoints = {
-  BASE_URL: `https://discord.com/api/v8`,
+  BASE_URL: "https://discord.com/api/v8",
   CDN_URL: "https://cdn.discordapp.com",
 };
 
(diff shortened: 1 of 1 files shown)


Skillz4Killz:
Hmmm, i think we could possibly just remove this entirely, the API version being changed to used new apis will no longer work. Discord is stopping that and we will not be able to use v9 before its officially released. I have to think about this

Skillz4Killz:
@Suyashtnt I've rebased this to the `next` branch so it can include breaking changes which we merge in the next major release. I believe the best way to handle this is to just remove this entire variable. Discord is no longer going to allow us to customize which version we can use. After the v7 headaches that other libs caused Discord is preventing libraries from using other versions that are not released which makes this variable pretty useless.

1. Remove the API_VERSION variable
2. Hardcode the v8 into the URL

Suyashtnt:
Done, changed accoding to that

src/constants/discord.ts
```diff
@@ -1,15 +1,8 @@
-let API_VERSION = "v8";
-
 export const baseEndpoints = {
-  /** Although, the version can be defaulted, keep the v6 as it can be changed to test newer versions when necessary. */
-  BASE_URL: `https://discord.com/api/${API_VERSION}`,
+  BASE_URL: `https://discord.com/api/V8`,
```
Skillz4Killz, line 2:
```suggestion
  BASE_URL: `https://discord.com/api/v8`,
```

Skillz4Killz (approved):

PR #152, opened 2020-10-28
rework permissions checking

Skillz4Killz:


src/utils/permissions.ts
```diff
@@ -2,6 +2,7 @@ import { cacheHandlers } from "../controllers/cache.ts";
 import { botID } from "../module/client.ts";
 import type { Guild } from "../structures/guild.ts";
 import type { Role } from "../structures/role.ts";
+import { RawOverwrite } from "../types/guild.ts";
```
ayntee, line 5: Since `RawOverwrite` is not used as a value, it should be imported using `import type` syntax.

src/utils/permissions.ts
```diff
@@ -2,6 +2,7 @@ import { cacheHandlers } from "../controllers/cache.ts";
...
 import type { Permission } from "../types/permission.ts";
 import { Permissions } from "../types/permission.ts";
```
ayntee, line 7: Since `Permissions` is not used as a value, it should be imported using `import type` syntax.

ayntee (changes requested):

src/utils/permissions.ts
```diff
@@ -104,87 +105,77 @@ export async function hasChannelPermissions(
...
-  );
-
-  const rolesOverwrites = channel.permission_overwrites?.filter((o) =>
-    member.roles.includes(o.id)
-  );
-
-  const everyoneOverwrite = channel.permission_overwrites?.find((o) =>
-    o.id === guild.id
-  );
+  let memberOverwrite: RawOverwrite | undefined = undefined;
```
ayntee, line 108: Variables are initialized with the value `undefined` if a value is not assigned.
```ts
let memberOverwrite: RawOverwrite | undefined;
```

src/utils/permissions.ts
```diff
@@ -104,87 +105,77 @@ export async function hasChannelPermissions(
...
+  let everyoneOverwrite: RawOverwrite | undefined = undefined;
```
ayntee, line 109: Variables are initialized with the value `undefined` if not a value is not assigned.
```ts
let everyoneOverwrite: RawOverwrite | undefined; 
```

ayntee (changes requested):

ayntee (approved):
:tada: LGTM

PR #163, opened 2020-11-01
Better docz

ayntee:
## Changes

- Switch to [`docz`](https://docz.site) from `gatsby-gitbook-starter`.
- Add workflow for docs deployment.
- Replace "bad" routes with meaningful routes.

Draft PR because of  doczjs/docz#1582

## Screenshots

dark mode

light mode

search bar

code block

preview

image

image


Skillz4Killz:
Can you try and run this locally and see how this is going to look? Just as an example. It's hard to decide if this is ready to merge if we can't see the final version.

Note: While your doing this maybe you can look into seeing if it is possible to host this through github pages. That wasmy biggest limitation last time and I would ideally like docs to be on github pages.

ayntee:
> Can you try and run this locally and see how this is going to look? Just as an example. It's hard to decide if this is ready to merge if we can't see the final version.

I will add some screenshots of the generated website. Also, you can fully-customize the theme if you do not like the current style.

> Note: While your doing this maybe you can look into seeing if it is possible to host this through github pages. That wasmy biggest limitation last time and I would ideally like docs to be on github pages.

And yes, it does support GitHub Pages.

PR #166, opened 2020-11-03
Lint code

ayntee:


src/controllers/cache.ts
```diff
@@ -79,6 +79,7 @@ function forEach(
   return cache[table].forEach(callback);
 }
 
+// deno-lint-ignore prefer-const
 export let cacheHandlers = {
```
Skillz4Killz, line 83:
```suggestion
export const cacheHandlers = {
```

Skillz4Killz: We can just make this a const no?

src/handlers/channel.ts
```diff
@@ -325,9 +325,13 @@ export function editChannel(
 
   const payload = {
     ...options,
+    // deno-lint-ignore camelcase
```
Skillz4Killz, line 328: I think if we move this payload entirely into the function call below and not in a variable deno lint will not have errors on this. It will also slightly improve our performance as we don't need to declare this const

src/structures/channel.ts
```diff
@@ -49,4 +49,4 @@ export async function createChannel(
   return channel;
 }
 
-export interface Channel extends Unpromise<ReturnType<typeof createChannel>> {}
+export type Channel = Unpromise<ReturnType<typeof createChannel>>;
```
Skillz4Killz, line 52: Never use `type` for these. Interfaces are so much better. Like no circular import errors with interfaces.

Skillz4Killz: I believe a simple deno lint ignore here is better than having this be types

src/controllers/cache.ts
ayntee: @Skillz4Killz in case, the users need to modify the value.

PR #167, opened 2020-11-03
Add Role.tags property

ayntee:
Reference: <https://github.com/discord/discord-api-docs/issues/1537>

**Status**

### DIFF
--- a/src/structures/role.ts
+++ b/src/structures/role.ts
@@ -6,6 +6,11 @@ export async function createRole(data: RoleData) {
     ...data,
     /** The @ mention of the role in a string. */
     mention: `<@&${data.id}>`,
+    tags: {
+      botID: data.tags?.botID,
+      premiumSubscriber: "premium_subscriber" in (data.tags ?? {}),
+      integrationID: data.tags?.integrationID,
+    },
   };
 }
 
--- a/src/structures/role.ts
+++ b/src/structures/role.ts
@@ -6,7 +6,7 @@ export async function createRole(data: RoleData) {
 
   const roleTags = {
     botID: tags?.bot_id,
-    premiumSubscriber: Boolean(data.tags?.premium_subscriber),
+    premiumSubscriber: "premium_subscriber" in (data.tags ?? {}),
     integrationID: data.tags?.integration_id,
   };
 
--- a/src/structures/role.ts
+++ b/src/structures/role.ts
@@ -6,8 +6,8 @@ export async function createRole(data: RoleData) {
 
   const roleTags = {
     botID: tags?.bot_id,
-    premiumSubscriber: "premium_subscriber" in (data.tags ?? {}),
-    integrationID: data.tags?.integration_id,
+    premiumSubscriber: "premium_subscriber" in (tags ?? {}),
+    integrationID: tags?.integration_id,
   };
 
   return {
--- a/src/structures/role.ts
+++ b/src/structures/role.ts
@@ -2,15 +2,19 @@ import { Unpromise } from "../types/misc.ts";
 import { RoleData } from "../types/role.ts";
 
 export async function createRole(data: RoleData) {
+  const { tags, ...rest } = data;
+
+  const roleTags = {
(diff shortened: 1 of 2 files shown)


src/structures/role.ts
```diff
@@ -2,10 +2,19 @@ import { Unpromise } from "../types/misc.ts";
 import { RoleData } from "../types/role.ts";
 
 export async function createRole(data: RoleData) {
+  const { tags, ...rest } = data;
+
+  const roleTags = {
+    botID: tags?.bot_id,
+    premiumSubscriber: "premium_subscriber" in (data.tags ?? {}),
```
Skillz4Killz, line 9:
```suggestion
    premiumSubscriber: tags?.premium_subscriber,
```

Skillz4Killz: Why can't it be like this?

src/types/role.ts
```diff
@@ -15,4 +15,12 @@ export interface RoleData {
   managed: boolean;
   /** whether this role is mentionable */
   mentionable: boolean;
+  /**  */
```
Skillz4Killz, line 18:
```suggestion
  /** Certain roles may have tags that allow you to determine if this role is related to a bot, an integration, or the booster role. */
```

Skillz4Killz (approved):

src/structures/role.ts
ayntee: The `tags.premium_subscriber` data sent by API may either exist with the value `null` or may not exist at all.

Skillz4Killz: If its just a boolean of true when it does exist, just put it inside a Boolean(tags?.premium_subscriber) that way for DD purposes it will ALWAYS be a boolean. true or false. We dont need to worry about optional or nullor anything

src/types/role.ts
```diff
@@ -15,4 +15,12 @@ export interface RoleData {
   managed: boolean;
   /** whether this role is mentionable */
   mentionable: boolean;
+  /** Certain roles may have tags that allow you to determine if this role is related to a bot, an integration, or the booster role. */
+  tags: RoleTags | null;
+}
+
+export interface RoleTags {
+  bot_id?: string;
+  premium_subscriber?: null;
```
Skillz4Killz, line 24: Is this meant to be ?: boolean | null;

ayntee: @Skillz4Killz no, like I mentioned earlier, it may exist (sent by the API) with the value `null` or may not exist (not sent by API) at all.

src/structures/role.ts
ayntee: Alright, sounds good!

src/structures/role.ts
```diff
@@ -2,10 +2,19 @@ import { Unpromise } from "../types/misc.ts";
 import { RoleData } from "../types/role.ts";
 
 export async function createRole(data: RoleData) {
+  const { tags, ...rest } = data;
+
+  const roleTags = {
+    botID: tags?.bot_id,
+    premiumSubscriber: Boolean(data.tags?.premium_subscriber),
```
Skillz4Killz, line 9: I think you were right before, this would fail if it is null. I didn't realize this is sent by discord as null that is very confusing sorry

Skillz4Killz: 1 think is you can do tags instead of data.tags

Skillz4Killz: We should probably also add a comment above each one so it is documented in VSC

ayntee: JSDoc comments are added to the RoleData interface already.

ayntee: As for RoleTags interface, waiting for it to be documented; once it is, will do!

Skillz4Killz:
@ayntee anything pending on this? merge to next branch? feels like its ready

ayntee:
> @ayntee anything pending on this? merge to next branch? feels like its ready

Yes, you can merge it if you want. Just waiting for Discord to announce official support for this.

PR #186, opened 2020-11-13
🎉 Server Templates 🎉

ayntee:

- Added `Template` interface to types.
- Added seven new functions:
image
image
image
image
image
image
image

Reference: https://github.com/discord/discord-api-docs/pull/2144#issue-500878328

**Status**

### DIFF
--- a/src/handlers/guild.ts
+++ b/src/handlers/guild.ts
@@ -739,7 +739,7 @@ export async function syncGuildTemplate(guildID: string, templateCode: string) {
   const hasPerm = await botHasPermission(guildID, [Permissions.MANAGE_GUILD]);
   if (!hasPerm) throw new Error(Errors.MISSING_MANAGE_GUILD);
 
-  const template: GuildTemplate = await RequestManager.put(
+  const template = await RequestManager.put(
     `${endpoints.GUILD_TEMPLATES(guildID)}/${templateCode}`,
   ) as GuildTemplate;
   return structures.createTemplate(template);
--- a/src/handlers/guild.ts
+++ b/src/handlers/guild.ts
@@ -623,30 +623,99 @@ export function getGuild(guildID: string, counts = true) {
   ) as Promise<UpdateGuildPayload>;
 }
 
-/** 
- * Get a template by its code
- * @param code The code of the template
+/**
+ * Returns the guild template if it exists
+ * @param guildID The ID of the guild
+ * @param code The code of the template to get
  */
-export function getTemplate(code: string) {
-  const endpoint = `${endpoints.GUILDS}/templates/${code}`;
+export function getGuildTemplate(guildID: string, code: string) {
+  const endpoint = `${endpoints.GUILD_TEMPLATES(guildID)}/${code}`;
   return RequestManager.get(endpoint);
 }
 
 /**
- * Create a new guild based on a template.
+ * Create a new guild based on a template
  * NOTE: This endpoint can be used only by bots in less than 10 guilds.
- * @param code the code of the template to create a new guild from
- * @param name name of the guild (2-100 characters)
- * @param icon base64 128x128 image for the guild icon
+ * @param code The code of the template to create a guild from
  */
 export function createGuildFromTemplate(
   code: string,
-  name: string,
-  icon?: any,
+  data: CreateGuildFromTemplate,
 ) {
-  const endpoint = `${endpoints.GUILDS}/templates/${code}`;
-  return RequestManager.post(endpoint, {
-    name,
-    icon,
-  });
+  return RequestManager.post(endpoints.GUILD_TEMPLATE(code), data);
+}
+
+export interface CreateGuildFromTemplate {
+  /** name of the guild (2-100 characters) */
+  name: string;
+  /** base64 128x128 image for the guild icon */
+  icon?: string;
+}
+
+/**
+ * Returns an array of guild templates
+ * @param guildID The ID of the guild
+ */
+export function getGuildTemplates(guildID: string) {
+  return RequestManager.get(endpoints.GUILD_TEMPLATES(guildID));
+}
+
+/**
+ * Deletes a template from a guild
+ * @param guildID The guild ID to delete the template from
+ */
+export function deleteGuildTemplate(guildID: string, code: string) {
+  const endpoint = `${endpoints.GUILD_TEMPLATES(guildID)}/${code}`;
+  return RequestManager.delete(endpoint);
+}
+
+/**
+ * Creates a template for the guild
+ * @param guildID The ID of the guild to create a template in
+ * @param name name of the template (1-100 characters)
+ * @param description description for the template (0-120 characters
+ */
+export function createGuildTemplate(
+  guildID: string,
+  data: CreateGuildTemplate,
+) {
+  return RequestManager.post(endpoints.GUILD_TEMPLATES(guildID), data);
+}
+
+export interface CreateGuildTemplate {
+  /** name of the template (1-100 characters) */
+  name: string;
+  /** description for the template (0-120 characters) */
+  description?: string | null;
+}
+
+/**
+ * Syncs the template to the guild's current state
+ * @param guildID The ID of the guild
+ * @param code The code of the template to sync
+ */
+export function syncGuildTemplate(guildID: string, code: string) {
+  const endpoint = `${endpoints.GUILD_TEMPLATES(guildID)}/${code}`;
+  return RequestManager.put(endpoint);
+}
+
+/**
+ * Edit a template's metadata
+ * @param guildID The ID of the guild
+ * @param code The code of the template to edit
+ */
+export function editGuildTemplate(
+  guildID: string,
+  code: string,
+  data: EditGuildTemplate,
+) {
+  const endpoint = `${endpoints.GUILD_TEMPLATES(guildID)}/${code}`;
+  return RequestManager.patch(endpoint, data);
+}
+
+export interface EditGuildTemplate {
+  /** name of the template (1-100 characters) */
+  name?: string;
+  /** description for the template (0-120 characters) */
+  description?: string | null;
 }
--- a/src/handlers/guild.ts
+++ b/src/handlers/guild.ts
@@ -637,7 +637,7 @@ export function getGuildTemplate(
 ) {
   return RequestManager.get(
     `${endpoints.GUILD_TEMPLATES(guildID)}/${templateCode}`,
-  ) as any;
+  ) as Promise<Template>;
 }
 
 /**
--- a/src/handlers/guild.ts
+++ b/src/handlers/guild.ts
@@ -634,7 +634,7 @@ export function getGuild(guildID: string, counts = true) {
 export function getGuildTemplate(
   guildID: string,
   templateCode: string,
-): Promise<Template> {
+) {
   return RequestManager.get(
     `${endpoints.GUILD_TEMPLATES(guildID)}/${templateCode}`,
   ) as any;
--- a/src/handlers/guild.ts
+++ b/src/handlers/guild.ts
@@ -623,26 +623,21 @@ export function getGuild(guildID: string, counts = true) {
   ) as Promise<UpdateGuildPayload>;
 }
 
-/**
- * Returns the guild template if it exists
- * @param guildID The ID of the guild
- * @param code The code of the template to get
- */
-export function getGuildTemplate(guildID: string, code: string) {
-  const endpoint = `${endpoints.GUILD_TEMPLATES(guildID)}/${code}`;
+/** Returns the guild template if it exists */
+export function getGuildTemplate(guildID: string, templateCode: string) {
+  const endpoint = `${endpoints.GUILD_TEMPLATES(guildID)}/${templateCode}`;
   return RequestManager.get(endpoint);
 }
 
 /**
  * Create a new guild based on a template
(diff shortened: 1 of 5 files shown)


src/handlers/guild.ts
```diff
@@ -622,3 +622,100 @@ export function getGuild(guildID: string, counts = true) {
...
 }
+
+/**
+ * Returns the guild template if it exists
+ * @param guildID The ID of the guild
+ * @param code The code of the template to get
+ */
+export function getGuildTemplate(guildID: string, code: string) {
+  const endpoint = `${endpoints.GUILD_TEMPLATES(guildID)}/${code}`;
+  return RequestManager.get(endpoint);
```
Skillz4Killz, line 633: I would just move the endpoint inside the function itself. Creating a variable is actually quite expensive function for 2 reasons.

1. its a much bigger performance issue than u realize. For example, i once ran a test 
and creating a var was slower by a pretty large factor.
2. This makes it so that its taking up memory until garbage collector or some part of v8 will clean this up.

You can avoid that entirely and move this directly inside the function

Skillz4Killz, line 633: Also id like to see if we can provide a better return type than Promise<unknown>

src/handlers/guild.ts
```diff
@@ -622,3 +622,100 @@ export function getGuild(guildID: string, counts = true) {
...
+/**
+ * Create a new guild based on a template
+ * NOTE: This endpoint can be used only by bots in less than 10 guilds.
+ * @param code The code of the template to create a guild from
+ */
+export function createGuildFromTemplate(
+  code: string,
+  data: CreateGuildFromTemplate,
+) {
+  return RequestManager.post(endpoints.GUILD_TEMPLATE(code), data);
```
Skillz4Killz, line 645: Should this return a guild object?

src/handlers/guild.ts
```diff
@@ -622,3 +622,100 @@ export function getGuild(guildID: string, counts = true) {
...
+}
+
+export interface CreateGuildFromTemplate {
```
Skillz4Killz, line 648: Should we move this to a separate types folder?

src/handlers/guild.ts
```diff
@@ -622,3 +622,100 @@ export function getGuild(guildID: string, counts = true) {
...
+  /** base64 128x128 image for the guild icon */
+  icon?: string;
+}
+
+/**
+ * Returns an array of guild templates
+ * @param guildID The ID of the guild
+ */
+export function getGuildTemplates(guildID: string) {
+  return RequestManager.get(endpoints.GUILD_TEMPLATES(guildID));
```
Skillz4Killz, line 660: Id really like to see some better return type on this

```ts
const result = await RequestManager.get();
return structures.createTemplate(result);
```

src/handlers/guild.ts
```diff
@@ -622,3 +622,100 @@ export function getGuild(guildID: string, counts = true) {
...
+ * @param code The code of the template to get
+ */
+export function getGuildTemplate(guildID: string, code: string) {
+  const endpoint = `${endpoints.GUILD_TEMPLATES(guildID)}/${code}`;
+  return RequestManager.get(endpoint);
+}
+
+/**
+ * Create a new guild based on a template
+ * NOTE: This endpoint can be used only by bots in less than 10 guilds.
```
Skillz4Killz, line 638: Can you add a check for this please. If the bot is in 10 guilds throw an error

src/handlers/guild.ts
Skillz4Killz: We  should also probably handle the urlTobase64 function here for the icon

Skillz4Killz: We should also have a check to handle and throw ane rror if the name is tooo long. or too short.

src/handlers/guild.ts
```diff
@@ -622,3 +622,100 @@ export function getGuild(guildID: string, counts = true) {
...
+export function getGuildTemplates(guildID: string) {
+  return RequestManager.get(endpoints.GUILD_TEMPLATES(guildID));
+}
+
+/**
+ * Deletes a template from a guild
+ * @param guildID The guild ID to delete the template from
+ */
+export function deleteGuildTemplate(guildID: string, code: string) {
+  const endpoint = `${endpoints.GUILD_TEMPLATES(guildID)}/${code}`;
```
Skillz4Killz, line 668: Move the endpoint to inside the function

src/handlers/guild.ts
```diff
@@ -622,3 +622,100 @@ export function getGuild(guildID: string, counts = true) {
...
+ * Creates a template for the guild
+ * @param guildID The ID of the guild to create a template in
+ * @param name name of the template (1-100 characters)
+ * @param description description for the template (0-120 characters
+ */
+export function createGuildTemplate(
+  guildID: string,
+  data: CreateGuildTemplate,
+) {
+  return RequestManager.post(endpoints.GUILD_TEMPLATES(guildID), data);
```
Skillz4Killz, line 682: Handle the name characcter limit checks, same for description checks

We should return a better type as well like structures.createTemplate()

src/handlers/guild.ts
```diff
@@ -622,3 +622,100 @@ export function getGuild(guildID: string, counts = true) {
...
+  description?: string | null;
+}
+
+/**
+ * Syncs the template to the guild's current state
+ * @param guildID The ID of the guild
+ * @param code The code of the template to sync
+ */
+export function syncGuildTemplate(guildID: string, code: string) {
+  const endpoint = `${endpoints.GUILD_TEMPLATES(guildID)}/${code}`;
```
Skillz4Killz, line 698: same as others

src/handlers/guild.ts
```diff
@@ -622,3 +622,100 @@ export function getGuild(guildID: string, counts = true) {
...
+ * Edit a template's metadata
+ * @param guildID The ID of the guild
+ * @param code The code of the template to edit
+ */
+export function editGuildTemplate(
+  guildID: string,
+  code: string,
+  data: EditGuildTemplate,
+) {
+  const endpoint = `${endpoints.GUILD_TEMPLATES(guildID)}/${code}`;
```
Skillz4Killz, line 712: same move this endpoint.

Add checks for name and description character lengths

src/types/guild.ts
```diff
@@ -606,3 +607,29 @@ export interface CreateServerOptions {
...
+// https://discord.com/developers/docs/resources/template#template-object
+export interface Template {
+  /** the template code (unique ID) */
+  code: string;
+  /** template name */
+  name: string;
+  /** the description for the template */
+  description: string | null;
+  /** number of times this template has been used */
+  usage_count: number;
```
Skillz4Killz, line 620: the template structure should try and conver these snake case to camelCase like the other structs

Skillz4Killz (commented):
Lastly, i didn't see any permission checks on this. If i had to guess some permission is needed to create or edit or delete etc these templates.

1. The /** */ comments should detail which permissions are needed
2. The permissions should be checked and if missing throw an error to prevent bots from getting autobanned by not checking permissions

src/handlers/guild.ts
ayntee: Should it be `cache.guilds.size` to obtain the size of the guilds the bot is in?

src/handlers/guild.ts
```diff
@@ -622,3 +629,148 @@ export function getGuild(guildID: string, counts = true) {
     { with_counts: counts },
   ) as Promise<UpdateGuildPayload>;
 }
+
+/** Returns the guild template if it exists */
+export function getGuildTemplate(
+  guildID: string,
+  templateCode: string,
+): Promise<Template> {
```
Skillz4Killz, line 637:
```suggestion
) {
```

src/handlers/guild.ts
```diff
@@ -622,3 +629,148 @@ export function getGuild(guildID: string, counts = true) {
...
+  return RequestManager.get(
+    `${endpoints.GUILD_TEMPLATES(guildID)}/${templateCode}`,
+  ) as any;
```
Skillz4Killz, line 640:
```suggestion
  ) as Promise<Template>;
```

src/handlers/guild.ts
```diff
@@ -622,3 +629,148 @@ export function getGuild(guildID: string, counts = true) {
...
+
+/**
+ * Create a new guild based on a template
+ * NOTE: This endpoint can be used only by bots in less than 10 guilds.
+ */
+export async function createGuildFromTemplate(
+  templateCode: string,
+  data: CreateGuildFromTemplate,
+) {
+  if (cache.guilds.size >= 10) {
```
Skillz4Killz, line 651: this should use cachehandlers

src/handlers/guild.ts
```diff
@@ -622,3 +629,148 @@ export function getGuild(guildID: string, counts = true) {
...
+    throw new Error(
+      "This function can only be used by bots in less than 10 guilds.",
+    );
+  }
+
+  if (data.icon) {
+    data.icon = await urlToBase64(data.icon);
+  }
+
+  try {
```
Skillz4Killz, line 661: I don't understand this try catch. There is no point of a try catch that just throws the error again. That just makes this code slower.

```ts
return RequestManager.post() as Promise<CreateGuildPayload>
```

src/handlers/guild.ts
```diff
@@ -622,3 +629,148 @@ export function getGuild(guildID: string, counts = true) {
...
+ * Requires the `MANAGE_GUILD` permission.
+ */
+export async function deleteGuildTemplate(
+  guildID: string,
+  templateCode: string,
+) {
+  const hasPerm = await botHasPermission(guildID, [Permissions.MANAGE_GUILD]);
+  if (!hasPerm) throw new Error(Errors.MISSING_MANAGE_GUILD);
+
+  const deletedTemplate: GuildTemplate = await RequestManager.delete(
```
Skillz4Killz, line 697:
```suggestion
  const deletedTemplate = await RequestManager.delete(
```

src/handlers/guild.ts
```diff
@@ -622,3 +629,148 @@ export function getGuild(guildID: string, counts = true) {
...
+    `${endpoints.GUILD_TEMPLATES(guildID)}/${templateCode}`,
+  ) as any;
```
Skillz4Killz, line 699:
```suggestion
  ) as GuildTemplate;
```

src/handlers/guild.ts
```diff
@@ -622,3 +629,148 @@ export function getGuild(guildID: string, counts = true) {
...
+
+/**
+ * Syncs the template to the guild's current state.
+ * Requires the `MANAGE_GUILD` permission.
+ */
+export async function syncGuildTemplate(guildID: string, templateCode: string) {
+  const hasPerm = await botHasPermission(guildID, [Permissions.MANAGE_GUILD]);
+  if (!hasPerm) throw new Error(Errors.MISSING_MANAGE_GUILD);
+
+  const template: GuildTemplate = await RequestManager.put(
```
Skillz4Killz, line 742:
```suggestion
  const template = await RequestManager.put(
```

src/handlers/guild.ts
```diff
@@ -622,3 +629,148 @@ export function getGuild(guildID: string, counts = true) {
...
+    `${endpoints.GUILD_TEMPLATES(guildID)}/${templateCode}`,
+  ) as any;
```
Skillz4Killz, line 744:
```suggestion
  ) as GuildTemplate;
```

src/handlers/guild.ts
```diff
@@ -622,3 +629,148 @@ export function getGuild(guildID: string, counts = true) {
...
+  }
+
+  if (
+    data.description?.length &&
+    data.description.length > 120
+  ) {
+    throw new Error("The description can only be in between 0-120 characters.");
+  }
+
+  const template: GuildTemplate = await RequestManager.patch(
```
Skillz4Killz, line 771: same as abovre

Skillz4Killz (commented):
Tiny nitpicks

src/handlers/guild.ts
ayntee: @Skillz4Killz, can you provide an example?

Skillz4Killz: 1. Add a .size option here https://github.com/Skillz4Killz/Discordeno/blob/0c8731679c376c2b75f3df3ca6401623fbce1912/src/controllers/cache.ts#L95
2. if (await cacheHandlers.size("guilds") >= 10) {

}

ayntee: But that is irrelevant to the PR, I don't think that should be done in this PR.

ayntee: #188

src/handlers/guild.ts
```diff
@@ -622,3 +628,144 @@ export function getGuild(guildID: string, counts = true) {
...
+
+/**
+ * Create a new guild based on a template
+ * NOTE: This endpoint can be used only by bots in less than 10 guilds.
+ */
+export async function createGuildFromTemplate(
+  templateCode: string,
+  data: CreateGuildFromTemplate,
+) {
+  if (cacheHandlers.size("guilds") >= 10) {
```
Skillz4Killz, line 650:
```suggestion
  if (await cacheHandlers.size("guilds") >= 10) {
```

Skillz4Killz: await this in cases of custom cache someone stores it on redis and needs an await

ayntee: @Skillz4Killz, size() is not an async function.

Issue #191, opened 2020-11-15
In RequestManager The Version Number is outdated, 


Skillz4Killz:
probably should make it dynamic somehow

ayntee:
Uh, this would have been automatically closed once the `v10` issue/PR is closed/merged.

PR #192, opened 2020-11-15
Rename and move constants/discord.ts to utils/constants.ts

ayntee:
Related: #191 

Once this is merged, a majorVersion var will be exported from constants.ts, which will be updated on every major release.


Skillz4Killz:
This probably needs to be redirected to the next branch. This is technically a breaking change for any users that were getting endpoints directly from the old file.

ayntee:
The master branch is not the stable branch, but development branch.
On each release, the master branch should be merged into stable branch.

Issue #193, opened 2020-11-15
v10


Skillz4Killz:
- [x] Overridable handlers

```ts
export const handlers = {
    addRole: () => {},
    ...
};
```

- [ ] Separate REST to separate package. Useful when someone wants to create a web dashboard or a REST proxy(big brain bots).
- [ ] Separate WS to separate package. Useful when someone wants to create a proxy ws for big brain bots
- [x] Better error messages for HTTP
- [x] Better error messages for WebSocket
- [x] Stickers support
- [x] Switch to web WebSocket API
- [x] Change `Permissions[]` to `Permission[]` for consistency
- [x] Slash commands and interactions (#247, #248)
- [x] Better directory structure (#258)
- [ ] Implement getters (#259, #262)

Skillz4Killz:
@ayntee Let me know ur thoughts on these and if u have any other ideas ud like to add

ayntee:
Better error messages for request manager

ayntee:
`Guild.roles` -> `cache.roles`

Skillz4Killz:
Error messages i totally agree with, but guild roles are you sure? Roles are unique to each server, none of them share anything  with another role from another server. cache.members make sense because they share username, discriminator, avatar etc... cache.channels makes sense because its literally just duplicated data.

Skillz4Killz:
Support for REST and WS proxy should be good to go but needs testing. We still have to consider if it's worth making our rest and ws an entirely separate package as well

ayntee:
> Support for REST and WS proxy should be good to go but needs testing. We still have to consider if it's worth making our rest and ws an entirely separate package as well #161

Although, this would be a good edition to the library, this will be a major change and we should discuss this and get opinions from the users. For now, an average user may not at all need to customize ws or rest.

Skillz4Killz:
Change all Permissions[] to Permission[] i think would also be good for users. This allows them to use the same consistent form to work with the lib as opposed to some confusion and also removes the need to have to import the permissions enum

ayntee:
Before merging the `v10` PR, maybe we should nuke versioned branches as well.

Skillz4Killz:
@ayntee ideally I would like to wait until Discord nukes v6 because then all those branches will stop functioning. In case anyone is using the raw github url from one of those commits/branches id like to keep it alive until discord kills v6

userterminal:
Anything on the status of stickers support?

Skillz4Killz:
[...]

PR #196, opened 2020-11-16
Better error messages for request manager

ayntee:
Related: #193

### DIFF
--- a/src/module/requestManager.ts
+++ b/src/module/requestManager.ts
@@ -200,7 +200,7 @@ async function runMethod(
     },
   );
 
-  const errorStack = new Error("Location In Your Files:");
+  const errorStack = new Error("Location:");
   Error.captureStackTrace(errorStack);
 
   return new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


src/module/requestManager.ts
```diff
@@ -321,13 +321,25 @@ function handleStatusCode(response: Response, errorStack?: unknown) {
 
   switch (status) {
     case HttpResponseCode.BadRequest:
+      throw new Error(
```
Skillz4Killz, line 324: Sadly, this would still not improve the stack frames showing where the error came from. I believe this should be changed slightly. Instead of `throw new Error()` this line would become `console.error(....)` and then below it we add `throw errorStack`

Skillz4Killz: side note: i am looking into how to get more than 3-4 stack frames i believe this might be something deno limits but v8 should give like 10 lines

ayntee: Noted, that should be done in another PR, again because it would be irrelevant to this PR. Title 🙄

PR #237, opened 2020-12-08
fix: if bot is guild owner, do not check role position

ayntee:

**Status**

### DIFF
--- a/src/handlers/member.ts
+++ b/src/handlers/member.ts
@@ -53,6 +53,16 @@ export async function addRole(
   roleID: string,
   reason?: string,
 ) {
+  const guild = await cacheHandlers.get("guilds", guildID);
+  if (!guild) return;
+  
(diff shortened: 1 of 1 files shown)


src/handlers/member.ts
```diff
@@ -53,6 +53,16 @@ export async function addRole(
   roleID: string,
   reason?: string,
 ) {
+  const guild = await cacheHandlers.get("guilds", guildID);
+  if (!guild) return;
+  
+  if (guild.ownerID === botID) {
```
Skillz4Killz, line 59: Could we replace this with 
```ts
if (guild && guild.ownerID === botID)
```

Skillz4Killz: Instead of if (!guild) this would allow us to continue checking and using these without needing guild in cache stored

PR #268, opened 2020-12-22
refactor: organize directories and files

ayntee:
`src/` directory:
```md
- bot.ts
- api/
    - controllers/
    - structures/
    - handlers/
- rest/
    - mod.ts
- ws/
    - mod.ts
    - shard_manager.ts
    - deps.ts
- types/
- util/
```


src/util/utils.ts
```diff
@@ -32,3 +32,11 @@ export async function urlToBase64(url: string) {
   const type = url.substring(url.lastIndexOf(".") + 1);
   return `data:image/${type};base64,${imageStr}`;
 }
+
+export function delay(ms: number): Promise<void> {
```
Skillz4Killz, line 36: Why not just use delay from deno.land

ayntee: Both are literally the same thing?

PR #302, opened 2020-12-30
Fix permissions

TriForMine:
Fix permissions (sorry my ide changed & to &&)

**Status**

- [ x] Code changes have been tested against the Discord API, or there are no code changes

### DIFF
--- a/src/util/permissions.ts
+++ b/src/util/permissions.ts
@@ -82,11 +82,11 @@ export async function botHasPermission(
 }
 
 /** Checks if the bot has the permissions in a channel */
-export function botHasChannelPermissions(
+export async function botHasChannelPermissions(
   channelID: string,
   permissions: Permission[],
 ) {
-  return hasChannelPermissions(channelID, botID, permissions);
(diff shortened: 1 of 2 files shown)


src/util/permissions.ts
```diff
@@ -74,19 +74,19 @@ export async function botHasPermission(
...
 }
 
 /** Checks if the bot has the permissions in a channel */
-export function botHasChannelPermissions(
+export async function botHasChannelPermissions(
   channelID: string,
   permissions: Permission[],
 ) {
-  return hasChannelPermissions(channelID, botID, permissions);
+  return await hasChannelPermissions(channelID, botID, permissions);
```
ayntee, line 89: Since hasChannelPermissions() returns a Promise, you do not have to convert the function to `async`.

TriForMine: It's seems to be needed the function doesn't work without it

ayntee: Confirmed, this does not seem to fix the issue either.

ayntee (approved):
Thank you, @TriForMine!

PR #304, opened 2020-12-30
fix: remove role from coll after the event

ayntee:


Skillz4Killz (approved):

src/api/controllers/roles.ts
```diff
@@ -27,9 +27,10 @@ export async function handleInternalGuildRoleDelete(data: DiscordPayload) {
   if (!guild) return;
 
   const cachedRole = guild.roles.get(payload.role_id)!;
-  guild.roles.delete(payload.role_id);
   eventHandlers.roleDelete?.(guild, cachedRole);
```
Skillz4Killz, line 30:
```suggestion
  if (cachedRole) eventHandlers.roleDelete?.(guild, cachedRole);
```

Skillz4Killz: Just to be safe

PR #331, opened 2021-01-02
chore: setup "deno lint"

ayntee:
- Changed `lint` workflow to use `deno lint` instead of `deno fmt` to lint source files.
    *Currently, `deno lint` is unstable, and hence the `--unstable` flag.
- Updated to comply with "best" and `deno lint` practices.
- Ignored lint warnings, if required, using `deno-lint-ignore` comments.

**This PR will be un-drafted when `deno lint` is stabilized.**


src/api/handlers/guild.ts
```diff
@@ -429,6 +429,8 @@ export function fetchMembers(guild: Guild, options?: FetchMembersOptions) {
     options.limit = options.userIDs.length;
   }
 
+  // TODO: redundant async function
```
ayntee, line 432: @Skillz4Killz any reason why this is `async` function.

Skillz4Killz:
There is a test failing

ayntee:
> There is a test failing

Was waiting for the "pin & unpin" PR to get merged since the issue is related to that.

Skillz4Killz (approved):

PR #333, opened 2021-01-02
Fix messageUpdate event

OSCAR-WOS:
Logic failure incurred receiving events. Strong case not to compare content, as modifying embes and removing images still classify as a messageUpdate and not a messageDelete

**Status**

  - [] This PR includes breaking changes (methods removed or renamed, parameters moved or removed)

### DIFF
--- a/src/api/controllers/messages.ts
+++ b/src/api/controllers/messages.ts
@@ -94,13 +94,6 @@ export async function handleInternalMessageUpdate(data: DiscordPayload) {
     pinned: cachedMessage.pinned,
   };
 
-  // Messages with embeds can trigger update but they wont have edited_timestamp
-  if (
-    !payload.edited_timestamp ||
-    (cachedMessage.content !== payload.content)
-  ) {
-    return;
-  }
-
+  if (!payload.edited_timestamp) return;
   eventHandlers.messageUpdate?.(cachedMessage, oldMessage);
 }
--- a/src/api/controllers/misc.ts
+++ b/src/api/controllers/misc.ts
@@ -30,8 +30,6 @@ export async function handleInternalReady(
   // Triggered on each shard
   eventHandlers.shardReady?.(shardID);
   if (payload.shard && shardID === payload.shard[1] - 1) {
-    // Wait for 3 seconds to allow all guild create events to be processed
-    await delay(3000);
     cache.isReady = true;
     eventHandlers.ready?.();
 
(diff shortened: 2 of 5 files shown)


itohatweb:
There are too many unnecessary commits.

ayntee:
I don't think some of the changes made are related to this PR.

src/api/controllers/messages.ts
```diff
@@ -94,13 +94,6 @@ export async function handleInternalMessageUpdate(data: DiscordPayload) {
...
 
-  // Messages with embeds can trigger update but they wont have edited_timestamp
-  if (
-    !payload.edited_timestamp ||
-    (cachedMessage.content !== payload.content)
-  ) {
-    return;
-  }
-
+  if (!payload.edited_timestamp) return;
```
Skillz4Killz, line 97: I am not sure this really should change. I'm not really convinced. IMO for 99% of bots this is the ideal version of this event. If a bot dev needs more its easily changeable with a controller.

Skillz4Killz: It also help prevent wasting CPU usage for all these bots that don't need it.

src/api/controllers/misc.ts
```diff
@@ -27,11 +27,31 @@ export async function handleInternalReady(
   const payload = data.d as ReadyPayload;
   setBotID(payload.user.id);
 
-  // Triggered on each shard
+  cache.guildsPartial += payload.guilds.length;
   eventHandlers.shardReady?.(shardID);
+
```
Skillz4Killz, line 32: All of this, is massively overcomplicating something as simple as an if() with a promise.race to check if anything is missing. IMO even that is overcomplicated a simple 5s delay but this would much morelikely pass as a simple if and not more vars and data eating up memory like this.

src/util/cache.ts
```diff
@@ -9,6 +9,7 @@ import { Collection } from "./collection.ts";
 
 export interface CacheData {
   isReady: boolean;
+  guildsPartial: number;
```
Skillz4Killz, line 12: Partials are confusing as hell we should always avoid whenever possible

Skillz4Killz (changes requested):

src/api/controllers/misc.ts
Skillz4Killz:
```ts
const arrived = await Promise.something(payload.guilds.map(g => cacheHandlers.get('guilds', g.id)))
if (!arrived) await delay(5000)
```

Skillz4Killz: Something like that

src/api/controllers/messages.ts
OSCAR-WOS: Having the original logic would not fire messageUpdate at all

PR #338, opened 2021-01-02
fix: cachedMessage content on messageUpdate

OSCAR-WOS:
Upon firing a messageUpdate event, cachedMessage's content was not reflected

**Status**


src/api/controllers/messages.ts
```diff
@@ -85,6 +85,8 @@ export async function handleInternalMessageUpdate(data: DiscordPayload) {
   const cachedMessage = await cacheHandlers.get("messages", payload.id);
   if (!cachedMessage) return;
 
+  cachedMessage.content = payload.content;
```
Skillz4Killz, line 88: This should bemuch lower likeone of the last things that is done in this function. If you overwrite it this early everything below will break

Skillz4Killz (changes requested):

PR #351, opened 2021-01-07
fix: slash command edit functions

itohatweb:
Please describe the changes this PR makes and why it should be merged:

upsertSlashCommand() does not actually edit the slash command. This is fixing this issue
Status

 Code changes have been tested against the Discord API, or there are no code changes
Semantic versioning classification:

 This PR changes the library's interface (methods or parameters added)
 This PR includes breaking changes (methods removed or renamed, parameters moved or removed)
 This PR only includes non-code changes, like changes to documentation, README, etc.

### DIFF
--- a/src/api/handlers/webhook.ts
+++ b/src/api/handlers/webhook.ts
@@ -9,7 +9,6 @@ import {
   ExecuteSlashCommandOptions,
   ExecuteWebhookOptions,
   MessageCreateOptions,
-  UpsertSlashCommandOptions,
   WebhookCreateOptions,
   WebhookPayload,
 } from "../../types/mod.ts";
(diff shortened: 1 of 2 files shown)


src/api/handlers/webhook.ts
```diff
@@ -227,23 +226,29 @@ export function getSlashCommands(guildID?: string) {
 /**
  * Edit an existing slash command. If this command did not exist, it will create it.
  */
-export function upsertSlashCommand(options: UpsertSlashCommandOptions) {
+export function upsertSlashCommand(
+  commandID: string,
```
Skillz4Killz, line 230: What is a commandID?

itohatweb: Its the ID od the command.
https://discord.com/developers/docs/interactions/slash-commands#edit-global-application-command

ayntee: What if the command does not exist?

itohatweb: Right we should check that before sending the request.

ayntee: But the parameter `commandID` is not optional, but previously it was.

itohatweb: No `guildID` was optional

itohatweb: No? you need it for the URL

Skillz4Killz: I still don't know what a commandID is

itohatweb: Every command has a unique ID. You need to pass this ID so discord knows what command you want to edit.

image

Skillz4Killz (approved):

PR #371, opened 2021-01-12
test(message): add missing tests

ayntee:
All tests are passing

### DIFF
--- a/tests/05_message.ts
+++ b/tests/05_message.ts
@@ -72,6 +72,8 @@ Deno.test({
 Deno.test({
   name: "[message] add a reaction to a message",
   fn() {
+    // TODO: add tests for a guild emoji ― <:name:id>
+
     addReaction(tempData.channelID, tempData.messageID, "👍");
   },
   ...defaultTestOptions,
--- a/tests/05_message.ts
+++ b/tests/05_message.ts
@@ -1,5 +1,15 @@
 import { defaultTestOptions, tempData } from "./01_main.ts";
-import { assertEquals, assertExists, getMessage, sendMessage } from "./deps.ts";
+import {
+  addReaction,
+  assertEquals,
+  assertExists,
+  getMessage,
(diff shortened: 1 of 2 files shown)


tests/05_message.ts
```diff
@@ -29,3 +39,50 @@ Deno.test({
...
+  fn() {
+    unpin(tempData.channelID, tempData.messageID);
+  },
+  ...defaultTestOptions,
+});
+
+Deno.test({
+  name: "[message] add a reaction to a message",
+  fn() {
+    addReaction(tempData.channelID, tempData.messageID, "👍");
```
Skillz4Killz, line 75: Please test the <:name:id> version also somehow

tests/05_message.ts
```diff
@@ -29,3 +39,50 @@ Deno.test({
     assertEquals(message.embeds[0].title, "Discordeno Test");
   },
 });
+
+Deno.test({
+  name: "[message] pin a message in a channel",
+  fn() {
+    pin(tempData.channelID, tempData.messageID);
```
Skillz4Killz, line 46: We should also test this through getters imo.
`message.pin(channelID, anotherMessge.id)`

Skillz4Killz: getters should be tested for everything as well

tests/05_message.ts
Skillz4Killz: We should also test addReactions

Skillz4Killz (approved):

tests/05_message.ts
ayntee: @Skillz4Killz, the getter is basically calling `pin()` internally.

tests/05_message.ts
ayntee: @Skillz4Killz, again, the `addReactions()` function `Promise.all()`s the `addReaction()` function.

PR #401, opened 2021-01-19
feat(handlers): inhibitors for deleteChannel()

itohatweb:
Adds check if channel to delete is a rules or updates channel since these cannot be deleted.

**Status**


ayntee:
It makes sense for the first parameter to be `guildID` rather than `channelID`. A guild channel is a constituent of a guild, that is, a guild is a larger and bigger element relative to a guild channel. Henceforth, the recommended argument order should be as follows:
`guildID` > `channelID`/`roleID`/`emojiID` > `messageID`

itohatweb:
You're right, then this PR is not breaking

ayntee:
Personally, I am moderately inert about this pull-request, but slightly, I am against this pull-request. It is not the library's obligation to inspect these edge-cases.

itohatweb:
Well it is one invalid request if it is not checked.
`The request was improperly formatted, or the server couldn't understand it`

ayntee:
> Well it is one invalid request if it is not checked.
> `The request was improperly formatted, or the server couldn't understand it`

Discordeno already does a lot of heavy-lifting for users like permission-checking, so on. I don't think this is necessary for users, we already have a rate-limiting handler.

Skillz4Killz (approved):

Skillz4Killz:
I think this is absolutely needed @ayntee In my opinion this is something that EVERY library should do but only we end up doing. This is essentially rate limiting https://discord.com/developers/docs/topics/rate-limits#invalid-request-limit

ayntee:
> I think this is absolutely needed @ayntee In my opinion this is something that EVERY library should do but only we end up doing. This is essentially rate limiting https://discord.com/developers/docs/topics/rate-limits#invalid-request-limit

Discordeno internally has a queue system for rate-limits.

PR #408, opened 2021-01-19
feat(handlers): add getGuildPreview()

itohatweb:

**Status**


src/api/handlers/guild.ts
```diff
@@ -595,6 +595,12 @@ export async function unban(guildID: string, id: string) {
   return RequestManager.delete(endpoints.GUILD_BAN(guildID, id));
 }
 
+/** Returns the guild preview object for the given id. If the bot is not in the guild, then the guild must be Discoverable. */
+export function getGuildPreview(guildID: string) {
+  // TODO: add check if guild is Discoverable?
```
ayntee, line 600: You can just retrieve the guild from the cache using the handler and check if the guild is discoverable or not?

itohatweb: No, it is not possible to get the guild from cache. Since that only needs to be checked if the bot is not member of the guild. That means we would need to get the guild from API and then check. Cause of that I added to do. What do you want?

ayntee: We can check whether the bot is in the guild or not―that is, the guild will not be cached then.

itohatweb: I have added a check can you take a look?

PR #412, opened 2021-01-20
feat(handlers): add getMembers()

itohatweb:

**Status**


src/api/handlers/guild.ts
```diff
@@ -445,6 +446,28 @@ export function fetchMembers(guild: Guild, options?: FetchMembersOptions) {
...
+*/
+export function getMembers(
+  guildID: string,
+  options?: GetMemberOptions,
+) {
+  if (!(identifyPayload.intents && Intents.GUILD_MEMBERS)) {
+    throw new Error(Errors.MISSING_INTENT_GUILD_MEMBERS);
+  }
+
+  return RequestManager.get(
```
Skillz4Killz, line 465: 1. Create the structures.
2. Add to cache
3. Return the full member structs
4. This function should really support pagination by default

Skillz4Killz (changes requested):

src/api/handlers/guild.ts
```diff
@@ -445,6 +446,46 @@ export function fetchMembers(guild: Guild, options?: FetchMembersOptions) {
...
+    throw new Error(Errors.MISSING_INTENT_GUILD_MEMBERS);
+  }
+
+  const result = (await RequestManager.get(
+    endpoints.GUILD_MEMBERS(guildID),
+    options,
+  )) as MemberCreatePayload[];
+
+  const memberStructures = (await Promise.all(
+    result.map(async (member) =>
```
Skillz4Killz, line 471:
```suggestion
    result.map(async (member) => {
```

src/api/handlers/guild.ts
```diff
@@ -445,6 +446,46 @@ export function fetchMembers(guild: Guild, options?: FetchMembersOptions) {
...
+      await structures.createMember(member, guildID)
+    ),
```
Skillz4Killz, line 473:
```suggestion
    }),
```

src/api/handlers/guild.ts
```diff
@@ -445,6 +446,46 @@ export function fetchMembers(guild: Guild, options?: FetchMembersOptions) {
...
+  }
+
+  const result = (await RequestManager.get(
+    endpoints.GUILD_MEMBERS(guildID),
+    options,
+  )) as MemberCreatePayload[];
+
+  const memberStructures = (await Promise.all(
+    result.map(async (member) =>
+      await structures.createMember(member, guildID)
```
Skillz4Killz, line 472:
```suggestion
      const member = await structures.createMember(member, guildID);
      await cacheHandlers.set("members", member.id, member)
```

src/api/handlers/guild.ts
```diff
@@ -445,6 +446,46 @@ export function fetchMembers(guild: Guild, options?: FetchMembersOptions) {
...
+    ),
+  )) as Member[];
+
+  const members = new Collection(
```
Skillz4Killz, line 476: return this

Skillz4Killz (changes requested):
This still is missing pagination, this endpoint only supports up to 1000 members max per request, you have to send multple requests if u want more so we should support that in an easy way

Skillz4Killz (commented):
Pending changes from previous revew

itohatweb:
Paginating is implemented now and should theoretically work. 
(Since I am not owner of a server with 1000+ members I was not able to test it with that high numbers.)

src/api/handlers/guild.ts
```diff
@@ -462,20 +462,40 @@ export async function getMembers(
     throw new Error(Errors.MISSING_INTENT_GUILD_MEMBERS);
   }
 
-  const result = (await RequestManager.get(
-    endpoints.GUILD_MEMBERS(guildID),
-    options,
-  )) as MemberCreatePayload[];
+  const members = new Collection<string, Member>();
+  let loops = 1;
+  while (options?.limit ?? 1 > members.size) {
```
Skillz4Killz, line 559: This needs some modifications, pagination system should only run if the user provides NO limit as in if limit is undefined OR if the limit is more than 1000

src/api/handlers/guild.ts
```diff
@@ -462,20 +462,40 @@ export async function getMembers(
...
+    console.log(
+      `Paginating get members from REST. #${loops} / ${
```
Skillz4Killz, line 561: Logging this even when the user only requests 1000 or less doesnt make sense

src/api/handlers/guild.ts
```diff
@@ -462,20 +462,40 @@ export async function getMembers(
...
+    );
 
-  const memberStructures = (await Promise.all(
-    result.map(async (member) =>
-      await structures.createMember(member, guildID)
-    ),
-  )) as Member[];
+    const result = (await RequestManager.get(
+      `${endpoints.GUILD_MEMBERS(guildID)}?limit=${
+        options?.limit ?? 1 > 1000 ? 1000 : options?.limit
```
Skillz4Killz, line 568: during pagination imo limit should always be 1000

src/api/handlers/guild.ts
```diff
@@ -462,20 +462,40 @@ export async function getMembers(
...
+      }${options?.after ? `&after=${options.after}` : ""}`,
+    )) as MemberCreatePayload[];
 
-  return new Collection(
-    memberStructures.map((member) => [member.id, member]),
-  );
+    const memberStructures = (await Promise.all(
```
Skillz4Killz, line 572: Seems like this has an extra ()

src/api/handlers/guild.ts
```diff
@@ -462,20 +462,40 @@ export async function getMembers(
...
+        await structures.createMember(member, guildID)
+      ),
+    ) as Member[]);
+
+    if (!memberStructures.length) break;
+
+    memberStructures.forEach((member) => members.set(member.id, member));
+
+    options = {
+      limit: options?.limit ?? 1 - 1000,
```
Skillz4Killz, line 583: This is no good imo, if its undefined now this suddenly becomes -999

src/api/handlers/guild.ts
```diff
@@ -462,20 +462,40 @@ export async function getMembers(
...
-  const result = (await RequestManager.get(
-    endpoints.GUILD_MEMBERS(guildID),
-    options,
-  )) as MemberCreatePayload[];
+  const members = new Collection<string, Member>();
+  let loops = 1;
+  while (options?.limit ?? 1 > members.size) {
+    console.log(
+      `Paginating get members from REST. #${loops} / ${
+        Math.ceil((options?.limit ?? 1) / 1000)
```
Skillz4Killz, line 562: if limit is undefined what would this be?

Skillz4Killz (changes requested):

src/api/handlers/guild.ts
itohatweb: Can do that but actually not needed I made it so that the system only runs once if limit is under 1000 or undefined..

Skillz4Killz: The console.log should not occur in that case

src/api/handlers/guild.ts
itohatweb: what if user has a limit of eg. 1823 then the function would get 2000 members which is unnecessary.

src/api/handlers/guild.ts
itohatweb: Cannot be undefined due to this
```ts
if (!memberStructures.length) break;
```

src/api/handlers/guild.ts
itohatweb: If limit is undefined it would be 1

src/api/handlers/guild.ts
itohatweb: Well it only loops one time if it is less than 1000.

src/api/handlers/guild.ts
ayntee: Also, can you remove the console.log statements?

itohatweb: Why? Skillz wanted an console.log here

Skillz4Killz (approved):

PR #423, opened 2021-01-20
feat(handlers/user): add getCurrentUser()

itohatweb:

**Status**

- [x] Code changes have been tested against the Discord API, or there are no
  code changes

  - [ ] This PR includes breaking changes (methods removed or renamed,
    parameters moved or removed)
  README, etc.


src/api/handlers/user.ts
```diff
@@ -0,0 +1,40 @@
...
+ * 
+ * ⚠️ **If you need this, you are probably doing something wrong. Always use cache.guilds
+ *
+ * Advanced Devs:
+ * This function gets you a list of all guilds in those your bot is in.
+ */
+export function getCurrentUserGuilds() {
+  return RequestManager.get(endpoints.USER_GUILDS) as Promise<
+    Partial<CreateGuildPayload>
+  >;
```
ayntee, line 28:
```suggestion
  const guilds = RequestManager.get(endpoints.USER_GUILDS) as Promise<Partial<CreateGuildPayload>[]>;
  return guilds.map(guild => structures.createGuild(guild));
```

itohatweb: I don't recommend that. This endpoint is only used to get the guilds the current user is in. 
Also it returns sth like that:
```ts
{
  "id": "80351110224678912",
  "name": "1337 Krew",
  "icon": "8342729096ea3675442027381ff50dfe",
  "owner": true,
  "permissions": "36953089",
  "features": ["COMMUNITY", "NEWS"]
}
```
Therefore guilds cannot be created by structures since they are partial (as far as I know).

Skillz4Killz (changes requested):

src/api/handlers/user.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+/** Returns a list of connection objects. Requires the `connections` OAuth2 scope. */
+export function getCurrentUserConnections() {
+  // TODO(itohatweb): types: better return type
+  return RequestManager.get(endpoints.USER_CONNECTIONS);
+}
+
+/** This function will return the raw user payload in the rare cases you need to fetch a user directly from the API. */
+export function getUser(userID: string) {
```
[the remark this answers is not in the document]
ayntee: Breaking change, why? We are just moving this function from `guild` handler to a new handler `user`.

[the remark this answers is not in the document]
Skillz4Killz: Not everyone will import from `mod.ts` moving this could be considerd a breaking change if someone is importing from say https://deno.land/x/discordeno@10.1.0/src/api/handlers/guild.ts

src/api/handlers/user.ts
```diff
@@ -0,0 +1,40 @@
...
+ * Advanced Devs:
+ * This function gets you a list of all guilds in those your bot is in.
+ */
+export function getCurrentUserGuilds() {
+  return RequestManager.get(endpoints.USER_GUILDS) as Promise<
+    Partial<CreateGuildPayload>
+  >;
+}
+
+/** Returns a list of connection objects. Requires the `connections` OAuth2 scope. */
```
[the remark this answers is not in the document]
itohatweb: Connection Objects represent the guild connections. Eg you can connect your guild with your Twitch and YouTube account so there are two connection objects one represents the Twitch and the other the YouTube connection.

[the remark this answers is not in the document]
Skillz4Killz: This should be written in that comment above to explain it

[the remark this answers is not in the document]
itohatweb: After thinking it does not make sense to cover this endpoint. I was wrong with my first response. This function returns the current users connections and not the guild ones and since bots cannot have connections we don't need to cover that one.

src/api/handlers/user.ts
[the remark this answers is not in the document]
ayntee: @Skillz4Killz, it is NOT RECOMMENDED TO or you SHOULD NOT export functions directly from a specific file, but always export an entity from the main file, that is, `mod.ts` conventionally.

src/api/handlers/user.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Requires the `guilds` OAuth2 scope.
+ * 
+ * ⚠️ **If you need this, you are probably doing something wrong. Always use cache.guilds
+ *
+ * Advanced Devs:
+ * This function gets you a list of all guilds in those your bot is in.
+ */
+export function getCurrentUserGuilds() {
+  return RequestManager.get(endpoints.USER_GUILDS) as Promise<
+    Partial<CreateGuildPayload>
```
Skillz4Killz, line 27: This type looks wrong to me. I seriously doubt Discord would send a payload like this where even the id is possibly undefined

src/api/handlers/user.ts
```diff
@@ -0,0 +1,29 @@
...
+  return RequestManager.get(endpoints.USER_BOT) as Promise<
+    UserPayload
+  >;
+}
+
+/**
+ * Returns a list of partial guild objects the current user is a member of. 
+ * Requires the `guilds` OAuth2 scope.
+ * 
+ * ⚠️ **If you need this, you are probably doing something wrong. Always use cache.guilds
```
Skillz4Killz, line 20: I still think this comment is wrong big time. This is not about getting the bots guilds but the end user who is logging in

Skillz4Killz: Please double check this is what the function does

ayntee: https://discord.com/developers/docs/resources/user#get-current-user-guilds

src/api/handlers/user.ts
ayntee: A list of `PartialGuild` is return type for this endpoint.
image

src/api/handlers/user.ts
itohatweb: Hmm I think ima remove that

src/api/handlers/user.ts
```diff
@@ -0,0 +1,14 @@
+import { RequestManager } from "../../rest/request_manager.ts";
+import { UserPayload } from "../../types/mod.ts";
+import { endpoints } from "../../util/constants.ts";
+
+/**
+ * Returns the user object of the requester's account.
```
Skillz4Killz, line 6: what happens when this function is called? what values are u getting back?

ayntee: What Skillz meant is the you should cast that to the return type...

src/api/handlers/user.ts
```diff
@@ -0,0 +1,14 @@
...
+ * For OAuth2, this requires the `identify` scope, which will return the object without an email,
+ * and optionally the email scope, which returns the object with an email.
+ */
+export function getCurrentUser() {
+  return RequestManager.get(endpoints.USER_BOT) as Promise<
+    UserPayload
+  >;
+}
```
ayntee, line 14:
```suggestion
export async function getCurrentUser() {
  const result = await RequestManager.get(endpoints.USER_BOT) as UserPayload;
  return result;
}
```

itohatweb:
Thought I have closed that one

PR #440, opened 2021-01-23
refactor(util): COMMANDS_GUILD_ID constant arg order

itohatweb:
**Please describe the changes this PR makes and why it should be merged:**

**Status**

- [x] Code changes have been tested against the Discord API, or there are no
  code changes

**Semantic versioning classification:**

- [x] This PR changes the library's interface (methods or parameters added)
  - [ ] This PR includes breaking changes (methods removed or renamed,
    parameters moved or removed)
- [ ] This PR **only** includes non-code changes, like changes to documentation,
  README, etc.

PR #441, opened 2021-01-23
fix(handlers): edit/upsertSlashCommand function

itohatweb:
IMPORTANT: This PR REQUIRES #440 to be merged.

NOTE: This PR would includes breaking changes but since it is not possible that anyone has used this functions because they do not work merging this PR should be no problem.

**Status**

- [ ] Code changes have been tested against the Discord API, or there are no
  code changes

  - [x] This PR includes breaking changes (methods removed or renamed,
    parameters moved or removed) (This function was never used)
  README, etc.

### DIFF
--- a/src/api/handlers/webhook.ts
+++ b/src/api/handlers/webhook.ts
@@ -253,6 +253,7 @@ export function upsertSlashCommand(
   return result;
 }
 
+// TODO: remove this function for v11
 /** 
  * Edit an existing slash command. 
  * @deprecated This function will be removed in v11. Use `upsertSlashCommand()` instead
--- a/src/api/handlers/webhook.ts
+++ b/src/api/handlers/webhook.ts
@@ -242,7 +242,7 @@ export function upsertSlashCommand(options: UpsertSlashCommandOptions) {
 export function editSlashCommand(
   commandID: string,
   options: EditSlashCommandOptions,
-  guildID: string,
+  guildID?: string,
 ) {
   // Use ... for content length due to unicode characters and js .length handling
   if ([...options.name].length < 2 || [...options.name].length > 32) {
@@ -261,6 +261,7 @@ export function editSlashCommand(
       : endpoints.COMMANDS_ID(applicationID, commandID),
     options,
   );
+
   return result;
 }
 
--- a/src/api/handlers/webhook.ts
+++ b/src/api/handlers/webhook.ts
@@ -239,15 +239,18 @@ export function upsertSlashCommand(options: UpsertSlashCommandOptions) {
 }
 
 /** Edit an existing slash command. */
-export function editSlashCommand(options: EditSlashCommandOptions) {
-  return RequestManager.patch(
-    options.guildID
-      ? endpoints.COMMANDS_GUILD_ID(applicationID, options.id, options.guildID)
-      : endpoints.COMMANDS_ID(applicationID, options.id),
-    {
-      ...options,
-    },
+export function editSlashCommand(
+  commandID: string,
+  options: EditSlashCommandOptions,
+  guildID: string,
+) {
+  const result = RequestManager.patch(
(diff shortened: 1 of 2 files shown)


src/api/handlers/webhook.ts
```diff
@@ -227,27 +227,69 @@ export function getSlashCommands(guildID?: string) {
...
-    options.guildID
-      ? endpoints.COMMANDS_GUILD_ID(applicationID, options.id, options.guildID)
-      : endpoints.COMMANDS_ID(applicationID, options.id),
-    {
-      ...options,
-    },
+// TODO: remove this function for v11
+/** 
+ * Edit an existing slash command. 
+ * @deprecated This function will be removed in v11. Use `upsertSlashCommand()` instead
```
ayntee, line 263: The endpoints still exists:
- https://discord.com/developers/docs/interactions/slash-commands#edit-global-application-command
- https://discord.com/developers/docs/interactions/slash-commands#edit-guild-application-command

itohatweb: Yes these exist and are covered by `upsertSlashCommand` imo `upsertSlashCommand` has a more descriptive name of what it is doing so that one should be kept.
Also theoretically `editSlashCommand` could be removed already without being a real breaking change since nobody uses it.

ayntee: > Yes these exist and are covered by upsertSlashCommand imo upsertSlashCommand has a more descriptive name of what it is doing so that one should be kept.
The usage of editSlashCommand() is different from that of upsertSlashCommand(), what if the user does not want to edit the slash command if the command does not exist? `upsertSlashCommand()` will create the command regardless and is a hacky solution... we should support both of the functions.
> ...since nobody uses it.
How do you know that?

itohatweb: Well `editSlashCommand` is broken if someone would use that he couldn't and therefor would either ask for help or just not use it.

ayntee: > Well `editSlashCommand` is broken if someone would use that he couldn't and therefor would either ask for help or just not use it.

Even if the handler has a bug does not mean that we should just remove it, does it?

itohatweb: Well rn you cannot do anything with this function. Rn it only accepts the `commandID` and the `guildID` as option but nothing that could be used to edit the command.

ayntee: You can add the new properties to the `EditSlashCommandOptions` and not deprecate the handler as well as interface.

itohatweb: I have done that. In this PR `editSlashCommand` does work. But it is completely unnecessary due to `upsertSlashCommand` therefore it is redundant and should be removed.

ayntee: As I said, they both have different usage.

itohatweb: No, they both have the exact same code and do the same request.

itohatweb: There is no `edit` according to the discord documentation. If the command exists it gets edited if not gets created. That's what `upsertSlashCommand` and `editSlashCommand` does. `upsertSlashCommand` has a better more descriptive name so this should be kept and `editSlashCommand` should be removed.

ayntee: Hmm, just took a look at the code, I seem to be mistaken, but I think editSlashCommand() is a better-suiting name than upsertSlashCommand(), therefore, you should deprecate the `upsertSlashCommand()~ instead of `editSlashCommand()`.

itohatweb: > Because Slash Commands have unique names within a scope, we treat POST requests for new commands as `upserts`

ayntee (approved):

PR #443, opened 2021-01-23
fix(controllers): applicationID may be undefined set fallback to botID

itohatweb:

**Status**

- [x] Code changes have been tested against the Discord API, or there are no
  code changes

  - [ ] This PR includes breaking changes (methods removed or renamed,
    parameters moved or removed)
  README, etc.

### DIFF
--- a/src/api/controllers/misc.ts
+++ b/src/api/controllers/misc.ts
@@ -24,7 +24,7 @@ export async function handleInternalReady(
 
   const payload = data.d as ReadyPayload;
   setBotID(payload.user.id);
-  setApplicationID(payload.application.id);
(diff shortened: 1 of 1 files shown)


ayntee:
According to the documentation, the "application" field in the ReadyPayload is NEVER undefined or null. `ReadyPayload.application` is a partial application object, which always includes the properties "id" and "flags".

itohatweb:
It can be undefined according to my tests

itohatweb:
In my test the application field did not exist therefor the `applicationID` var was undefined. So the slash command endpoints did not work. I have set the botID as fall-back and then everything has worked. The documentation is wrong in this case.

PR #446, opened 2021-01-24
fix: valid type "member" in "interactionCreate" event

ayntee:


src/api/controllers/interactions.ts
```diff
@@ -4,17 +4,13 @@ import {
...
 
-export async function handleInternalInteractionCreate(data: DiscordPayload) {
+export function handleInternalInteractionCreate(data: DiscordPayload) {
   if (data.t !== "INTERACTION_CREATE") return;
 
   const payload = data.d as InteractionCommandPayload;
   eventHandlers.interactionCreate?.(
-    {
-      ...payload,
-      member: await structures.createMember(payload.member, payload.guild_id),
```
itohatweb, line 16: Why remove member from event?

src/api/structures/guild.ts
```diff
@@ -210,7 +210,7 @@ export async function createGuild(data: CreateGuildPayload, shardID: number) {
 
   initialMemberLoadQueue.set(guild.id, members);
 
-  return guild;
+  return guild as Guild;
```
itohatweb, line 213: Irrelevant change, should be done in a other PR.

src/api/structures/member.ts
```diff
@@ -126,7 +126,7 @@ export async function createMember(data: MemberCreatePayload, guildID: string) {
 
   await cacheHandlers.set("members", member.id, member);
 
-  return member;
+  return member as Member;
```
itohatweb, line 129: Same as above

src/api/structures/template.ts
```diff
@@ -40,7 +40,7 @@ export function createTemplate(
     sourceGuildID: createNewProp(sourceGuildID),
     serializedSourceGuild: createNewProp(serializedSourceGuild),
     isDirty: createNewProp(isDirty),
-  });
+  }) as Template;
```
itohatweb, line 43: Same as above

itohatweb (changes requested):

src/api/controllers/interactions.ts
ayntee: `member` still exists in the data, but only the type and value is altered.

itohatweb: Well yes member still exists but member does not get cached into memory. Also all other events either create the member or get it from cache. So why removing it here?

ayntee: The type of the first parameter of the "interactionCreate" event is `InteractionCommandPayload`, it includes the "member" property, which is of type `MemberCreatePayload`, **not** `Member`; however, the `createMember()` function resolves to the type `Member`, consequently conflict arises, to resolve this conflict, the raw member payload is passed to the "interactionCreate" event.

itohatweb: Then interaction create event should return the interaction and the member imo.

ayntee: Currently, the "interactionCreate" event has one parameter *data* which is of type `InteractionCommandPayload`, however, if I were to remove this, the type "Member" and "MemberCreatePayload" would result in errors.

itohatweb: What I mean is not remove it but add a second arg for the member.

Issue #462, opened 2021-01-25
Membership Pending Passed Event


Skillz4Killz:
It would be nice to have a special event that is triggered when a member finished the membership gating. We have a lot of nice helper events like roleAdded roleRemoved voiceChannelJoin voiceChannelLeave and i believe this would be a nice one to  add as well. I imagine a lot of bots will want to handle autoroles or such and this event would be really a great help instead of the end user having to handle it themself.

PR #467, opened 2021-01-25
feat(controllers): add membershipScreeningPassed event

ayntee:
- Added "membershipScreeningPassed" event to `EventHandlers`.
- Emit "membershipScreeningPassed" event when the `member.pending` field is true in "GUILD_MEMBER_UPDATE" event.

Closes: #462

### DIFF
--- a/src/api/controllers/members.ts
+++ b/src/api/controllers/members.ts
@@ -81,7 +81,7 @@ export async function handleInternalGuildMemberUpdate(data: DiscordPayload) {
     );
   }
 
-  if (guildMember?.pending === false) {
+  if (guildMember?.pending === false && cachedMember?.pending) {
     eventHandlers.membershipScreeningPassed?.(guild, member);
   }
 
--- a/src/api/controllers/members.ts
+++ b/src/api/controllers/members.ts
@@ -81,7 +81,7 @@ export async function handleInternalGuildMemberUpdate(data: DiscordPayload) {
     );
   }
 
-  if (guildMember?.pending === false && cachedMember?.pending) {
+  if (payload.pending === false && guildMember?.pending === true) {
     eventHandlers.membershipScreeningPassed?.(guild, member);
   }
 
--- a/src/api/controllers/members.ts
+++ b/src/api/controllers/members.ts
@@ -81,7 +81,7 @@ export async function handleInternalGuildMemberUpdate(data: DiscordPayload) {
     );
   }
 
-  if (guildMember?.pending) {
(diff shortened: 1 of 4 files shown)


src/api/structures/member.ts
```diff
@@ -156,6 +156,8 @@ export interface Member {
   premiumType?: number;
   /** The guild related data mapped by guild id */
   guilds: Collection<string, GuildMember>;
+  /** whether the user has not yet passed the guild's Membership Screening requirements */
+  pending?: boolean;
```
itohatweb, line 160: This should not be there since members are cached separately from guilds. Means if this is true the bot would think the member is pending on all guilds he is in. The pending parameter should be set in GuildMember (`member.guilds`)

itohatweb (changes requested):

src/api/controllers/members.ts
```diff
@@ -80,6 +80,11 @@ export async function handleInternalGuildMemberUpdate(data: DiscordPayload) {
       guildMember?.nick,
     );
   }
+
+  if (guildMember?.pending === false && cachedMember?.pending) {
```
itohatweb, line 84: There only guildMember should be checked.

DjDeveloperr: It is always present, if there is only check for guildMember?.pending, it'll fire on even nickname changes.

itohatweb: That check should be improved then since `cachedMember` should not have the property `.pending`

itohatweb: Due to memory efficiency members are cached independent to guilds in DD. So if the .pending parameter would exist in `cachedMemeber` that would mean that the member is pending on all guilds which is not good.

Skillz4Killz, line 84:
```suggestion
  if (payload.pending === false && guildMember?.pending === true) {
```

src/api/structures/member.ts
Skillz4Killz: This should be moved to GuildMember

Skillz4Killz (changes requested):

src/api/controllers/members.ts
itohatweb: Yes that's better but the `=== true` is not necessary I think.

src/api/structures/member.ts
itohatweb, line 160: You still need to remove that one and add it to the GuildMember interface
```suggestion
```

Skillz4Killz (approved):

PR #478, opened 2021-01-27
refactor: better name for parameters & functions

### LINKED ISSUE #455: Better name for executeSlashCommand
Imo `executeSlashCommand` is a bad name and other functions are named eg. `deleteSlashResponse`. So imo we should find a better name for that.

### LINKED ISSUE #476: slowmode -> rateLimitPerUser
Replace all instances and occurrences of `slowmode` with `rateLimitPerUser`, as previously discussed. Now you might have a question, why is this necessary?

This change is proposed to mainly avoid confusion between the library and the API. Additionally, we should be consistent with the API and try to follow their spec wherever we can as well as with our code. Besides, majority of Discordeno users migrate from Discord.js which uses `rateLimitPerUser`. Feel free to leave your thoughts in the comments.

ayntee:
Closes #476 
Closes #455


src/api/handlers/channel.ts
```diff
@@ -467,8 +467,9 @@ export async function editChannel(
 
   const payload = {
     ...options,
+    reason,
     // deno-lint-ignore camelcase
-    rate_limit_per_user: options.slowmode,
+    rate_limit_per_user: options.rateLimitPerUser,
```
itohatweb, line 472: Here you can use the camel to snake case function.

ayntee: If only it existed. As I said, that PR was reverted. :sweat_smile:

itohatweb: Again? lol why? All tests are passing your/gh deno is weird.

ayntee: > Again? lol why? All tests are passing your/gh deno is weird. Even on my linux machine tests are passing.

No idea, but it won't be merged unless the official GitHub checks are passing, that's for sure.

itohatweb: I give up as long as I don't know whats the problem, I cannot solve it.

ayntee: > I give up as long as I don't know whats the problem, I cannot solve it.

Anything I can help with? If not, that's fine too

itohatweb: If tests are failing deno should log what went wrong. So in this case what parameter was expected and what deno got. Would be helpful if you can send me that.

Skillz4Killz (approved):

ayntee:
Separate this PR into multiple PRs for better changelog and commit history

PR #481, opened 2021-01-28
fix(util/permissions): include guild ID while resolving bits

### LINKED ISSUE #480: permission bits should also include @everyone role ID
If and only if the bot has the `@everyone` role, the `memberHasPermission()` function regards that as bot has no permissions. We should fix this.

ayntee:
Closes #480


itohatweb (approved):
LGTM

Issue #485, opened 2021-01-28
Role delete controller is emitting two role delete events


itohatweb:
So if a role gets deleted, the role delete controller emits two role delete events instead of just one.

PR #486, opened 2021-01-28
fix(controllers/roles): emit roleDelete only once

itohatweb:
Closes: #485


src/api/controllers/roles.ts
```diff
@@ -29,6 +29,9 @@ export async function handleInternalGuildRoleDelete(data: DiscordPayload) {
   if (!guild) return;
 
   const cachedRole = guild.roles.get(payload.role_id)!;
```
ayntee:
```suggestion
  const cachedRole = guild.roles.get(payload.role_id);
```

ayntee (approved):
LGTM

PR #493, opened 2021-01-30
Adding toJSON function on util/collection

JeanCarlosVR:
Add the function to convert all Map values to JSON, this can be very useful.

### DIFF
--- a/src/util/collection.ts
+++ b/src/util/collection.ts
@@ -88,7 +88,7 @@ export class Collection<K, V> extends Map<K, V> {
   }
 
   toJSON() {
-    const _json: any = {};
+    const _json: {[k: any]: any} = {};
 
     for (const key of this.keys()) {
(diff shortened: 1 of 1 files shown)


itohatweb:
What if it is a Collection of a Collection

JeanCarlosVR:
> What if it is a Collection of a Collection. Also pls explain the use cases for this.



> What if it is a Collection of a Collection. Also pls explain the use cases for this.

The person may not need those collections within the collection to be converted to JSON; They are also converted to JSON, and if you need it you can simply call that collection that is inside and execute the same converting it into JSON, Personally this is more flexible to the needs.

itohatweb:
image

JeanCarlosVR (commented):
Fixing

itohatweb:
> Fixing

Still failing

PR #509, opened 2021-02-06
fix(types/channel): change FileContent.blob type to Blob

### LINKED ISSUE #506: Missing annotation and/or documentation for file attachments
The current documentation describes MessageContent.file.blob as `unknown`, which is absolutely useless. The website <https://discordeno.mod.land/> also doesn't contain any documentation on how to add attachments to a message.

Including something like `file: {blob: Deno.readFileSync("some_file.whatever"), name: "some_file.whatever"}` in MessageContent causes a `The request was improperly formatted, or the server couldn't understand it.` error.

Since the option exists, I'm sure there's a particular type that this function accepts so that file attachments work, but, the lack of documentation means I have no way o figuring this argument out.

So, what should be the type of `MessageContent.file.blob`? And why isn't this type described in the MessageContent description?

ayntee:
Closes #506


itohatweb (approved):
LGTM

Skillz4Killz (approved):

PR #487, opened 2021-01-29
feat(handlers/guild): better return type for getAuditLogs()

itohatweb:

PR #510, opened 2021-02-06
feat(types): add and detach API types

Skillz4Killz:


src/types/api/oauth2.ts
```diff
@@ -0,0 +1,37 @@
...
+  guildID?: string;
+  /** if this application is a game sold on Discord, this field will be the id of the "Game SKU" that is created, if exists */
+  primarySkuID?: string;
+  /** if this application is a game sold on Discord, this field will be the URL slug that links to the store page */
+  slug?: string;
+  /** if this application is a game sold on Discord, this field wil be the hash of the image on store embeds */
+  coverImage?: string;
+  /** the application's public flags */
+  flags: number;
+}
```
ayntee, line 37: This uses camel case, but is marked as a Discord API type, can you recheck this?

src/types/cdn.ts
```diff
@@ -1,2 +1,6 @@
+// TODO: v11 Remove
+/** @deprecated Use DiscordImageSize */
 export type ImageSize = 16 | 32 | 64 | 128 | 256 | 512 | 1024 | 2048;
+// TODO: v11 Remove
+/** @deprecated Use DiscordImageFormat */
```
ayntee, line 5: The API types are for internal use, it is not recommended to use the types exclusively... so we can just delete the entire file?

src/types/util.ts
```diff
@@ -6,4 +6,4 @@ export type CamelizeString<T extends PropertyKey> = T extends string
   : T
   : T;
 
-export type Camelize<T> = { [K in keyof T as CamelizeString<K>]: T[K] }
+export type Camelize<T> = { [K in keyof T]: T[K] };
```
ayntee, line 9: This is wrong, if `CamelizeString<K>` is removed, it will not convert the property into camel case.

src/types/api/auditlog.ts
```diff
@@ -0,0 +1,217 @@
...
+  /** name of the role if type is "0" (not present if type is "1"), types: CHANNEL_OVERWRITE_CREATE & CHANNEL_OVERWRITE_UPDATE & CHANNEL_OVERWRITE_DELETE */
+  role_name: string;
+}
+
+/** https://discord.com/developers/docs/resources/audit-log#audit-log-change-object-audit-log-change-structure */
+export interface DiscordAuditLogChangePayload {
+  /** new value of the key */
+  new_value?: DiscordAuditLogChangeKeyPayload;
+  /** old value of the key */
+  old_value?: DiscordAuditLogChangeKeyPayload;
```
itohatweb, line 104: These types are wrong. You can use this files as reference https://github.com/discordeno/discordeno/pull/487/files

src/types/api/auditlog.ts
```diff
@@ -0,0 +1,217 @@
...
+  id: string;
+  /** object: any; type of entity created */
+  type: DiscordChannelTypes | string;
+  /** object: integration; integration emoticons enabled/disabled */
+  enable_emoticons: boolean;
+  /** object: integration; integration expiring subscriber behavior changed */
+  expire_behavior: number;
+  /** object: integration; integration expire grace period changed */
+  expire_grace_period: number;
+}
```
itohatweb, line 205: This is not correct (https://github.com/discordeno/discordeno/pull/487/files)

itohatweb (changes requested):

itohatweb:
This linter requires snake cased params in interfaces to be a string.
Eg.
```ts
interface DiscordType {
  "some_snake_case": string
}
```

(tbh this is pretty annoying and if that cannot be deactivated we should remove the linter imo)

ayntee (changes requested):

src/types/api/auditlog.ts
ayntee: We should just remove the types that are covered in #487 already from this PR

itohatweb: No we cannot to this since this are two different files

ayntee: Does not matter, this type is not used in production, so it should be fine

itohatweb: This file here is in `src/types/api/auditlog.ts` and in #487 it is `src/types/guild.ts` so it does matter

ayntee: And?

itohatweb: So we cannot remove the types from here and merge the other PR.

ayntee: Your reason?

itohatweb: wdym

ayntee: What is the reason as to why this cannot be removed from this PR?

itohatweb: As said #487 edits `src/types/guild.ts` but in this PR it is `src/types/api/auditlog.ts` This PR is here to improve the API types then having some audit log types in guild.ts again does not make sense

src/types/cdn.ts
Skillz4Killz: IMO we can keep this as it's not really causing any difficulty/issues and can easily be nuked once we ready for v11. Before we do it we just search TODO: v11 and we will find all the stuff we want done for it.

This prevents even the tiniest of breaking change and tbh i would love for a librarby to handle breaking changes with this much care. Even the smallest of breaks is a breaking change. Atm nothing on this is marking it private and therefore technically speaking it is open to the public api.

itohatweb (approved):
LGTM

ayntee (approved):

PR #512, opened 2021-02-09
fix(controllers/guild): guildDelete listener not emitted & guildUpdate listener caching irrelevant fields

### LINKED ISSUE #511: DD never emits Guild Delete Event
DD never emits the guild delete event. This is due following issue: The guild is deleted from cache and then there is an if check which only lets the guild delete event to be dispatched if the guild is cached.

itohatweb:
Closes #511


ayntee (approved):
Thank you

PR #515, opened 2021-02-10
refactor: resolve promises

TriForMine:


src/rest/request.ts
```diff
@@ -52,7 +52,7 @@ export function createRequestBody(queuedRequest: QueuedRequest) {
   // IF A REASON IS PROVIDED ENCODE IT IN HEADERS
   if (queuedRequest.payload.body?.reason) {
     headers["X-Audit-Log-Reason"] = encodeURIComponent(
-      queuedRequest.payload.body.reason,
+      queuedRequest.payload.body?.reason,
```
ayntee, line 55: This is unnecessary change.

ayntee:
Can you spin `deno lint src/** test/** --unstable --ignore=./src/types/`?

src/rest/request.ts
TriForMine: Reverted

src/ws/shard.ts
```diff
@@ -15,8 +15,9 @@ import {
 import { BotStatusRequest, delay } from "../util/utils.ts";
 import { decompressWith } from "./deps.ts";
 import { handleDiscordPayload } from "./shard_manager.ts";
+import { Collection } from "../util/collection.ts";
 
-const basicShards = new Map<number, BasicShard>();
+const basicShards = new Collection<number, BasicShard>();
```
ayntee, line 20: New Collection, why? We don't really need extra methods

TriForMine: It's needed for Promise.all at this line: https://github.com/TriForMine/discordeno/blob/70a3eef41aac9407c33e940584f584de8d6f605f/src/ws/shard.ts#L331
I'm not sure if there is an other way to do it.

src/ws/shard.ts
```diff
@@ -327,7 +328,7 @@ async function processGatewayQueue() {
     return;
   }
 
-  basicShards.forEach((shard) => {
+  await Promise.all(basicShards.map(async (shard) => {
```
ayntee, line 331: This is unnecessary, since we are not really returning a promise from the .map()

TriForMine: It's actually needed since to handle the rate limit we need to wait until everything is done and then wait 2 seconds if I'm not wrong.

src/ws/shard_manager.ts
```diff
@@ -40,7 +40,7 @@ export async function spawnShards(
       data.shards > lastShardID ? data.shards : lastShardID,
     ];
     // Start The shard
-    await createShard(data, payload, false, shardID);
+    createShard(data, payload, false, shardID);
```
ayntee, line 43: `await` removed, can you explain?

TriForMine: createShard isn't an async function 🤔

src/ws/shard_manager.ts
```diff
@@ -99,7 +99,15 @@ export function requestAllMembers(
...
-  return requestGuildMembers(guild.id, guild.shardID, nonce, options);
+
+  const result = await requestGuildMembers(
+    guild.id,
+    guild.shardID,
+    nonce,
+    options,
+  );
+
+  return result;
```
ayntee, line 110:
```suggestion
  await requestGuildMembers(
    guild.id,
    guild.shardID,
    nonce,
    options,
  );
```
Not returning anything also returns undefined, so basically the assignment is redundant.

ayntee (approved):
LGTM

test/deps.ts
```diff
@@ -2,5 +2,5 @@ export {
   assertArrayIncludes,
   assertEquals,
   assertExists,
-} from "https://deno.land/std@0.85.0/testing/asserts.ts";
+} from "https://deno.land/std@0.86.0/testing/asserts.ts";
```
ayntee, line 5: Can you remove this change?

TriForMine (commented):
_ignore this_

ayntee (approved):
LGTM!

PR #523, opened 2021-02-14
feat(structures/message): add DM redirect for send message methods

### LINKED ISSUE #520: `sendMessage` use `sendDirectMessage` as fall back
Currently `sendMessage` does not work with DMs if the channel is not cached. We should add a fall back to `sendDirectMessage` in case the dm channel isn't cached.

Skillz4Killz:
Closes #520

### DIFF
--- a/src/api/structures/message.ts
+++ b/src/api/structures/message.ts
@@ -15,6 +15,7 @@ import { cache } from "../../util/cache.ts";
 import { createNewProp } from "../../util/utils.ts";
 import { cacheHandlers } from "../controllers/cache.ts";
 import { sendMessage } from "../handlers/channel.ts";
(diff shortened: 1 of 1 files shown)


Skillz4Killz:
@itohatweb @ayntee review plz

src/api/structures/message.ts
```diff
@@ -32,7 +33,8 @@ import { Role } from "./role.ts";
 
 const baseMessage: Partial<Message> = {
   get channel() {
-    return cache.channels.get(this.channelID!);
+    return cache.channels.get(this.channelID!) ||
+      cache.channels.get(this.author?.id!);
```
ayntee, line 37: Can you add a note to the JSDoc comment block for this header regarding the change?

ayntee (approved):
LGTM

Issue #527, opened 2021-02-17
createGuildChannel() should require a guild ID


itohatweb:
The `createGuildChannel()` handler requires a `Guild` object as an argument as of now, we should change that to use ID of the guild; internally, the handler only uses the `guild.id` property.

itohatweb:
On it

PR #528, opened 2021-02-17
feat(handlers/guild): change createGuildChannel() to use guild ID

itohatweb:
closes: #527

### DIFF
--- a/test/mod.test.ts
+++ b/test/mod.test.ts
@@ -149,7 +149,7 @@ Deno.test({
   name: "[channel] create a channel in a guild",
   async fn() {
 
-    const channel = await createGuildChannel(guild.id, "test");
+    const channel = await createGuildChannel(tempData.guildID, "test");
 
     // Assertions
     assertExists(channel);
--- a/test/mod.test.ts
+++ b/test/mod.test.ts
@@ -148,7 +148,6 @@ Deno.test({
 Deno.test({
   name: "[channel] create a channel in a guild",
   async fn() {
-
     const channel = await createGuildChannel(tempData.guildID, "test");
 
     // Assertions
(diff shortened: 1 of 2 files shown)


test/mod.test.ts
```diff
@@ -151,7 +151,7 @@ Deno.test({
     const guild = cache.guilds.get(tempData.guildID);
     if (!guild) throw new Error("Guild not found");
 
-    const channel = await createGuildChannel(guild, "test");
+    const channel = await createGuildChannel(guild.id, "test");
```
ayntee, line 154:
```suggestion
    const channel = await createGuildChannel(tempData.guildID, "test");
```

test/mod.test.ts
```diff
@@ -151,7 +151,7 @@ Deno.test({
     const guild = cache.guilds.get(tempData.guildID);
     if (!guild) throw new Error("Guild not found");
```
ayntee, line 152:
```suggestion
```

test/mod.test.ts
```diff
@@ -148,10 +148,8 @@ Deno.test({
 Deno.test({
   name: "[channel] create a channel in a guild",
   async fn() {
-    const guild = cache.guilds.get(tempData.guildID);
-    if (!guild) throw new Error("Guild not found");
```
itohatweb, line 152: why removing that? there we can check indirectly if guild cache is working.

ayntee: The test case is not to check whether or not, cache is working.

itohatweb: then we should add that in another pr

Skillz4Killz (approved):

ayntee (approved):

PR #531, opened 2021-02-20
refactor: use updated identify payload instead

ayntee:


src/bot.ts
```diff
@@ -29,18 +30,6 @@ export const identifyPayload: IdentifyPayload = {
   shard: [0, 0],
 };
 
-export interface IdentifyPayload {
```
Skillz4Killz, line 32: Bring this back and mark it deprecated to make this pr non-breaking

ayntee: @Skillz4Killz the types are not supposed to be private, hence not supposed to be used by the users, but internally.

Skillz4Killz: That doesnt mean its not part of the public api. Realistically, you correct. Technically, this is still breaking.

ayntee: > That doesnt mean its not part of the public api. Realistically, you correct. Technically, this is still breaking.

I don't really mind, but just putting that out there...

src/bot.ts
```diff
@@ -29,18 +30,6 @@ export const identifyPayload: IdentifyPayload = {
   shard: [0, 0],
 };
```
ayntee, line 32:
```suggestion
/** deprecated Use "DiscordIdentify" instead */
export interface IdentifyPayload {
  token: string;
  compress: boolean;
  properties: {
    $os: string;
    $browser: string;
    $device: string;
  };
  intents: number;
  shard: [number, number];
}
```

src/bot.ts
```diff
@@ -28,7 +29,7 @@ export const identifyPayload: IdentifyPayload = {
   intents: 0,
   shard: [0, 0],
 };
```
ayntee, line 31:
```suggestion
};

```

src/bot.ts
```diff
@@ -29,6 +30,7 @@ export const identifyPayload: IdentifyPayload = {
...
 
+/** deprecated Use "DiscordIdentify" instead */
```
ayntee, line 33:
```suggestion
/** @deprecated Use "DiscordIdentify" instead */
```

src/bot.ts
```diff
@@ -40,7 +42,6 @@ export interface IdentifyPayload {
   intents: number;
   shard: [number, number];
 }
```
ayntee, line 44:
```suggestion
}

```

PR #537, opened 2021-02-21
fix(rest): proxy

itohatweb:


src/rest/types/server.ts
```diff
@@ -9,6 +9,8 @@ export interface RestServerOptions {
   token: string;
   /** When a request is rate limited, how many times should it keep retrying the request. Recommended: 10 */
   maxRetryCount: number;
+  /** The Version of the Discord API which the requests should be forwarded to. */
+  apiVersion?: "v6" | "v7" | "v8";
```
Skillz4Killz, line 13: Can we make this a `number` and use `v` in the url itself? That would allow this to be much more dynamic.

itohatweb: then ima change it to a number

PR #547, opened 2021-02-25
Add 401 if the authorization header is not provided

TriForMine:


itohatweb:
I am not sure if we should do this. Yes it is good to have but we did it to prevent attacks. I don't know what you have discussed on dc though so if skillz wants it I have nothing against.

PR #549, opened 2021-02-26
fix(rest/queue): logical error in response status handling

### LINKED ISSUE #546: Response status checking for readable errors will never be true
Relevant code: https://github.com/discordeno/discordeno/blob/997189ec4af56a362c6460471d808a7ad9c2a574/src/rest/queue.ts#L65

I believe this should be a `||` (though `< 200` does not feel necessary, given 1xx http codes are not *technically* an error).

ayntee:
Closes #546

### DIFF
--- a/src/rest/queue.ts
+++ b/src/rest/queue.ts
@@ -62,7 +62,7 @@ export async function processQueue(id: string) {
         response.headers,
       );
 
-      if (response.status < 200 && response.status >= 400) {
+      if (response.status < 200 || response.status >= 400) {
         restCache.eventHandlers.error(
           "httpError",
           queuedRequest.payload,


src/rest/queue.ts
```diff
@@ -62,7 +62,7 @@ export async function processQueue(id: string) {
         response.headers,
       );
 
-      if (response.status < 200 && response.status >= 400) {
+      if (response.status < 200 || response.status >= 400) {
```
itohatweb, line 65: I like this more if you have nothing against changing it.
```suggestion
      if (response.status < 200 || 400 <= response.status) {
```

itohatweb (approved):
Then merge it

ayntee:
> Then merge it

image

itohatweb:
I guess Skillz needs to approve it

Skillz4Killz (approved):

PR #551, opened 2021-02-26
fix(structures/channel): empty guildID in Channel struct

itohatweb:


ayntee:
This will not completely fix the issue

itohatweb:
I just did quick testing on start rn.

ayntee (approved):

PR #596, opened 2021-03-06
feat(types/message): add fail_if_not_exists to Reference interface

### LINKED ISSUE #581: Support new failthrough option for replies
fail_if_not_exists

https://github.com/discord/discord-api-docs/commit/84104023837d038fe6dff4674e2f650b32760a99

ayntee:
Closes #581


src/types/message.ts
```diff
@@ -377,3 +384,19 @@ export enum MessageStickerFormat {
...
+   * ID of the originating message's channel
+   * 
+   * NOTE: `channel_id` is optional when creating a reply, but will always be present when receiving an event/response that includes this data model
+   */
+  channel_id?: string;
+  /** ID of the originating message's guild */
+  guild_id?: string;
+  /** When sending, whether to error if the referenced message doesn't exist instead of sending as a normal (non-reply) message, default true */
+  fail_if_not_exists?: boolean;
+}
```
ayntee, line 402:
```suggestion
```

src/types/message.ts
```diff
@@ -196,13 +196,20 @@ export interface Application {
...
-  /** The id of the originating message's channel */
-  "channel_id": string;
-  /** The id of the originating message's guild */
-  "guild_id"?: string;
+  /** ID of the originating message */
+  message_id?: string;
+  /**
+   * ID of the originating message's channel
+   * 
+   * NOTE: `channel_id` is optional when creating a reply, but will always be present when receiving an event/response that includes this data model
```
Skillz4Killz, line 206: This can be handled better imo, we just need to split this into 2 interfaces.

export interface ReferenceBase {

}

export interface Reference extends ReferenceBase {
  channel_id: string;
}

src/types/message.ts
```diff
@@ -196,13 +196,20 @@ export interface Application {
...
+   */
+  channel_id?: string;
+  /** ID of the originating message's guild */
+  guild_id?: string;
+  /** When sending, whether to error if the referenced message doesn't exist instead of sending as a normal (non-reply) message, default true */
+  fail_if_not_exists?: boolean;
```
Skillz4Killz, line 212: Please make this default to false in the code! We should not allows bots to fail and crash!

Skillz4Killz (changes requested):

src/types/message.ts
itohatweb: keep in mind if you set it to false here you won't be able to set it to true.

ayntee: > keep in mind if you set it to false here you won't be able to set it to true.
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

src/types/message.ts
ayntee: Can you explain as to why this is necessary?

src/types/message.ts
Skillz4Killz: No i meant in the code, the function itself that handles this.

```ts
{
...options,
fail_if_not_exists: options.failIfNotExists === true
}
```

src/types/message.ts
Skillz4Killz: For a few reasons:
1. accuracy of typings although this isnt that important
2. user experience. with the current solution u have to do `if (data.channel_id)` to make sure a channel id is present before using it but if this typing is more detailed when a user gets it inside an interaction they wont have to leading to cleaner code

Skillz4Killz (changes requested):

src/types/message.ts
ayntee: What? That's equivalent of passing it directly

itohatweb (approved):

PR #600, opened 2021-03-06
feat(types/auditlog): add missing props to DiscordAuditLogChange interface

### LINKED ISSUE #582: Better Audit Log handling & new typings
If `new_value` is not present in the change object, while `old_value` is, that means the property that was changed has been reset, or set to `null`

https://github.com/discord/discord-api-docs/commit/1bf244cc650b482cc1579519fbf935e933c18046

Skillz4Killz:
Closes #582


ayntee:
Can you move this to the `./types` directory and not `./types/api`? We should gradually move all of the types from `types/api` to main directory and eventually, remove `types/api` entirely.

ayntee (approved):
LGTM

Issue #593, opened 2021-03-06
Add resolved objects in interactions


itohatweb:
Pending: https://github.com/discord/discord-api-docs/pull/2630

Skillz4Killz:
This PR was recently merged, i think this should already be done in discordeno but just wanted to confirm please

Skillz4Killz:
Yea i think #603 added it in

PR #603, opened 2021-03-06
feat(types/interaction): add "resolved" field to DiscordInteractionCommand

itohatweb:
Closes: #593

### DIFF
--- a/src/types/api/interaction.ts
+++ b/src/types/api/interaction.ts
@@ -36,6 +36,8 @@ export interface DiscordInteractionData {
   id: string;
   /** the name of the invoked command */
   name: string;
+  /** converted users + roles + channels */
+  resolved?: DiscordInteractionDataResolved;
   /** the params + values from the user */
   options: DiscordInteractionDataOption[];
 }
--- a/src/types/api/interaction.ts
+++ b/src/types/api/interaction.ts
@@ -39,7 +39,7 @@ export interface DiscordInteractionData {
   /** converted users + roles + channels */
   resolved?: DiscordInteractionDataResolved;
   /** the params + values from the user */
-  options: DiscordInteractionDataOption[];
+  options?: DiscordInteractionDataOption[];
 }
 
 export interface DiscordInteractionDataResolved {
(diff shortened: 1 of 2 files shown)


ayntee:
Can you confirm if that's the valid related issue?

itohatweb:
> Can you confirm if that's the valid related issue?

Oh sorry, GitHub auto completed my number...

src/types/api/interaction.ts
```diff
@@ -31,8 +36,24 @@ export interface DiscordInteractionData {
...
   /** the name of the invoked command */
   name: string;
+  /** converted users + roles + channels */
+  resolved?: DiscordInteractionDataResolved;
   /** the params + values from the user */
-  options: DiscordInteractionDataOption[];
+  options?: DiscordInteractionDataOption[];
+}
+
+export interface DiscordInteractionDataResolved {
```
ayntee, line 45:
```suggestion
export interface DiscordApplicationCommandInteractionDataResolved {
```
To be consistent with documentation

itohatweb: So 2 options:
1. Keep it as it is now
2. Name it `DiscordApplicationCommandInteractionDataResolved` but also rename all other interaction interfaces

Skillz4Killz: I would say option 3 is best, change this one real quick and then change others in another pr in future in v11

Skillz4Killz (changes requested):

src/types/api/interaction.ts
```diff
@@ -31,8 +36,24 @@ export interface DiscordInteractionData {
   id: string;
   /** the name of the invoked command */
   name: string;
+  /** converted users + roles + channels */
+  resolved?: DiscordInteractionDataResolved;
```
itohatweb, line 40:
```suggestion
  resolved?: DiscordApplicationCommandInteractionDataResolved;
```

Skillz4Killz (approved):

ayntee (approved):

PR #611, opened 2021-03-06
fix(handlers/webhook): return Message object instead of raw payload

### LINKED ISSUE #592: Edit Webhook Message Returns A Message Object On Success
Reference: https://github.com/discord/discord-api-docs/pull/2461

(Add code `50027`)

# PENDING

Skillz4Killz:
Closes #592


ayntee (dismissed):
LGTM

src/api/handlers/webhook.ts
```diff
@@ -552,5 +553,10 @@ export async function editSlashResponse(
     options,
   );
 
-  return result;
+  if (options.messageID) return result;
+
+  const message = await structures.createMessage(
+    result as MessageCreateOptions,
+  );
+  return message;
```
itohatweb, line 561: Are you sure that `Delete Original Interaction Response` returns the message?

Skillz4Killz: Github is confusing u this line if part of `editSlashResponse`. For quick confirmation check th eline numbers

itohatweb: Ah lol you're right

src/api/handlers/webhook.ts
```diff
@@ -552,5 +553,10 @@ export async function editSlashResponse(
     options,
   );
 
-  return result;
+  if (options.messageID) return result;
```
itohatweb, line 556: `Edit Webhook Message` does return a message object too, so why not using `structures.createMessage` here too?

Skillz4Killz: Good catch that should have been if (!options.messageID

src/api/handlers/webhook.ts
```diff
@@ -552,5 +553,11 @@ export async function editSlashResponse(
     options,
   );
 
-  return result;
+  // If the original message was edited, his will not return a message
```
itohatweb, line 556:
```suggestion
  // If the original message was edited, this will not return a message
```

itohatweb (changes requested):

ayntee (approved):
LGTM

itohatweb (approved):

PR #668, opened 2021-03-13
refactor(helpers): rename & replace deleteMessageByID() with deleteMessage()

ayntee:


src/helpers/messages/delete_message.ts
```diff
@@ -1,25 +1,28 @@
...
-/** Delete a message */
+/** Delete a message with the channel id and message id only. */
 export async function deleteMessage(
-  message: Message,
+  messageID: string,
   reason?: string,
   delayMilliseconds = 0,
 ) {
+  const message = await cacheHandlers.get("messages", messageID);
+  if (!message) throw new Error(Errors.MESSAGE_NOT_FOUND);
```
Skillz4Killz, line 16: A vast majority of the time u wont have the message cached. We should not throw this.

```ts
if (message) {
 // checks here
 if (message.author.id !== botID){}
}

```

src/types/errors.ts
```diff
@@ -19,6 +19,7 @@ export enum Errors {
   // Message Delete Errors
   DELETE_MESSAGES_MIN = "DELETE_MESSAGES_MIN",
   PRUNE_MIN_DAYS = "PRUNE_MIN_DAYS",
+  MESSAGE_NOT_FOUND = "MESSAGE_NOT_FOUND",
```
Skillz4Killz, line 22: Not really needed imo

src/helpers/messages/delete_message.ts
ayntee: We are also using `message.channelID` property... :thinking:

src/types/errors.ts
ayntee, line 22:
```suggestion

```

Skillz4Killz (approved):

PR #680, opened 2021-03-23
fix(util/collection): .first(), .last(), .random() may return undefined

ayntee:
Closes #680


itohatweb (approved):
lgtm

itohatweb:
Ah this is breaking

PR #685, opened 2021-03-23
refactor(rest): separate functions into diff files

Skillz4Killz:


src/rest/rest.ts
```diff
@@ -0,0 +1,46 @@
...
+import { createRequestBody } from "./create_request_body.ts";
+import { handlePayload } from "./handle_payload.ts";
+import { processQueue } from "./process_queue.ts";
+import { processRateLimitedPaths } from "./process_rate_limited_paths.ts";
+import { processRequest } from "./process_request.ts";
+import { processRequestHeaders } from "./process_request_headers.ts";
+import { RequestManager } from "./request_manager.ts";
+import { runMethod } from "./run_method.ts";
+
+export const rest: RestCache = {
```
ayntee, line 12: Shouldn't this be `let` so users can redeclare the variable?

ayntee (dismissed):

itohatweb (changes requested):
you need to call `processRateLimitedPaths()` somewhere else `globallyRateLimited` will always be true. After a rate limit. Also the rate limit cache will never be cleared.

src/rest/rest.ts
Skillz4Killz: Nope, u can edit a const {} without redeclaring

itohatweb: yes and no:
```ts
const foo = {
  bar: 20,
}

// works
foo.bar = 40;

// Cannot assign to 'foo' because it is a constant.
foo = {
  bar: 40,
}
```

src/rest/process_request_headers.ts
```diff
@@ -0,0 +1,59 @@
...
+    if (bucketID) {
+      rest.ratelimitedPaths.set(bucketID, {
+        url: "global",
+        resetTimestamp: reset,
+        bucketID,
+      });
+    }
+  }
+
+  if (!rest.processingRateLimitedPaths) rest.processRateLimitedPaths(false);
```
itohatweb, line 57:
```suggestion
  if (ratelimited && !rest.processingRateLimitedPaths) rest.processRateLimitedPaths(false);
```

itohatweb (approved):
then LGTM

itohatweb:
All conflicts resolved

src/rest/cleanup_queue.ts
```diff
@@ -0,0 +1,13 @@
...
+export function cleanupQueues() {
+  for (const [key, queue] of rest.pathQueues) {
+    if (queue.length) continue;
+    // REMOVE IT FROM CACHE
+    rest.pathQueues.delete(key);
+  }
+
+  // NO QUEUE LEFT, DISABLE THE QUEUE
+  if (!rest.pathQueues.size) rest.processingQueue = false;
+}
```
ayntee, line 13: Rename the name of the file to `cleanup_queues.ts`

ayntee (approved):

PR #720, opened 2021-03-31
types: add team types

itohatweb:
These types are required by the oauth types which are required for the Ready type

### DIFF
--- a/src/types/users/partial_user.ts
+++ b/src/types/users/partial_user.ts
@@ -1,5 +1,0 @@
-import { User } from "./user.ts";
-
-export type PartialUser =
-  & Partial<User>
-  & Pick<User, "avatar" | "discriminator" | "id" | "username">;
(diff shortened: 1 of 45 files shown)


src/types/users/partial_user.ts
```diff
@@ -0,0 +1,5 @@
+import { User } from "./user.ts";
+
```
ayntee, line 2: Missing link

itohatweb: link for?

ayntee: Any documentation link regarding the fields of the partial user object.

ayntee: In that case, you should just do `Partial<User>`

itohatweb: to be exact there is this but this is only for the team member object
image

ayntee: In that case, we should just go with your solution :stuck_out_tongue_closed_eyes:

ayntee (approved):

PR #726, opened 2021-03-31
fix(handlers/channels): edit logic

itohatweb:
- added missing imports
- removed processEditChannelQueue since it is strange


ayntee:
@Skillz4Killz

PR #736, opened 2021-04-03
types: fix gateway types

itohatweb:


src/types/gateway/identify.ts
```diff
@@ -12,11 +12,11 @@ export interface Identify {
   /** Value between 50 and 250, total number of members where the gateway will stop sending offline members in the guild member list */
   largeThreshold?: number;
   /** Used for Guild Sharding */
-  shard?: [number, number];
+  shard?: [shardId: number, numberOfShards: number];
```
Skillz4Killz, line 15: ????

ayntee: Basically, a tuple with named elements

PR #739, opened 2021-04-03
fix(structures/member): use new types & add MemberStruct interface

itohatweb:


src/structures/member.ts
```diff
@@ -6,10 +6,18 @@ import { rawAvatarURL } from "../helpers/members/raw_avatar_url.ts";
 import { sendDirectMessage } from "../helpers/members/send_direct_message.ts";
 import { addRole } from "../helpers/roles/add_role.ts";
 import { removeRole } from "../helpers/roles/remove_role.ts";
+import {
+  CreateGuildBan,
+  CreateMessage,
+  DiscordGuildMember,
+  GuildMember,
+  ModifyGuildMember,
+  User,
```
Skillz4Killz, line 15: Separate the imports

src/structures/member.ts
```diff
@@ -60,44 +68,35 @@ const baseMember: Partial<Member> = {
...
+  const props: Record<string, ReturnType<typeof createNewProp>> = {};
 
   for (const key of Object.keys(rest)) {
     // @ts-ignore index signature
-    restProps[key] = createNewProp(rest[key]);
+    props[key] = createNewProp(rest[key]);
   }
 
-  for (const key of Object.keys(user)) {
+  for (const key of Object.keys(user!)) {
```
Skillz4Killz, line 88: Are u certain?

src/structures/member.ts
```diff
@@ -60,44 +68,35 @@ const baseMember: Partial<Member> = {
...
-    ...restProps,
-    mfaEnabled: createNewProp(mfaEnabled),
-    premiumType: createNewProp(premiumType),
+    ...props,
     /** The guild related data mapped by guild id */
     guilds: createNewProp(new Collection<string, GuildMember>()),
   });
 
-  const cached = await cacheHandlers.get("members", user.id);
+  const cached = await cacheHandlers.get("members", user!.id);
```
Skillz4Killz, line 99: are u certain it will never be undefinewd here?

src/structures/member.ts
itohatweb: yes 
> The field `user` won't be included in the member object attached to `MESSAGE_CREATE` and `MESSAGE_UPDATE` gateway events.

src/structures/member.ts
itohatweb: yes 
> The field `user` won't be included in the member object attached to `MESSAGE_CREATE` and `MESSAGE_UPDATE` gateway events.

src/structures/member.ts
ayntee: **won't** :eye:

itohatweb: For the `MESSAGE_CREATE` event we do this: 
```ts
    const memberStruct = await structures.createMemberStruct(
      { ...payload.member, user: payload.author },
      guild.id,
    );
```
and for `MESSAGE_UPDATE` we don't create a member.

ayntee:
Also, please use a better PR title, reopen when done

src/structures/member.ts
```diff
@@ -106,13 +110,64 @@ export async function createMemberStruct(
...
     joinedAt: Date.parse(joinedAt),
     premiumSince: premiumSince ? Date.parse(premiumSince) : undefined,
-    deaf: deaf,
-    mute: mute,
+    deaf: props.deaf,
+    mute: props.mute,
   });
 
-  return member as Member;
+  return member as MemberStruct;
```
ayntee, line 121: Avoid using type assertions and try to use type casting instead.
```suggestion
  return member;
```

src/structures/member.ts
```diff
@@ -60,39 +71,32 @@ const baseMember: Partial<Member> = {
 };
 
 export async function createMemberStruct(
-  data: MemberCreatePayload,
+  data: Omit<DiscordGuildMember, "user"> & { user: DiscordUser },
```
ayntee, line 74: Add a comment explaining why this is done for future reference

itohatweb: Is this good?
`DiscordUser contains important informations like: id, username and discriminator which are required to build this structure`

ayntee, line 74: Why is this omitted and added back?

itohatweb: because user is optional in DiscordGuildMember

ayntee: You should add a comment explaining that, why it is optional non-optional for createMemberStruct, maybe a note from Discord docs.

itohatweb: https://github.com/discordeno/discordeno/pull/739/files/fb9ace6d11d790800a87a2280c99f25e732ec83d#diff-f3eedaf51b97c44f868e9f10f7e2f15d0948c19a362d90ab252fe00415074750

ayntee: Not why it is required, but why it is non-optional in this case and optional in the interface.

itohatweb: The `user` param in `DiscordGuildMember` is optional since discord does not send it in `MESSAGE_CREATE` and `MESSAGE_UPDATE` events. But this data in there is required to build this structure so it is required in this case?

ayntee (approved):

PR #753, opened 2021-04-04
fix(util/permissions): only check for permission if channel is cached

itohatweb:
As discussed on discord we decided it is the best to only validate the permissions if the according guild/channel is cached. 
Among other things this fixes the permission issue with DM channels.


ayntee:
Can you describe why this change is made in the PR body? Please abide by the contributing Guide for PR titles

Skillz4Killz (approved):

PR #759, opened 2021-04-06
fix(structures/template): add TemplateStruct back

itohatweb:


ayntee:
The entire structure is to be removed in v11

PR #769, opened 2021-04-06
types: add missing types

itohatweb:


src/types/discordeno/create_message.ts
```diff
@@ -0,0 +1,7 @@
+import { CreateMessage } from "../messages/create_message.ts";
+import { FileContent } from "../misc/file_content.ts";
+
+export interface DiscordenoCreateMessage extends CreateMessage {
+  /** The contents of the file being sent */
+  file?: FileContent | FileContent[];
+}
```
ayntee, line 7: You can update the `file` field in `CreateMessage` and then extend it to change the file property back in DiscordCreateMessage interface.

Skillz4Killz (approved):

PR #775, opened 2021-04-08
fix(types): fixing export paths

TriForMine:
Fix export since create_message was moved from ./channels to ./messages and integrations was moved to it's own folder


itohatweb:
we should just remove this file imo. all types exports can be in mod.ts

Skillz4Killz:
I think its good to have this file as it allows DD to basically have an entry point directly and only for typings. So if we need typings on another project we can simply import typings and not everything in dd

Skillz4Killz (approved):

PR #785, opened 2021-04-09
tests: add deleteChannel tests

Skillz4Killz:


tests/channels/create_channel.ts
```diff
@@ -0,0 +1,26 @@
...
+  async fn() {
+    const channel = await createChannel(tempData.guildId, {
+      name: "Discordeno-test",
+    });
+
+    // Assertions
+    assertExists(channel);
+
+    tempData.channelId = channel.id;
+    assertEquals(tempData.channelId, channel.id);
```
itohatweb, line 16: Why do you have this check?

tests/channels/delete_channel.ts
```diff
@@ -0,0 +1,58 @@
...
+      // wait 5 seconds to give it time for CHANNEL_DELETE event
+      await delay(5000);
+      // Make sure it is gone from cache
+      if (cache.channels.has(channel.id))
+        throw new Error(
+          "The channel should have been deleted but it is still in cache."
+        );
+    },
+    ...defaultTestOptions,
+  });
```
itohatweb, line 58: Just formatting
```suggestion
  });

```

tests/channels/create_channel.ts
Skillz4Killz: When i was first testing, it wasn't working properly. IDK why but guildId remained empty and so i added that to alert just in case. We can probably remove these in the future but i think for the time being its safe and basically no impact to performance to have this sanity check.

itohatweb (approved):

PR #829, opened 2021-04-12
types: fix new types issues

Skillz4Killz:


itohatweb (approved):
Few errors but we can fix them in another pr

PR #830, opened 2021-04-12
types: fix some new types for handlers & helpers & structures

itohatweb:


src/handlers/commands/APPLICATION_COMMAND_UPDATE.ts
```diff
@@ -9,9 +13,9 @@ export function handleApplicationCommandUpdate(data: DiscordGatewayPayload) {
...
-  eventHandlers.applicationCommandUpdate?.({
-    ...rest,
-    guildId,
-    applicationId,
-  });
+  eventHandlers.applicationCommandUpdate?.(
+    snakeKeysToCamelCase<ApplicationCommandCreateUpdateDelete>(
+      data.d as DiscordApplicationCommandCreateUpdateDelete,
+    ),
+  );
```
ayntee, line 20:
```ts
const { ...rest } = ...
```
This should be removed entirely

src/handlers/emojis/GUILD_EMOJIS_UPDATE.ts
```diff
@@ -11,7 +11,7 @@ export async function handleGuildEmojisUpdate(data: DiscordGatewayPayload) {
 
   const cachedEmojis = guild.emojis;
   guild.emojis = new Collection(
-    payload.emojis.map((emoji) => [emoji.id ?? emoji.name, emoji]),
+    payload.emojis.map((emoji) => [emoji.id!, emoji]),
```
ayntee, line 14: Explain why this is done

itohatweb: This event is only emitted when a custom emoji is updated which always has an id.

src/handlers/guilds/GUILD_UPDATE.ts
```diff
@@ -30,6 +30,8 @@ export async function handleGuildUpdate(data: DiscordGatewayPayload) {
         if (Array.isArray(cachedValue) && Array.isArray(value)) {
           const different = (cachedValue.length !== value.length) ||
             cachedValue.find((val) => !value.includes(val)) ||
+            // TODO: check if this really works hehe
+            // @ts-ignore this works ts is wrong
```
ayntee, line 34: "this works ts is wrong," please refrain from adding these trivial comments, and add meaningful and descriptive reason as to why `// @ts-ignore` is added

Skillz4Killz (approved):

PR #844, opened 2021-04-13
feat: implement gateway queue

### LINKED ISSUE #836: WS ratelimiter
Currently we don't have a ws request queue we should add that.

Skillz4Killz:
Closes #836

### DIFF
--- a/src/ws/process_queue.ts
+++ b/src/ws/process_queue.ts
@@ -12,6 +12,8 @@ export async function processQueue(id: number) {
 
   while (shard.queue.length) {
     if (shard.ws.readyState !== WebSocket.OPEN) {
+      shard.processingQueue = false;
+      return;
     }
 
     // Send a request that is next in line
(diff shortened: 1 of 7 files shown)


src/ws/process_queue.ts
```diff
@@ -0,0 +1,34 @@
...
+  // If no items or its already processing then exit
+  if (!shard?.queue.length || shard.processingQueue) return;
+
+  shard.processingQueue = true;
+
+  let counter = 0;
+
+  while (shard.queue.length) {
+    if (shard.ws.readyState !== WebSocket.OPEN) {
+    }
```
itohatweb, line 15: This can block the event loop. So do sth. like this.
```suggestion
    if (shard.ws.readyState !== WebSocket.OPEN) {
      setTimeout(ws.processQueue, 1000);
      break;
    }
```

itohatweb (approved):
Love it

PR #858, opened 2021-04-14
feat: implement channel cloning

### LINKED ISSUE #761: Add a function to clone a channel
**Is your feature request related to a problem?**
Currently there is not a way to clone a channel in this library.

**Describe the solution you'd like**
Add a function which can clone  a channel.

**Describe alternatives you've considered**
There is not any alternative.

**Additional context**
You can look at discord.js documentation for an example of what it should probably do.
(The link is a search, because it differs between different types of channels(voice, text....))
https://discord.js.org/#/docs/main/master/search?q=clone

existentiality:
Adding:

- createClone channel function
- .clone(reason) method 
- Modified createChannel to work with createClone (Unsure about this system; might need reworking)
- Added tests for createClone function

closes #761

### DIFF
--- a/src/helpers/channels/clone_channel.ts
+++ b/src/helpers/channels/clone_channel.ts
@@ -10,7 +10,7 @@ export async function cloneChannel(channelId: string, reason?: string) {
     channelId
   );
   //Return undefined if channel is not cached (unsure about error handling)
-  if (!channelToClone) return;
+  if (!channelToClone) throw new Error(Errors.CHANNEL_NOT_FOUND);
 
   //If "name" is undefined as specified by types
   channelToClone.name ??= "new-channel";
--- a/src/helpers/channels/clone_channel.ts
+++ b/src/helpers/channels/clone_channel.ts
@@ -15,14 +15,10 @@ export async function cloneChannel(channelId: string, reason?: string) {
   //If "name" is undefined as specified by types
   channelToClone.name ??= "new-channel";
 
-  //Merge channel data with reason for createChannel options
-  const creationData = {
-    reason,
-    ...channelToClone,
-  };
   //Create the channel (also handles permissions)
   return createChannel(
     channelToClone.guildId!,
-    creationData as CreateGuildChannel
+    channelToClone,
+    reason,
   );
 }
--- a/src/helpers/channels/clone_channel.ts
+++ b/src/helpers/channels/clone_channel.ts
@@ -5,7 +5,7 @@ import { DiscordenoChannel } from "../../structures/channel.ts";
 
 /** Create a copy of a channel */
 export async function cloneChannel(channelId: string, reason?: string) {
-  const channelToClone: DiscordenoChannel | undefined = await cacheHandlers.get(
+  const channelToClone = await cacheHandlers.get(
     "channels",
     channelId
   );
--- a/src/helpers/channels/clone_channel.ts
+++ b/src/helpers/channels/clone_channel.ts
@@ -1,0 +1,28 @@
+import { cacheHandlers } from "../../cache.ts";
+import { createChannel } from "./create_channel.ts";
+import { CreateGuildChannel } from "../../types/guilds/create_guild_channel.ts";
+import { DiscordenoChannel } from "../../structures/channel.ts";
+
+/** Create a copy of a channel */
+export async function cloneChannel(channelId: string, reason?: string) {
+  const channelToClone: DiscordenoChannel | undefined = await cacheHandlers.get(
+    "channels",
+    channelId
+  );
+  //Return undefined if channel is not cached (unsure about error handling)
+  if (!channelToClone) return;
+
+  //If "name" is null or undefined as specified by types
+  channelToClone.name ??= "new-channel";
+
+  //Merge channel data with reason for createChannel options
+  const creationData = {
+    reason,
+    ...channelToClone,
+  };
+  //Create the channel (also handles permissions)
+  return createChannel(
+    channelToClone.guildId!,
+    creationData as CreateGuildChannel
+  );
+}
--- a/src/helpers/channels/clone_channel.ts
+++ b/src/helpers/channels/clone_channel.ts
@@ -2,6 +2,8 @@ import { cacheHandlers } from "../../cache.ts";
 import { createChannel } from "./create_channel.ts";
 import { Errors } from "../../types/misc/errors.ts";
 import { DiscordChannelTypes } from "../../types/channels/channel_types.ts";
+import { calculatePermissions } from "../../util/permissions.ts";
+import { Overwrite } from "../../types/channels/overwrite.ts";
 
 /** Create a copy of a channel */
 export async function cloneChannel(channelId: string, reason?: string) {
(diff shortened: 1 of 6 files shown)


src/helpers/channels/clone_channel.ts
```diff
@@ -0,0 +1,28 @@
+import { cacheHandlers } from "../../cache.ts";
+import { createChannel } from "./create_channel.ts";
+import { CreateGuildChannel } from "../../types/guilds/create_guild_channel.ts";
+import { DiscordenoChannel } from "../../structures/channel.ts";
+
+/** Create a copy of a channel */
+export async function cloneChannel(channelId: string, reason?: string) {
+  const channelToClone: DiscordenoChannel | undefined = await cacheHandlers.get(
```
itohatweb, line 8: Don't use explicit types if cacheHandlers.get already has return types
```suggestion
  const channelToClone = await cacheHandlers.get(
```

src/helpers/channels/clone_channel.ts
```diff
@@ -0,0 +1,28 @@
...
+    "channels",
+    channelId
+  );
+  //Return undefined if channel is not cached (unsure about error handling)
+  if (!channelToClone) return;
```
itohatweb, line 13: You need to import Errors.
```suggestion
  if (!channelToClone) throw new Error(Errors.CHANNEL_NOT_FOUND);
```

src/helpers/channels/clone_channel.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  //If "name" is null or undefined as specified by types
```
itohatweb, line 15:
```suggestion
  //If "name" is undefined as specified by types
```

src/helpers/channels/clone_channel.ts
```diff
@@ -0,0 +1,28 @@
...
+  const creationData = {
+    reason,
+    ...channelToClone,
+  };
+  //Create the channel (also handles permissions)
+  return createChannel(
+    channelToClone.guildId!,
+    creationData as CreateGuildChannel
+  );
+}
```
itohatweb, line 28: This is small you don't need to assign a new variable to it
```suggestion
  //Create the channel (also handles permissions)
  return createChannel(
    channelToClone.guildId!,
    {
      reason,
      ...channelToClone,
    }
  );
}
```

src/helpers/channels/create_channel.ts
```diff
@@ -17,14 +17,22 @@ import { calculateBits } from "../../util/permissions.ts";
 /** Create a channel in your server. Bot needs MANAGE_CHANNEL permissions in the server. */
 export async function createChannel(
   guildId: string,
-  options?: CreateGuildChannel,
+  options?: CreateGuildChannel
 ) {
   const requiredPerms: Set<PermissionStrings> = new Set(["MANAGE_CHANNELS"]);
 
+  //Integration with permissions for channel cloning
```
itohatweb, line 24: I'm not so a fan of this solution maybe discuss more on discord. My preference would be that the cloneChannel function converts the overwrites to a compatible version.

src/structures/channel.ts
```diff
@@ -62,19 +64,23 @@ const baseChannel: Partial<DiscordenoChannel> = {
...
       this.id!,
       overwrites,
-      permissions,
+      permissions
     );
   },
   edit(options, reason) {
     return editChannel(this.id!, options, reason);
   },
+
```
itohatweb, line 73: formatting btw it does not look like you use the default formatter from deno?
```suggestion
```

tests/channels/clone_channel.ts
```diff
@@ -0,0 +1,48 @@
+import { tempData } from "../ws/start_bot.ts";
+import { assertEquals, assertExists } from "../deps.ts";
+import { cache } from "../../src/cache.ts";
+import { cloneChannel } from "../../src/helpers/channels/clone_channel.ts";
+import { delayUntil } from "../util/delay_until.ts";
+
```
itohatweb, line 6: Please add a `ifItFailsBlameWolf` function use the other tests for reference.

tests/channels/clone_channel.ts
```diff
@@ -0,0 +1,48 @@
...
+Deno.test({
+  name: "[channel] clone a channel",
+  async fn() {
+    const cloned = await cloneChannel(tempData.channelId, "testing");
```
itohatweb, line 10: I would prefer you create a channel with bitrate & topic etc. and clone this channel

itohatweb (changes requested):
Found a few things I would want to be changed. I don't know if you had a reason for any change if so please let me know.

src/helpers/channels/clone_channel.ts
```diff
@@ -0,0 +1,24 @@
...
+export async function cloneChannel(channelId: string, reason?: string) {
+  const channelToClone = await cacheHandlers.get(
+    "channels",
+    channelId
+  );
+  //Return undefined if channel is not cached (unsure about error handling)
+  if (!channelToClone) throw new Error(Errors.CHANNEL_NOT_FOUND);
+
+  //If "name" is undefined as specified by types
+  channelToClone.name ??= "new-channel";
```
Skillz4Killz, line 16: 1. Add a check to make sure dm channels don't work
2. Remove this name thing as it will always have a name

src/helpers/channels/create_channel.ts
Skillz4Killz: Agreed!

tests/channels/clone_channel.ts
Skillz4Killz: i think its fine if it just uses 1 test since u can't reuse the code here

tests/channels/clone_channel.ts
Skillz4Killz: You can't create a channel with a bitrate and topic. Topic would require text channel bitrate would be voice.

I would recommend every type of channel being cloned with each property

src/helpers/channels/create_channel.ts
existentiality: This was my original idea, I just wasn't sure how to do it. I don't really like how I did this either. Definitely open to change here.

itohatweb: You can loop over the overwrites and just use the `calculateBits()` function to change the allow and deny props.

existentiality: Changed it to use `calculatePermissions()` in cloneChannel and removed the integration in createChannel.

tests/channels/clone_channel.ts
existentiality: Reworked clone tests (see tests/channels/clone_channel.ts)

tests/channels/clone_channel.ts
```diff
@@ -0,0 +1,159 @@
...
+    await ifItFailsBlameWolf(
+      { name: "Discordeno-clone-test", nsfw: true },
+      false,
+    );
+  },
+  ...defaultTestOptions,
+});
+
+Deno.test({
+  name: "[channel] create a new text channel with permission overwrites",
```
Skillz4Killz, line 141:
```suggestion
  name: "[channel] clone a new text channel with permission overwrites",
```

existentiality: Missed that when moving tests over, thanks!

Skillz4Killz:
@existentiality Thank you!

PR #876, opened 2021-04-30
Some things

itohatweb:


src/helpers/messages/remove_reaction.ts
```diff
@@ -1,12 +1,18 @@
...
 import { endpoints } from "../../util/constants.ts";
+import { requireBotChannelPermissions } from "../../util/permissions.ts";
 
-/** Removes a reaction from the bot on this message. Reaction takes the form of **name:id** for custom guild emoji, or Unicode characters. */
+/** Removes a reaction from the given user on this message, defaults to bot. Reaction takes the form of **name:id** for custom guild emoji, or Unicode characters. */
 export async function removeReaction(
   channelId: string,
   messageId: string,
   reaction: string,
+  userId?: string,
```
Skillz4Killz, line 10: Make this into an object so in future we can add more optional things and not have a breaking change

Skillz4Killz: { userId?: string }

Skillz4Killz (approved):

PR #878, opened 2021-04-30
add cache members on start option

itohatweb:


Skillz4Killz:
use case?

itohatweb:
I thought you mentioned that we should add an option to cache members on start in case users don't want to. If thats not the case, I will revert it.

Skillz4Killz:
Ahh no this isn't exactly what i meant. I meant we should consider whether or not we should force all members to be cached from intial queue thing before ready event is run

Skillz4Killz:
Closing this as it needs to be rewroked at a later time as discussed in discord

PR #887, opened 2021-05-01
add(types): nsfw field

keenkairos:
Reference: https://github.com/discord/discord-api-docs/commit/5e63e309bbc7e93b683b47bc8401d2d079580251


src/types/guilds/guild.ts
```diff
@@ -107,4 +107,6 @@ export interface Guild {
   approximatePresenceCount?: number;
   /** The welcome screen of a Community guild, shown to new members, returned when in the invite object */
   welcomeScreen?: WelcomeScreen;
+  /** `true` if this guild is designated as NSFW. */
```
itohatweb, line 110:
```suggestion
  /** `true` if this guild is designated as NSFW */
```

PR #920, opened 2021-05-10
change(util): util-specific deps.ts

ghost:
The working-directory-level ﻿dependencies file (/deps.ts) should contain dependencies that are shared between several modules. Identical to the existing modules, the util module similarly should also have an accompanying dependencies file that accomodates the used dependencies inside that specific module. This pull request does just that.


ayntee:
run-tests

PR #925, opened 2021-05-11
fix: return undefined on 204 status

itohatweb:


src/rest/run_method.ts
```diff
@@ -62,7 +62,11 @@ export async function runMethod<T = any>(
         method,
         reject,
         respond: (data: { status: number; body?: string }) =>
-          resolve(camelize<T>(JSON.parse(data.body || "{}"))),
+          resolve(
+            data.status !== 204
+              ? camelize<T>(JSON.parse(data.body ?? "{}"))
+              : <T> (<unknown> undefined),
```
rigormorrtiss, line 68:
```suggestion
              : (undefined as unknown) as T,
```
Both of these are equivalent to each other, however, using `as` for type assertions is the recommended syntax.

src/rest/run_method.ts
```diff
@@ -62,7 +62,11 @@ export async function runMethod<T = any>(
         method,
         reject,
         respond: (data: { status: number; body?: string }) =>
-          resolve(camelize<T>(JSON.parse(data.body || "{}"))),
+          resolve(
+            data.status !== 204
```
rigormorrtiss, line 66: ideally, you should use the `DiscordHTTPResponseCodes` enum here, specifically, `DiscordHTTPResponseCodes.NoContent` (204).

PR #931, opened 2021-05-12
fix(types): HTTPResponseCodes alias typo

rigormorrtiss:


src/types/codes/http_response_codes.ts
```diff
@@ -13,5 +13,4 @@ export enum DiscordHTTPResponseCodes {
   GatewayUnavailable = 502,
 }
 
-export type HTTPPResponseCodes = DiscordHTTPResponseCodes;
-export const HTTPPResponseCodes = DiscordHTTPResponseCodes;
+export { DiscordHTTPResponseCodes as HTTPResponseCodes };
```
itohatweb, line 16:
```suggestion
export type HTTPResponseCodes = DiscordHTTPResponseCodes;
export const HTTPResponseCodes = DiscordHTTPResponseCodes;
```

rigormorrtiss: Any reason as to why this is necessary?

itohatweb: Skillz decided to use this to export enum aliases

PR #936, opened 2021-05-13
change: check shard unavailability using last received guild create event 

itohatweb:


src/bot.ts
```diff
@@ -94,7 +94,7 @@ export async function startBigBrainBot(options: BigBrainBotConfig) {
 
     // Initial API connection to get info about bots connection
     ws.botGatewayData = await getGatewayBot();
-    ws.maxShards = ws.maxShards ||
+    ws.maxShards = options.lastShardId || ws.maxShards ||
```
Skillz4Killz, line 97: these two are very different things should never conflict like this

Skillz4Killz: Bad change please undo

src/handlers/misc/READY.ts
```diff
@@ -12,83 +12,75 @@ export function handleReady(
...
-          `3. Running setTimeout in READY file.`,
-        );
-        loaded(shardId);
-      }, 2000);
-    } else {
-      cache.isReady = true;
-      eventHandlers.ready?.();
-    }
+  // If it is not the last shard we can't go full ready
+  if (shard.id !== ws.maxShards - 1) return;
```
Skillz4Killz: use lastShardId here imo

PR #946, opened 2021-05-16
Implement Class Based 

Skillz4Killz:
This PR begins a Class based implementation of the library separate from the functional api side. This will allow users who wish to use classes the ability to do so.

Note: We should aim to reuse as much code as possible from our functional side. For example you can see how i added the `sendMessage` method to the Client class. Instead of rewriting it, that function is simply used.

Please leave your thoughts/feedback below


Cormanz:
This seems pretty nice and useful, might try out Discordeno once it's ready!

Skillz4Killz:
I am marking this pr as not going to merge unless someone can find some ways to fix the remaining errors without effecting the end user experience in the functional side.

Issue #955, opened 2021-05-20
editBotProfile function can crash the bot with invalid image url


Yaikava:
**Describe the bug** The bot will crash if invalid image url is passed into `editBotProfile()`.
This error is not solvable by using `.catch()`.

**To Reproduce** Steps to reproduce the behavior:
1. code (using template)
```ts
import { editBotProfile } from "../../../deps.ts";
import { Embed } from "../../utils/Embed.ts";
import { createCommand } from "../../utils/helpers.ts";

createCommand({
    name: "setbotavatar",
    arguments: [
        {
            name: "link_to_picture",
            type: "string",
            required: true,
        }
    ],
    execute: async (message, args) => {
        await editBotProfile(undefined, args.link_to_picture).catch((err) => {
            console.error(`Failed to edit bot profile: ${err}`);
            return message.reply("Error...");
        });
        message.reply("Changed avatar")
    }
});
```
2. pass an invalid url here

**Expected behavior**
Error will be "solved" by `.catch()`. (Instead whole bot crashes.)

**Screenshots** If applicable, add screenshots to help explain your problem.

**Version details (please complete the following information):**

- Discordeno version: v11 latest (Logs show a locked commit, but its the same.)
- Deno version: 1.10.2

**Additional context** Add any other context about the problem here.
Error
```ts
error: Uncaught (in promise) TypeError: Cannot set property verified of #<Object> which has only a getter
    if (member[key] !== value) return member[key] = value;
                                                 ^
    at https://raw.githubusercontent.com/discordeno/discordeno/0ad38809c450dfb6bf9d36310e3737b1aa8a9bcd/src/handlers/misc/USER_UPDATE.ts:22:50
    at Array.forEach (<anonymous>)
    at Object.handleUserUpdate [as USER_UPDATE] (https://raw.githubusercontent.com/discordeno/discordeno/0ad38809c450dfb6bf9d36310e3737b1aa8a9bcd/src/handlers/misc/USER_UPDATE.ts:16:28)
```

PR #969, opened 2021-05-21
change: createDiscordenoMember make guild member optional

itohatweb:
Just some way of thinking. Since we don't have anything which depends on at least having one guildMember we can just make it optional.

Why:
- usage of message.member in dms
- members in discordeno are users just with getters so making guild member optional makes sense imo
- getters for users which is pretty neat

also closes: #955


Skillz4Killz:
Not a fan of this PR

PR #985, opened 2021-05-23
fix: role.permissions should be bigint

itohatweb:


src/structures/role.ts
```diff
@@ -129,6 +129,7 @@ export async function createDiscordenoRole(
 
   const role: DiscordenoRole = Object.create(baseRole, {
     ...props,
+    permissions: createNewProp(snowflakeToBigint(rest.permissions)),
```
Skillz4Killz, line 132:
```suggestion
    permissions: createNewProp(BigInt(rest.permissions)),
```

Skillz4Killz: this is perms not a snowflake although result is same, using it here is confusing

Skillz4Killz (approved):

PR #993, opened 2021-05-27
change: interactionTypes rename Button to MessageComponent

rigormorrtiss:


itohatweb:
If you do breaking changes check if you need to update code at other places accordingly

PR #1002, opened 2021-05-28
add: select-dropdowns

Skillz4Killz:


src/types/messages/components/select_data.ts
```diff
@@ -0,0 +1,8 @@
+export interface SelectMenuData {
+  /** The type of component */
+  componentType: 3;
```
itohatweb, line 3: use the enum here

src/types/messages/components/select_menu.ts
```diff
@@ -0,0 +1,15 @@
+import { SelectOption } from "./select_option.ts";
+
+export interface SelectMenuComponent {
+  type: 3;
```
itohatweb, line 4: use the enum here

itohatweb (changes requested):

PR #1029, opened 2021-06-10
Check if token is provided

lumap:
better error handling - a "no token" error is better than 

Error: [401] The Authorization header was missing or invalid.
    at Object.processQueue (https://deno.land/x/discordeno@11.0.3/src/rest/process_queue.ts:101:40)
    at async Object.processRequest (https://deno.land/x/discordeno@11.0.3/src/rest/process_request.ts:31:5)


src/bot.ts
```diff
@@ -17,6 +17,7 @@ export let eventHandlers: EventHandlers = {};
 export let proxyWSURL = `wss://gateway.discord.gg`;
 
 export async function startBot(config: BotConfig) {
+  if (!config.token) throw Error("No token provided!")
```
itohatweb, line 20: Good Idea but please add the error to our Errors enum and don't forget the new keyword so like
`throw new Error(Errors.NO_TOKEN_PROVIDED)`

itohatweb (changes requested):

itohatweb:
oh wait shouldn't ts prevent this?

lumap:
> oh wait shouldn't ts prevent this?

What do you mean?

itohatweb:
Typescript gives you a visual error if you do not provide a token
image

lumap:
if you provide an empty string typescript says it's okay tho

abhiss:
> I don't agree with this PR. If you give a bad token, that's really on the developer themself. TS will already check and warn the user. Even if they don't it will error in a second and you can see the token is invalid

The current error doesn't tell you the token is invalid, it it reads `Error: [401] The Authorization header was missing or invalid`, followed by a stack trace. A lot of new developers make discord bots and messing up the token is a reasonable mistake for a beginner to make. I think it would help new users if the error mentioned the API token.

itohatweb:
`The Authorization header was missing or invalid` is just a nicer error message from our side which can easily be changed. Maybe we find a better name and then change it in our code would this be ok for you?

abhiss:
@itohatweb I think that would be great :)

PR #1030, opened 2021-06-10
v11.1.0

itohatweb:
### Added

- CreateMessage/EditMessage#embeds property
- JSON Error Code Maximum Number Of Animated Emojis Reached
- JSON Error Code Maximum Number Of Server Members Reached

### Fixed

- cache.emojis now really uses a bigint id as the key

### Deprecated

- CreateMessage/EditMessage#embed property (Will be removed in v12 use embeds instead)

[11.1.0]: https://github.com/discordeno/discordeno/compare/11.0.3...11.1.0

### DIFF
--- a/src/helpers/messages/edit_message.ts
+++ b/src/helpers/messages/edit_message.ts
@@ -29,6 +29,12 @@ export async function editMessage(channelId: bigint, messageId: bigint, content:
     validateComponents(content.components);
   }
 
+  // TODO: v12 remove
+  if (content.embed) content.embeds = [content.embed, ...(content.embeds || [])];
+  if (content.embeds && content.embeds.length > 10) {
+    content.embeds.splice(10);
+  }
+
   if (content.content && content.content.length > 2000) {
     throw new Error(Errors.MESSAGE_MAX_LENGTH);
   }
--- a/src/helpers/messages/edit_message.ts
+++ b/src/helpers/messages/edit_message.ts
@@ -30,10 +30,11 @@ export async function editMessage(channelId: bigint, messageId: bigint, content:
   }
 
   // TODO: v12 remove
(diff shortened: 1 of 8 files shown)


src/helpers/messages/edit_message.ts
```diff
@@ -29,6 +29,12 @@ export async function editMessage(channelId: bigint, messageId: bigint, content:
     validateComponents(content.components);
   }
 
+  // TODO: v12 remove
+  if (content.embed) content.embeds = [content.embed, ...(content.embeds || [])];
```
Skillz4Killz, line 33: I think content.embed should be updated to be undefined after this

src/helpers/messages/send_message.ts
```diff
@@ -34,7 +34,13 @@ export async function sendMessage(channelId: bigint, content: string | CreateMes
     const requiredPerms: Set<PermissionStrings> = new Set(["SEND_MESSAGES", "VIEW_CHANNEL"]);
 
     if (content.tts) requiredPerms.add("SEND_TTS_MESSAGES");
-    if (content.embed) requiredPerms.add("EMBED_LINKS");
+    // TODO: v12 remove
+    if (content.embed) content.embeds = [content.embed, ...(content.embeds || [])];
+    if (content.embeds?.length) requiredPerms.add("EMBED_LINKS");
+    if (content.embeds && content.embeds.length > 10) {
```
Skillz4Killz, line 40:
```
    content.embeds?.splice(10);
```
I think this covers it all

src/helpers/messages/edit_message.ts
itohatweb: Discord does not care about `embed` if we send an `embeds` too. It will always check for `embeds` first

PR #1041, opened 2021-06-13
feat(types): add stageInstance field to Invite

rigormorrtiss:
Reference: https://github.com/discord/discord-api-docs/pull/3120

### DIFF
--- a/src/types/invites/invite_stage_instance.ts
+++ b/src/types/invites/invite_stage_instance.ts
@@ -2,7 +2,7 @@ import { GuildMember } from "../members/guild_member.ts";
 
 export interface InviteStageInstance {
   /** The members speaking in the Stage */
-  members: Partial<GuildMember>;
+  members: Partial<GuildMember>[];
   /** The number of users in the Stage */
   participantCount: number;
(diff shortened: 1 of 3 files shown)


src/types/invites/invite_stage_instance.ts
```diff
@@ -0,0 +1,12 @@
+import { GuildMember } from "../members/guild_member.ts";
+
+export interface InviteStageInstance {
+  /** The members speaking in the Stage */
+  members: Partial<GuildMember>;
```
Skillz4Killz, line 5: This doesnt look right can u double check this plz

Skillz4Killz: 1. members but not an array?
2. what is it giving? Please double check because `{}` this is valid as per ur typings.
3. Any snowflakes that need to be made into bigints? because invites get cached

rigormorrtiss, line 5:
```suggestion
  members: Partial<GuildMember>[];
```

ayntee (approved):
LGTM, thank you!

PR #1044, opened 2021-06-15
fix(types): mark id, userId as optional in ThreadMember

rigormorrtiss:
The id and user_id fields are omitted on the member sent within each thread in the `GUILD_CREATE` event.

## Links

- Reference: https://github.com/discord/discord-api-docs/pull/3097﻿
- Files Diff: https://github.com/discord/discord-api-docs/pull/3097/files#diff-160acb4654c611c67c4ee46b2ce5dda2be82212af8d1d00cd0256ef6edda1e95R543

### DIFF
--- a/src/types/channels/threads/thread_member.ts
+++ b/src/types/channels/threads/thread_member.ts
@@ -1,8 +1,8 @@
 export interface ThreadMember {
   /** The id of the thread */
-  id: string;
+  id?: string;
   /** The id of the user */
-  userId: string;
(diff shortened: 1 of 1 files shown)


ayntee:
We can further improve our types by separating those each case into its own interface and extending the base interface. This can also be done with various types of channel, but for now, it should be fine.
```ts
export interface GuildCreateThreadMember { ... }
```

ayntee (approved):
LGTM

PR #1054, opened 2021-06-18
feat: add rateLimit event handler

rigormorrtiss:


itohatweb:
This is bad it requires users of rest proxy to import whole discordeno

itohatweb:
Also there is already an event handler for this

PR #1063, opened 2021-06-21
change: make thread#guildId a getter

itohatweb:


Skillz4Killz:
The interface also

ayntee:
@itohatweb any update on this?

ayntee:
> The interface also

The interface already has the guildId property (treated as getter).

PR #1068, opened 2021-06-29
feat: requiredStructureProperties

Skillz4Killz:
This pr adds a new feature to the library for advanced bot developers. 

The purpose of this feature is for optimizing memory in an easier way than having to customize structures and overriding them. For example, if you simply want to remove some properties you don't use, this would be the ideal way.

```ts
// This example is just to easily show you but you should ideally use an array and loop over the array
cache.requiredStructureProperties.channels.add("topic");
cache.requiredStructureProperties.channels.add("id");
```

This would now make it so that your channels will are ONLY storing the `topic` and `id` properties and nothing else.

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -8,6 +8,12 @@ import type { PresenceUpdate } from "./types/activity/presence_update.ts";
 import type { Emoji } from "./types/emojis/emoji.ts";
 import { DiscordenoThread } from "./util/transformers/channel_to_thread.ts";
 import { Collection } from "./util/collection.ts";
+import { Channel } from "./types/channels/channel.ts";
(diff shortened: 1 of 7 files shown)


ayntee:
Wouldn't this break some functions (permission checks) if required properties are missing from their respective structures similar to how custom structures can?

ayntee:
Also, I think we should omit the fields from the returned structure for better UX.

Skillz4Killz:
@ayntee @itohatweb ready for review

ayntee (approved):
LGTM

itohatweb (approved):

itohatweb:
LGTM but maybe add a `requiredPropsSize` const for the other structures too

PR #1092, opened 2021-08-09
fix(helpers): pass null if options.channelId is falsy

rigormorrtiss:
image

### DIFF
--- a/src/helpers/members/edit_member.ts
+++ b/src/helpers/members/edit_member.ts
@@ -54,7 +54,7 @@ export async function editMember(guildId: bigint, memberId: bigint, options: Mod
     endpoints.GUILD_MEMBER(guildId, memberId),
     snakelize({
       ...options,
-      channelId: options.channelId ? bigintToSnowflake(options.channelId) : undefined,
+      channelId: options.channelId ? bigintToSnowflake(options.channelId) : null,
     }) as ModifyGuildMember
   );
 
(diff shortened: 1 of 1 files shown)


itohatweb (changes requested):
This is a bad solution. With that one you require the user to always pass a channelId if the bot shall not move the user.

Skillz4Killz:
We have `disconnectMember()` no?

golfIII:
> 
> 
> We have `disconnectMember()` no?

Hi, I'm the Discord member in the image above.

Same issue occurs when using `disconnectMember()`. Perms are valid, channel is valid, and the member being kicked is valid and is successfully returned from both `channel.disconnect()` and `disconnectMember()`.

I believe @ayntee was able to reproduce the issue on his end as well.

ayntee:
@Skillz4Killz disconnectMember() calls editMember() under the hood

ayntee:
> This is a bad solution. With that one you require the user to always pass a channelId if the bot shall not move the user.

@itohatweb looks like the PR has been updated

src/helpers/members/edit_member.ts
```diff
@@ -54,7 +54,7 @@ export async function editMember(guildId: bigint, memberId: bigint, options: Mod
     endpoints.GUILD_MEMBER(guildId, memberId),
     snakelize({
       ...options,
-      channelId: options.channelId ? bigintToSnowflake(options.channelId) : undefined,
+      channelId: typeof options.channelId === "bigint" ? bigintToSnowflake(options.channelId) : options.channelId,
```
Skillz4Killz, line 57:
```suggestion
      channelId: options.channelId ? bigintToSnowflake(options.channelId) : options.channelId,
```

Skillz4Killz (approved):

PR #1100, opened 2021-08-18
fixing REQUEST_TO_SPEAK to not toggle another perm

RanAwaySuccessfully:
I guess they do say that "This permission is under active development and may be changed or removed." but still. I was wondering why `CREATE_INSTANT_INVITE` was just magically appearing in my permission list.

I have no idea why my server has `REQUEST_TO_SPEAK` in some roles, it just appeared one day.


itohatweb:
fixed already .-.

RanAwaySuccessfully:
welp

PR #1399, opened 2021-11-01
added type undefined to interface BotConfig.token

IshanKBG:
when we use an environment variable using `Deno.env.get("VAR") ` it shows a type error which says that `type undefined is not assignable to type 'string'`

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -63,7 +63,7 @@ export function setApplicationId(id: string) {
 }
 
 export interface BotConfig {
-  token: string;
+  token: string | undefined;
   compress?: boolean;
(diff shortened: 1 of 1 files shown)


remagnedlogeht:
but token is necessary to start up the bot
and for your issue you can use ``Deno.env.get("VAR")!``

itohatweb:
This change is bad. You should create a var in which you store the token and then check if it is defined.

lts372005:
it doesnt hurt to check
```ts
const token = Deno.env.get("TOKEN")
if(!token) throw new Error("NO TOKEN")
startBot({
  token,
  intents: [],
  eventHandlers: {
    ready(){
      console.log("READY")
    }
    // ...
  }
})
```

IshanKBG:
ok thanks

PR #1810, opened 2021-11-21
fix(rest): only shared rate limit scopes can be ignored

itohatweb:


itohatweb:
Actually needs more thinking

PR #1831, opened 2021-12-05
FIxing typings of sendInteractionResponse() because is necessary

Le-Val:
I only tested the typings btw


itohatweb:
can you remove the format changes pls? easier to review then

PR #1841, opened 2021-12-09
for yui

itohatweb:
Fixing type bugs which occured since modals. 

We really need a better way to do this stuff tbh. had to change 7 files...


itohatweb:
PS: tests failing cuz RC types broken. Will pass once a new tag has been pushed.

PR #1867, opened 2021-12-23
Move discordeno user and transform user from /src/transformers/member.ts to /src/transformers/user.ts

lts20050703:
move `DiscordenoUser` interface and `transformUser` function from `src/transformers/member.ts` to `/src/transformers/user.ts`

add `export * from "./user.ts"` to /src/transformers/mod.ts


lts20050703:
[closed] it's a mess, I need to split this pr into 3

PR #1871, opened 2021-12-28
fix(tests): delete invite 

lts20050703:
fix delete invite test

`deleteInvite` function return `Promise<void>`, therefore the correct way to test is to check if `getInvite` function THROWS/REJECTS

### DIFF
--- a/tests/invite/deleteInvite.ts
+++ b/tests/invite/deleteInvite.ts
@@ -1,4 +1,4 @@
-import { assertExists } from "../deps.ts";
+import { assertExists, assertRejects } from "../deps.ts";
 import { bot, channel } from "../mod.ts";
 
 Deno.test({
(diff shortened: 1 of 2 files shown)


itohatweb:
run-tests

tests/invite/deleteInvite.ts
```diff
@@ -14,9 +14,9 @@ Deno.test({
     // Assertions
     assertExists(invite);
 
-    const deletedInvite = await bot.helpers.deleteInvite(invite.code);
+    await bot.helpers.deleteInvite(invite.code);
 
     // Assertions
-    assertExists(deletedInvite);
```
Skillz4Killz, line 20: Very bad idea, this is throwing a request at the api every 100ms until it magically appears

Skillz4Killz (changes requested):

tests/invite/deleteInvite.ts
lts20050703: deleteInvite() returns `Promise<void>` so I use assertRejects getInvite() (getInvite throws an error if not found)
what do you mean by "this is throwing a request at the api every 100ms until it magically appears"

Skillz4Killz: I misread, sorry was tired. Looks good

Skillz4Killz (approved):

Issue #1904, opened 2022-01-19
[api-docs] Clarification on channel icons (#4315)

* specify group dms

* Update docs/resources/Channel.md

Co-authored-by: Zeyla Hellyer <zeyla@hellyer.dev>


itohatweb:
A new commit was made into the api-docs repo.
https://github.com/discord/discord-api-docs/commit/13e45f4d51ab3e35c6ce4698718874f1fd277b20

This is a bot created issue.

PR #1925, opened 2022-01-26
Clarification on channel icons

lts20050703:
Closes: #1904


Skillz4Killz (changes requested):
Bots cant be in group dms id prefer we remove this instead

Skillz4Killz (approved):
Does this need removing in a helper/transformer?

lts20050703:
> Does this need removing in a helper/transformer?

Helpers like getChannel use channel transformer so yeah we only need to fix the channel transformer

Skillz4Killz (approved):

Issue #1905, opened 2022-01-19
[api-docs] document the required permission for retrieving the welcome screen when disabled (#4344)


itohatweb:
A new commit was made into the api-docs repo.
https://github.com/discord/discord-api-docs/commit/00a0ae27c22b257ca1a154ec25e17bd3fffb80ec

This is a bot created issue.

PR #1926, opened 2022-01-26
document the required permission for retrieving the welcome screen when disabled

lts20050703:
Closes: #1905


Skillz4Killz (approved):
This should also be done in perms plugin

Issue #1906, opened 2022-01-19
[api-docs] Document X-Audit-Log-Reason on webhook endpoints. (#4276)

* Document X-Audit-Log-Reason on webhook endpoints.


itohatweb:
A new commit was made into the api-docs repo.
https://github.com/discord/discord-api-docs/commit/c7501f10a0acaec4f77b5ddc6dd9139c5509ef16

This is a bot created issue.

PR #1927, opened 2022-01-26
Document X-Audit-Log-Reason on webhook endpoints

lts20050703:
Closes: #1906

### DIFF
--- a/src/types/webhooks/createWebhook.ts
+++ b/src/types/webhooks/createWebhook.ts
@@ -3,5 +3,6 @@ export interface CreateWebhook {
   name: string;
   /** Image for the default webhook avatar */
   avatar?: string | null;
+  /** The reason you are creating this webhook */
   reason?: string;
 }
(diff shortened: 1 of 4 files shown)


src/types/webhooks/createWebhook.ts
```diff
@@ -3,4 +3,5 @@ export interface CreateWebhook {
   name: string;
   /** Image for the default webhook avatar */
   avatar?: string | null;
+  reason?: string;
```
Skillz4Killz, line 6: add a /** */ also what about the function itself?

src/types/webhooks/modifyWebhook.ts
```diff
@@ -6,4 +6,5 @@ export interface ModifyWebhook {
   avatar?: string | null;
   /** The new channel id this webhook should be moved to */
   channelId?: string;
+  reason?: string;
```
Skillz4Killz, line 9: add a /** */ also what about the function itself?

Skillz4Killz (changes requested):

src/types/webhooks/createWebhook.ts
lts20050703: the function should work?
image

src/types/webhooks/modifyWebhook.ts
lts20050703: The function should work?
image

Skillz4Killz (approved):

Issue #1917, opened 2022-01-25
[api-docs] Document partial member data in message interaction object (#4354)


itohatweb:
A new commit was made into the api-docs repo.
https://github.com/discord/discord-api-docs/commit/56cea9462fae6a9b6399e5224b090dc2b68a77e0

This is a bot created issue.

PR #1958, opened 2022-01-28
discord-api-docs: Document partial member data in message interaction object

lts20050703:
discord/discord-api-docs@56cea94
Closes: #1917


Skillz4Killz (commented):
Needs to reflect it in the transformer

lts20050703:
In case you can't tell I am extremely confused what's the different between `message.member` and `message.interaction.member` so feel free to correct me if I got something wrong

Skillz4Killz (approved):

PR #1968, opened 2022-02-01
chore(.github): remove @ayntee from codeowners

itohatweb:
Ayntee is currently not here so we can remove him from codeowners for now and add him back when he is back.


Skillz4Killz (approved):
Message us when your back ayntee!

PR #1979, opened 2022-02-02
fix(gatway): resharder bug

Skillz4Killz:
This PR is intended to finalize the Resharder system which was semi complete until now and also fix a bug that another dev found.

1. We now enable resharding by default once a day checks for all bots made with createBot() Although we have to think and decide if this should be moved to createGatewayManager or startBot()
2. The previous style of resharding was: to start another shard behind the scenes and as soon is it gets a READY we closed the original shard. However, a bug was caught in this by a genius mind who saw that if a guild were to switch shards upon resharding since the formula would change this can cause issues. To solve this we created a new system to reshard.

a. Same old checks as before to make sure resharding is necessary and doable.
b. Create a new GatewayManager, using the same values/options from the old one. However, set the handleDiscordPayload to do nothing so it ignores events to avoid duplicate event handling.
c. Edit the new gateway settings with new resharding values.
d. Begin spawning the shards behind the scenes.
e. Once all shards have been completed, swap the discord handlers so new gateway is handling events and old is now ignoring on all shards. Then begin removing all shards on old gateway that had NO queues. Set a timer to remove leftover shards after 5 minutes to give em time to complete any leftover requests.


src/ws/resharder.ts
```diff
@@ -1,31 +1,84 @@
...
 import { GetGatewayBot } from "../types/gateway/getGatewayBot.ts";
 
 /** The handler to automatically reshard when necessary. */
-export async function resharder(gateway: GatewayManager) {
+export async function resharder(oldGateway: GatewayManager) {
+  oldGateway.debug("[Resharding] Checking if resharding is needed.");
   // TODO: is it possible to route this to REST?
-  const results = (await fetch(`https://discord.com/api/gateway/bot`, {
-    headers: { Authorization: gateway.token },
+  const results = (await fetch(`https://discord.com/api/oldGateway/bot`, {
```
Skillz4Killz, line 8:
```suggestion
  const results = (await fetch(`https://discord.com/api/gateway/bot`, {
```

src/ws/resharder.ts
```diff
@@ -1,31 +1,84 @@
...
   }
 
   gateway.spawnShards(gateway, gateway.firstShardId);
+
+  return new Promise((resolve) => {
+    // TIMER TO KEEP CHECKING WHEN ALL SHARDS HAVE RESHARDED
+    const timer = setInterval(() => {
+      let pending = false;
+      for (let i = oldGateway.firstShardId; i < oldGateway.lastShardId; i++) {
+        const shard = oldGateway.shards.get(i);
```
Skillz4Killz, line 52:
```suggestion
        const shard = gateway.shards.get(i);
```

src/ws/resharder.ts
```diff
@@ -1,31 +1,84 @@
...
+
+      // ENABLE EVENTS ON NEW SHARDS AND IGNORE EVENTS ON OLD
+      gateway.handleDiscordPayload = oldGateway.handleDiscordPayload;
+      oldGateway.handleDiscordPayload = function () {};
+
+      // SHUT DOWN ALL SHARDS IF NOTHING IN QUEUE
+      oldGateway.shards.forEach((shard) => {
+        // CLOSE THIS SHARD IT HAS NO QUEUE
+        if (!shard.processingQueue && !shard.queue.length) {
+          oldGateway.closeWS(shard.ws, 3066, "Shard has been resharded. Closing shard since it has no queue.");
```
Skillz4Killz, line 70:
```suggestion
          return oldGateway.closeWS(shard.ws, 3066, "Shard has been resharded. Closing shard since it has no queue.");
```

src/ws/resharder.ts
```diff
@@ -1,31 +1,84 @@
...
+        }
+
+        // IF QUEUE EXISTS GIVE IT 5 MINUTES TO COMPLETE
+        setTimeout(() => {
+          oldGateway.closeWS(shard.ws, 3066, "Shard has been resharded. Delayed closing shard since it had a queue.");
+        }, 300000);
+      });
+      // STOP TIMER
+      clearInterval(timer);
+      oldGateway.debug("[Resharding] Complete.");
```
Skillz4Killz, line 80:
```suggestion
      gateway.debug("[Resharding] Complete.");
```

src/ws/resharder.ts
```diff
@@ -1,31 +1,102 @@
...
+      // STILL PENDING ON SOME SHARDS TO BE CREATED
+      if (pending) return;
+
+      // ENABLE EVENTS ON NEW SHARDS AND IGNORE EVENTS ON OLD
+      gateway.handleDiscordPayload = oldGateway.handleDiscordPayload;
+      oldGateway.handleDiscordPayload = function () {};
+
+      // STOP TIMER
+      clearInterval(timer);
+      await gateway.resharderCloseOldShards(oldGateway);
```
itohatweb, line 47: The handler function gets removed from the old gateway. 
The shards are closed later whilst the closer waits for the queues to finish. The problem now is that if it was for example a request guild members request, then this will never finish since the handler function does not receive the response from discord anymore.
Has this been considered?

src/ws/resharder.ts
```diff
@@ -1,31 +1,102 @@
...
+      gateway.debug("[Resharding] Complete.");
+      resolve(gateway);
+    }, 30000);
+  }) as Promise<GatewayManager>;
+}
+
+/** Handler that by default will check all new shards are online in the new gateway. The handler can be overriden if you have multiple servers to communicate through redis pubsub or whatever you prefer. */
+export async function resharderIsPending(gateway: GatewayManager, oldGateway: GatewayManager) {
+  let pending = false;
+  for (let i = gateway.firstShardId; i < gateway.maxShards; i++) {
```
itohatweb, line 57: We should come up with a better naming for `lastShardId` and `maxShards`. Those happen to get pretty confusing.
Or improve the jsdoc of those props

This is just a note and should be done in a separate pr.

src/ws/resharder.ts
```diff
@@ -1,31 +1,102 @@
...
+    const shard = gateway.shards.get(i);
+    if (!shard?.ready) {
+      pending = true;
+      break;
+    }
+  }
+
+  return pending;
```
itohatweb, line 65:
```suggestion
  for (let i = gateway.firstShardId; i < gateway.maxShards; i++) {
    const shard = gateway.shards.get(i);
    if (!shard?.ready) {
      return true;
    }
  }

  return false;
```

src/ws/resharder.ts
```diff
@@ -1,31 +1,102 @@
...
+  });
+}
+
+/** Handler that by default will check to see if resharding should occur. Can be overriden if you have multiple servers and you want to communicate through redis pubsub or whatever you prefer. */
+export async function startReshardingChecks(gateway: GatewayManager) {
+  gateway.debug("[Resharding] Checking if resharding is needed.");
+
+  // TODO: is it possible to route this to REST?
+  const results = (await fetch(`https://discord.com/api/gateway/bot`, {
+    headers: { Authorization: gateway.token },
```
itohatweb, line 90: `token` should be renamed to `authorization` for better differenciation.

This is for a separate PR.

itohatweb (changes requested):

Skillz4Killz:
Missing: Just thought of something we need to update all guild objects with new shard id on reshard

src/ws/gateway_manager.ts
```diff
@@ -166,8 +180,21 @@ export interface GatewayManager {
...
-  resharder: typeof resharder;
+  /** The methods related to resharding. */
+  resharding: {
+    /** Handles resharding the bot when necessary. */
+    resharder: typeof resharder;
+    /** Handles checking if all new shards are online in the new gateway. */
+    isPending: typeof resharderIsPending;
+    /** Handles closeing all shards in the old gateway. */
+    closeOldShards: typeof resharderCloseOldShards;
+    /** Handles checking if it is time to reshard and triggers the resharder */
```
itohatweb, line 191:
```suggestion
    /** Handles checking if it is time to reshard and triggers the resharder. */
```

src/ws/resharder.ts
```diff
@@ -1,31 +1,141 @@
...
-  }
+  const gateway = createGatewayManager({
+    ...oldGateway,
+    // IGNORE EVENTS FOR NOW
+    handleDiscordPayload: async function (_, data, shardId) {
+      if (data.t === "READY") {
+        const payload = data.d as DiscordReady;
+        for (const guild of payload.guilds) {
+          gateway.resharding.markNewGuildShardId(BigInt(guild.id), shardId);
+        }
```
itohatweb, line 21: instead of having this a function which updates only one guild id, it should take an array of guild ids and the new shard id

src/ws/resharder.ts
```diff
@@ -1,31 +1,141 @@
...
+      gateway.handleDiscordPayload = oldHandler;
+      oldGateway.handleDiscordPayload = function (_, data, shardId) {
+        // ALLOW EXCEPTION FOR CHUNKING TO PREVENT REQUESTS FREEZING
+        if (data.t !== "GUILD_MEMBERS_CHUNK") return;
+        oldHandler(_, data, shardId);
+      };
+
+      // STOP TIMER
+      clearInterval(timer);
+      await gateway.resharding.editGuildShardIds();
```
itohatweb, line 64: not sure whether this should be done like this.

src/ws/resharder.ts
```diff
@@ -1,31 +1,141 @@
...
+    if (!shard?.ready) {
+      return true;
+    }
+  }
+
+  return false;
+}
+
+/** Handler that by default closes all shards in the old gateway. Can be overriden if you have multiple servers and you want to communicate through redis pubsub or whatever you prefer. */
+export async function resharderCloseOldShards(oldGateway: GatewayManager) {
```
itohatweb, line 88: This will still cause a potential bug when users wait for a response from the gateway. Thoughts?

itohatweb (changes requested):

src/ws/resharder.ts
```diff
@@ -1,31 +1,141 @@
...
+    // TIMER TO KEEP CHECKING WHEN ALL SHARDS HAVE RESHARDED
+    const timer = setInterval(async () => {
+      const pending = await gateway.resharding.isPending(gateway, oldGateway);
+      // STILL PENDING ON SOME SHARDS TO BE CREATED
+      if (pending) return;
+
+      // ENABLE EVENTS ON NEW SHARDS AND IGNORE EVENTS ON OLD
+      const oldHandler = oldGateway.handleDiscordPayload;
+      gateway.handleDiscordPayload = oldHandler;
+      oldGateway.handleDiscordPayload = function (_, data, shardId) {
```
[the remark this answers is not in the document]
Skillz4Killz: This means its unused and makes deno lint and ts ignore errors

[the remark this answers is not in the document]
itohatweb: you use the `_` later as an argument

src/ws/resharder.ts
itohatweb: If the gateway gets closed, then the response from discord will never come.

itohatweb (approved):

PR #2005, opened 2022-02-07
fix: update dnt and settings

Skillz4Killz:


dnt.ts
```diff
@@ -1,16 +1,88 @@
...
+  },
+  entryPoints: [
+    "./mod.ts",
+    {
+      name: "./rest",
+      path: "src/rest/mod.ts",
+    },
+    {
+      name: "./gateway",
+      path: "src/gateway/mod.ts",
```
itohatweb, line 34:
```suggestion
      path: "src/ws/mod.ts",
```

dnt.ts
```diff
@@ -1,16 +1,88 @@
...
+      "discord-api",
+      "deno",
+      "discordeno",
+    ],
+    author: "Skillz4Killz",
+    license: "Apache License 2.0",
+    bugs: {
+      url: "https://github.com/discordeno/discordeno/issues",
+    },
+    homepage: "https://github.com/discordeno/discordeno#readme",
```
itohatweb, line 85: Maybe we should use `discordeno.mod.land` here?
```suggestion
    homepage: "https://discordeno.mod.land",
```

plugins/mod.ts
```diff
@@ -0,0 +1,4 @@
+export * from "./cache/mod.ts";
```
itohatweb, line 1: Remove this file from this PR

itohatweb (changes requested):

plugins/mod.ts
Skillz4Killz: This is important to this pr

itohatweb (approved):

PR #2022, opened 2022-02-09
style: deny changes to `deno.json`

itohatweb:


Skillz4Killz:
Not a fan of this

itohatweb:
Either this or I will remove the 120 with from the `deno.json` and then only use it for ignoring files.

Skillz4Killz (approved):

PR #2025, opened 2022-02-09
fix(types): add missing message component exports

Yaikava:


itohatweb (dismissed):

Yaikava:
Added textStyles export too

Yaikava:
> Can you check whether other files are missing too, and add them accordingly? (only for this mod.ts)

I think this is all. Did not find another one.

itohatweb (approved):

PR #2028, opened 2022-02-09
fix: add ending / in registry url

Skillz4Killz:


itohatweb:
Looks like the workflow already ran successfully. Closing this PR to prevent it from breaking.

PR #2030, opened 2022-02-10
update dockerfile and devcontainer

lts20050703:


itohatweb:
The defaults are there to provide a nice dev experience out of the box.

Issue #1973, opened 2022-02-01
[api-docs] Fix welcome screen required permissions (#4382)

Co-authored-by: devsnek <me@gus.host>


itohatweb:
A new commit was made into the api-docs repo.
https://github.com/discord/discord-api-docs/commit/c144941547e2e2207e0074e83255b966a313bc32

This is a bot created issue.

PR #2034, opened 2022-02-11
fix(plugins/permissions): getting disabled welcome screens requires `MANAGE_GUILD` 

Skillz4Killz:
Closes: https://github.com/discordeno/discordeno/issues/1973

### DIFF
--- a/plugins/permissions/src/guilds/welcomeScreen.ts
+++ b/plugins/permissions/src/guilds/welcomeScreen.ts
@@ -5,7 +5,10 @@ export function getWelcomeScreen(bot: BotWithCache) {
   const getWelcomeScreenOld = bot.helpers.getWelcomeScreen;
 
   bot.helpers.getWelcomeScreen = function (guildId) {
-    requireBotGuildPermissions(bot, guildId, ["MANAGE_GUILD"]);
+    const guild = bot.guilds.get(guildId);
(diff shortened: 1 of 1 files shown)


plugins/permissions/src/guilds/welcomeScreen.ts
```diff
@@ -1,6 +1,16 @@
 import { BotWithCache } from "../../deps.ts";
 import { requireBotGuildPermissions } from "../permissions.ts";
 
+export function getWelcomeScreen(bot: BotWithCache) {
+  const getWelcomeScreenOld = bot.helpers.getWelcomeScreen;
+
+  bot.helpers.getWelcomeScreen = function (guildId) {
+    requireBotGuildPermissions(bot, guildId, ["MANAGE_GUILD"]);
```
lts20050703, line 8: Closes: #1973 

## If the welcome screen is not enabled, the `MANAGE_GUILD` permission is required.

@Skillz4Killz Is there a way to check if the welcome screen is enabled or not? Because we need to check `MANAGE_GUILD` permission ONLY IF THE WELCOME SCREEN IS DISABLED, otherwise you can still get it WITHOUT the `MANAGE_GUILD` permission

lts20050703 (changes requested):

itohatweb (approved):

Issue #1913, opened 2022-01-25
[api-docs] clarify embed character limit (#4379)


itohatweb:
A new commit was made into the api-docs repo.
https://github.com/discord/discord-api-docs/commit/b3cc70e1d053464a962787e1637c10ddb4d83a0f

This is a bot created issue.

lts20050703:
```diff
- Additionally, the characters in all `title`, `description`, `field.name`, `field.value`, `footer.text`, and `author.name` fields must not exceed 6000 characters in total. Violating any of these constraints will result in a `Bad Request` response.
+ Additionally, the combined sum of characters in all `title`, `description`, `field.name`, `field.value`, `footer.text`, and `author.name` fields across all embeds attached to a message must not exceed 6000 characters. Violating any of these constraints will result in a `Bad Request` response.
```
Discord just clarify embed character limit
We can close this?

Skillz4Killz:
We can but we should open an issue in template repo where we have the embed class

lts20050703:
> We can but we should open an issue in template repo where we have the embed class

@Skillz4Killz The template is now here lol

lts20050703:
> We can but we should open an issue in template repo where we have the embed class

I have checked [template/utils/Embed.ts](https://github.com/discordeno/discordeno/blob/main/template/bigbot/src/utils/Embeds.ts)

Our implementation is correct.

<img width="245" alt="image" src="https://user-images.githubusercontent.com/87189679/153697562-03ed59ae-b1bf-466a-8ec9-22b3ab7f33a4.png">
<img width="428" alt="image" src="https://user-images.githubusercontent.com/87189679/153697568-95f56619-3d51-435a-bec5-d37d431c9e52.png">

all methods like setAuthor, setDescription, addField, etc... use fitData to make sure it's under the limit.

lts20050703:
> We can but we should open an issue in template repo where we have the embed class

@Skillz4Killz I have checked [template/utils/Embed.ts](https://github.com/discordeno/discordeno/blob/main/template/bigbot/src/utils/Embeds.ts)

Our implementation is correct to make sure it's under the limit.

<img width="245" alt="image" src="https://user-images.githubusercontent.com/87189679/153697562-03ed59ae-b1bf-466a-8ec9-22b3ab7f33a4.png">
<img width="428" alt="image" src="https://user-images.githubusercontent.com/87189679/153697568-95f56619-3d51-435a-bec5-d37d431c9e52.png">

[...]

Issue #1882, opened 2022-01-02
Validation plugin


itohatweb:
- [ ] Remove all validations from Discordeno
- [ ] Remove text validations from the permissions plugin
- [ ] Create a validation plugin

lts20050703:
cross reference to "duplicates"
#1907 
#1908 
#1913 
#2094 
#2137

lts20050703:
Closes: #2137
Closes: #2094
Closes: #1908

PR #2040, opened 2022-02-11
feat: validation plugin

Skillz4Killz:
The validation plugin is a plugin that will validate the args provided in the helpers. Note this will not validate with cache so this can be useful for bots without cache as well. The bots that use cache should also enable the permissions plugin which can add additional validation checks which require cached objects


lts20050703:
Does this closes #1882 ?

Skillz4Killz:
No this is only part of #1882

lts20050703:
If possible can you try also close #1913 ?

Skillz4Killz:
RIP THIS IS WHAT HAPPENS WHEN PRS DON"T GET MERGED

PR #2050, opened 2022-02-14
update CONTRIBUTING guidelines

lts20050703:


itohatweb:
I will completely rewrite this. Do not use this as any reference for now.

PR #2054, opened 2022-02-14
fix(types, helpers)!: EditMessage helper transformer & change `Embed` to `DiscordenoEmbed`

lts20050703:
Note: I changed `transformer/embed.ts` because I looked at `helpers/messages/editMessage.ts` and see `provider: embed.provider` `fields: embed.fields` because both of them's properties are only one word therefore no camelCase to snake_case manual conversion needed

### DIFF
--- a/transformers/embed.ts
+++ b/transformers/embed.ts
@@ -42,12 +42,7 @@ export function transformEmbed(bot: Bot, payload: SnakeCasedPropertiesDeep<Embed
         width: payload.video.width,
       }
       : undefined,
-    provider: payload.provider
-      ? {
-        name: payload.provider.name,
-        url: payload.provider.url,
-      }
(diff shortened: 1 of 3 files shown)


transformers/embed.ts
```diff
@@ -42,12 +42,7 @@ export function transformEmbed(bot: Bot, payload: SnakeCasedPropertiesDeep<Embed
         width: payload.video.width,
       }
       : undefined,
-    provider: payload.provider
```
itohatweb, line 45: Unrelated change please do this in a different PR.
I want to stick to the 1 PR 1 change schema.

itohatweb (changes requested):

Skillz4Killz (approved):

PR #2066, opened 2022-02-17
fix(types): `SelectOption#default` is not required

itohatweb:


lts20050703 (approved):
LGTM!

Reference: <https://discord.com/developers/docs/interactions/message-components#select-menu-object-select-option-structure>

Skillz4Killz (approved):

PR #2077, opened 2022-02-26
Fix typo

Amatsagu:
Emoji objects refer always to a custom guild emoji, not emoticons!
`id`, `name` & `animated` are always present, no matter in which context gateway/api sent emoji object.

### DIFF
--- a/types/emojis/emoji.ts
+++ b/types/emojis/emoji.ts
@@ -3,9 +3,9 @@ import { User } from "../users/user.ts";
 /** https://discord.com/developers/docs/resources/emoji#emoji-object-emoji-structure */
 export interface Emoji {
   /** Emoji id */
-  id?: string;
+  id: string;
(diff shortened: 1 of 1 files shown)


Skillz4Killz:
This is not true, refer to discord api docs.

Amatsagu:
> 

It is, discord docs are made nearly fully by community and it lacks a lot of info. Some of the topics are well documented and some other were not touched for like 3 years. Emoji object that discord gateway/api sends back always refer to a custom emoji. Characters like ✨ or 💸 are emoticons which are nothing else than a fancy text which is supported by like the most of the web (standard). Discord renders then a bit differently cause of using twitter emoji pack for that instead using system assets.

Skillz4Killz:
If you feel the docs are wrong, please PR the docs first. We go by what the docs say

PR #2079, opened 2022-02-28
fix(gateway): rate-limit handling

### LINKED ISSUE #1886: Gateway Ratelimiting
- [ ] Calculate how many requests are reserved by heartbeats
- [ ] Reserve the same amount for Hearbeat requests from the gateway
- [ ] Queue should wait until the rate limit is over not one minute

Skillz4Killz:
This PR will close #1886 

It improves on some minor but crucial aspects of rate limiting in the gateway side.

- Instead of hardcoding 120/m we now make it dynamic so that bots can override in case discord will bump their max amount or such.
- Instead of hardcoding the reserved requests for heartbeating we now dynamically calculate this based on the interval discord will provide. This protects against any possible change discord may decide to change the heartbeat interval.
- Optimized the delay. Before when you hit the max, we just hardcoded a 1 min wait to reset the timer since discord didn't provide a good value to know. Now we will reset based on the amount of time left since the first request was sent.
- Changed gateway/ws.ts to gateway/shard.ts. The filename was just weird.

### DIFF
--- a/gateway/resume.ts
+++ b/gateway/resume.ts
@@ -43,6 +43,7 @@ export function resume(gateway: GatewayManager, shardId: number) {
     processingQueue: false,
     queueStartedAt: Date.now(),
     queueCounter: 0,
+    safeRequestsPerShard: oldShard.safeRequestsPerShard || 120,
   });
 
   // Resume on open
--- a/gateway/safeRequestsPerShard.ts
+++ b/gateway/safeRequestsPerShard.ts
@@ -1,0 +1,9 @@
+import { GatewayManager } from "./gateway_manager.ts";
+import { DiscordenoShard } from "./shard.ts";
+
+export function safeRequestsPerShard(gateway: GatewayManager, shard: DiscordenoShard) {
+  // * 2 adds extra safety layer for discords OP 1 requests that we need to respond to
+  const safeRequests = gateway.maxRequestsPerInterval -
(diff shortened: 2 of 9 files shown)


gateway/resume.ts
```diff
@@ -43,6 +43,7 @@ export function resume(gateway: GatewayManager, shardId: number) {
     processingQueue: false,
     queueStartedAt: Date.now(),
     queueCounter: 0,
+    safeRequestsPerShard: oldShard.safeRequestsPerShard || 120,
```
lts20050703, line 46: idk what I am reading, maybe `??` ?

gateway/safeRequestsPerShard.ts
```diff
@@ -0,0 +1,9 @@
+import { GatewayManager } from "./gateway_manager.ts";
+import { DiscordenoShard } from "./shard.ts";
+
+export function safeRequestsPerShard(gateway: GatewayManager, shard: DiscordenoShard) {
+  // * 2 adds extra safety layer for discords OP 1 requests that we need to respond to
+  const safeRequests = gateway.maxRequestsPerInterval -
+    Math.ceil(gateway.queueResetInterval / shard.heartbeat.interval) * 2;
```
lts20050703, line 7: idk what I am reading, maybe Math.floor?

itohatweb (approved):

PR #2126, opened 2022-03-23
ci: cache template code

itohatweb:
This PR will break tests!!
As long as the templates are not fixed on the main branch.


Skillz4Killz (approved):
I woulda preferred if the folder name was `templates` but rest looks good

Issue #2125, opened 2022-03-23
[api-docs] Slash command localization (#4653)


itohatweb:
A new commit was made into the api-docs repo: https://github.com/discord/discord-api-docs/commit/931af47f253369d004e1c56476f4c40a3af9dc09

* Document slash command localization

* fix links?

* prettify tables

* some style and functionality updates

* style changes

This is a bot created issue.

PR #2129, opened 2022-03-23
feat(helpers,types)!: slash command localization

itohatweb:
Discord has documented the slash command localization feature now.
This adds the full functionality for those to Discordeno.

Additionally a `Locales` has also been added to allow for better typing.

Reference: https://github.com/discord/discord-api-docs/pull/4653

Closes: #2125


helpers/interactions/commands/getApplicationCommand.ts
```diff
@@ -2,14 +2,23 @@ import type { Bot } from "../../../bot.ts";
...
+export async function getApplicationCommand(bot: Bot, commandId: bigint, options?: GetApplicationCommand) {
   const result = await bot.rest.runMethod<DiscordApplicationCommand>(
     bot.rest,
     "get",
-    guildId
-      ? bot.constants.endpoints.COMMANDS_GUILD_ID(bot.applicationId, guildId, commandId)
+    options?.guildId
+      ? bot.constants.endpoints.COMMANDS_GUILD_ID(bot.applicationId, options.guildId, commandId)
       : bot.constants.endpoints.COMMANDS_ID(bot.applicationId, commandId),
+    options,
```
Skillz4Killz, line 12: Need to make this snake case here

Skillz4Killz (dismissed):

Skillz4Killz:
Consider adding a huge warning not to use this:

**If you want to use the guild_locale on interactions:**
Only guilds with Community enabled can set guild locale so for all others ull get en_US which isn't accurate. I would wager most servers are non-community but maybe im wrong.

**If you want to use the user locale on interactions:**
- Privacy nightmare of you exposing a users language/country should they not want this exposed. Imagine your in a server and use x command and suddenly get a bot responding in your language for all to see.
- This means the only time u should use this is ephemerals and that means half the time your responding to them in 1 language and the other time your in another language. Bad consistency.
- Some servers have rules in fact of only allowing a certain language like English only. Now your bot will start breaking server moderation rules and sending messages in others languages.

Skillz4Killz (approved):

Issue #2123, opened 2022-03-22
[api-docs] Modal title max length (#4649)


itohatweb:
A new commit was made into the api-docs repo: https://github.com/discord/discord-api-docs/commit/4bebe4c662650a5b79648992397b4633feffbd64
No details given.

This is a bot created issue.

PR #2130, opened 2022-03-23
feat(plugins): validate interaction response title

itohatweb:
Modal titles must have a length between 1 and 45 characters.
This adds a check to the permissions plugin.

Reference: https://github.com/discord/discord-api-docs/pull/4649

Closes: #2123

### DIFF
--- a/plugins/permissions/src/interactions/mod.ts
+++ b/plugins/permissions/src/interactions/mod.ts
@@ -6,6 +6,14 @@ export function sendInteractionResponse(bot: BotWithCache) {
   const sendInteractionResponseOld = bot.helpers.sendInteractionResponse;
 
   bot.helpers.sendInteractionResponse = function (id, token, options) {
+    if (options.data?.title !== undefined) {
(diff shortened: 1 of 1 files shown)


plugins/permissions/src/interactions/mod.ts
```diff
@@ -6,6 +6,14 @@ export function sendInteractionResponse(bot: BotWithCache) {
   const sendInteractionResponseOld = bot.helpers.sendInteractionResponse;
 
   bot.helpers.sendInteractionResponse = function (id, token, options) {
+    if (options.data?.title !== undefined) {
```
Skillz4Killz, line 9:
```suggestion
    if (options.data?.title) {
```

itohatweb: What if title is `""`

Skillz4Killz: oh i see. TBH i woulda preferred .title?.length >= 0 but whatever this works also

Skillz4Killz (approved):

Issue #2097, opened 2022-03-12
[transformer] need a transformTemplate


Skillz4Killz:
Transformer that converts a guild template object

PR #2132, opened 2022-03-23
feat(helpers,types)!: add guild template transformer

itohatweb:
This adds a template transformer. Also this fixes incorrect type definitions regarding templates.

Breaking Change: The previously called `Template` interface which was used for creating templates has been renamed to `CreateTemplate`

Closes: #2097

### DIFF
--- a/helpers/templates/createGuildTemplate.ts
+++ b/helpers/templates/createGuildTemplate.ts
@@ -4,7 +4,7 @@ import { User } from "../../transformers/member.ts";
 import { DiscordTemplate } from "../../types/discord.ts";
 
 /** Creates a template for the guild. Requires the `MANAGE_GUILD` permission. */
-export async function createGuildTemplate(bot: Bot, guildId: bigint, data: Template) {
+export async function createGuildTemplate(bot: Bot, guildId: bigint, data: CreateTemplate) {
   if (data.name.length < 1 || data.name.length > 100) {
     throw new Error("The name can only be in between 1-100 characters.");
   }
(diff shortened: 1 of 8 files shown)


helpers/templates/createGuildTemplate.ts
```diff
@@ -14,41 +14,14 @@ export async function createGuildTemplate(bot: Bot, guildId: bigint, data: Templ
   }
 
   return await bot.rest.runMethod<DiscordTemplate>(bot.rest, "post", bot.constants.endpoints.GUILD_TEMPLATES(guildId), {
```
itohatweb, line 16: just use `data` there

helpers/templates/createGuildTemplate.ts
```diff
@@ -13,42 +11,17 @@ export async function createGuildTemplate(bot: Bot, guildId: bigint, data: Templ
...
-    updated_at: data.updatedAt,
-    source_guild_id: data.sourceGuildId,
-    serialized_source_guild: data.serializedSourceGuild,
-    is_dirty: data.isDirty,
-  });
+  return await bot.rest.runMethod<DiscordTemplate>(
+    bot.rest,
+    "post",
+    bot.constants.endpoints.GUILD_TEMPLATES(guildId),
+    data,
```
Skillz4Killz, line 18: need to be snake cased

itohatweb: No it does not?

Skillz4Killz: Why not? ur sending this camelcase to discord? discord accepts camel case for this endpoint?

itohatweb: Look at the type, there is no camelcase

Skillz4Killz (approved):

PR #2159, opened 2022-04-07
fix: prevent crashing the rest process

dev-viinz:
There where instances where some requests crashed the rest process. I found that checking for a body pretty much fixed it. 
There is probably a better way to do this, please improve this if you have an idea :)

### DIFF
--- a/template/bigbot/src/rest/mod.ts
+++ b/template/bigbot/src/rest/mod.ts
@@ -42,13 +42,7 @@ async function handleRequest(conn: Deno.Conn) {
       );
     }
 
-    // deno-lint-ignore no-explicit-any
-    let json: any;
-    if (requestEvent.request.body) {
-        json = (await requestEvent.request.json());
-    } else {
-        json = undefined
-    }
+    const json = requestEvent.request.body ? (await requestEvent.request.json()) : undefined;
 
     try {
       const result = await rest.runMethod(
--- a/template/bigbot/src/rest/mod.ts
+++ b/template/bigbot/src/rest/mod.ts
@@ -47,7 +47,7 @@ async function handleRequest(conn: Deno.Conn) {
     if (requestEvent.request.body) {
         json = (await requestEvent.request.json());
     } else {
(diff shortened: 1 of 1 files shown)


itohatweb:
Hey! Can you make json undefined instead of an empty string?

itohatweb:
Hey! Can you make json undefined instead of an empty string?

itohatweb (changes requested):
Hey! Can you make json undefined instead of an empty string?

template/bigbot/src/rest/mod.ts
```diff
@@ -42,7 +42,13 @@ async function handleRequest(conn: Deno.Conn) {
       );
     }
 
-    const json = (await requestEvent.request.json());
+    // deno-lint-ignore no-explicit-any
+    let json: any;
+    if (requestEvent.request.body) {
+        json = (await requestEvent.request.json());
+    } else {
+        json = ""
```
lts20050703, line 50:
```suggestion
        json = undefined
```

lts20050703 (changes requested):

template/bigbot/src/rest/mod.ts
dev-viinz: Of course, thx for the suggestion xD

template/bigbot/src/rest/mod.ts
```diff
@@ -42,7 +42,13 @@ async function handleRequest(conn: Deno.Conn) {
       );
     }
 
-    const json = (await requestEvent.request.json());
+    // deno-lint-ignore no-explicit-any
+    let json: any;
```
Skillz4Killz, line 46: Why no ternary like
const json = requestEvent.request.body ? await ..... : undefined;

dev-viinz: You're right, that would be much nicer!

Skillz4Killz (approved):

itohatweb (approved):

PR #2174, opened 2022-04-28
Proxy delete channel tests

Skillz4Killz:
This PR introduces a new style of doing unit tests. Since our library is oriented towards big bot developers, our unit tests should also be oriented towards using proxies. This PR introduces a proxy REST and adds the first test to showcase that this design works.

The general idea of the new design is to eliminate the need of having startBot() and leaking ops in tests. This will allow us to run all our tests in parallel as well as check any leaks. Each test will be entirely standalone and run in entirely separate contexts.

**Main Concerns:**

- Since this no longer uses the internal REST code, we need to remember to update the PROXY when REST code undergoes major changes.
- Currently this PR does not introduce the Gateway PROXY. I am still unsure how to implement that at this time.

**Pending Discussions:**

Everything is still open to discussion and changing at this time. This is mainly just to see if the idea works, which it does,. For example, the file structure and file names are open to changes.

---

HUGE THANKS TO @TriForMine

### DIFF
--- a/testss/utils.ts
+++ b/testss/utils.ts
@@ -1,0 +1,21 @@
+export const CACHED_COMMUNITY_GUILD_ID = 907350958810480671n;
+
+export function delayUntil(
+  maxMs: number,
+  isReady: () => boolean | undefined | Promise<boolean | undefined>,
+  timeoutTime = 100,
+): Promise<void> {
+  const maxTime = Date.now() + maxMs;
+
+  async function hackyFix(resolve: () => void) {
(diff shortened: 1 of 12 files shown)


testss/utils.ts
```diff
@@ -0,0 +1,21 @@
+export const CACHED_COMMUNITY_GUILD_ID = 907350958810480671n;
```
lts20050703, line 1: image
Instead of hard code, consider Deno.env.get("CACHED_COMMUNITY_GUILD_ID")

PR #2189, opened 2022-05-05
Fixing the BigBot Template

Reboot-Codes:
You might want to cherry pick, and remove the reference to `create-discordeno-bot` if you would like. However, the main change here is adding support for deno tasks for the bigbot template. These changes work as of deno v1.20.3.


Reboot-Codes:
Fixed the parsing of the `DISCORD_TOKEN` env variable.

Reboot-Codes:
Added a fix to the `gateway` workers so now they actually work.

Skillz4Killz (approved):

PR #2195, opened 2022-05-11
update createGuild

lts20050703:
icon should be bigint then converted to hash


itohatweb (commented):
This is invalid.
You implicitly upload a new image by sending an encoded base64 string here not an image hash.

Issue #2166, opened 2022-04-15
[api-docs] Add default_auto_archive_duration to Create Guild Channel endpoint (#4767)


itohatweb:
A new commit was made into the api-docs repo: https://github.com/discord/discord-api-docs/commit/032dec6c5680c9c249f3ab403c920e5cbb2d9b50

It was missing in #3108 and #3126

This is a bot created issue.

PR #2202, opened 2022-05-12
Add default_auto_archive_duration to createChannel helper options

lts20050703:
Closes: #2166 
Upstream: https://github.com/discord/discord-api-docs/commit/032dec6c5680c9c249f3ab403c920e5cbb2d9b50

### DIFF
--- a/helpers/channels/createChannel.ts
+++ b/helpers/channels/createChannel.ts
@@ -30,6 +30,7 @@ export async function createChannel(bot: Bot, guildId: bigint, options?: CreateG
         })),
         type: options?.type || ChannelTypes.GuildText,
         reason,
+        default_auto_archive_duration: options?.defaultAutoArchiveDuration,
       }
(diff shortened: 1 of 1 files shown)


helpers/channels/createChannel.ts
```diff
@@ -58,4 +59,6 @@ export interface CreateGuildChannel {
   parentId?: bigint;
   /** Whether the channel is nsfw */
   nsfw?: boolean;
+  /** the default duration that the clients use (not the API) for newly created threads in the channel, in minutes, to automatically archive the thread after recent activity */
```
itohatweb, line 62:
```suggestion
  /** Default duration (in minutes) that clients (not the API) use for newly created threads in this channel, to determine when to automatically archive the thread after the last activity */
```

itohatweb (changes requested):

Issue #2155, opened 2022-03-31
[api-docs] localization updates (#4724)


itohatweb:
A new commit was made into the api-docs repo: https://github.com/discord/discord-api-docs/commit/2f854d3fd675a9953d1bb8d6858a03c8a9642337
No details given.

This is a bot created issue.

PR #2206, opened 2022-05-12
localization updates

lts20050703:
Closes: #2155 

Upstream: https://github.com/discord/discord-api-docs/commit/2f854d3fd675a9953d1bb8d6858a03c8a9642337

Types: DiscordApplicationCommmand and DiscordApplicationCommandOption: Update documentation

Types: DiscordActivityButton: Add documentation link

Helper: upsertApplicationCommand: All fields are optional, but any fields provided will entirely overwrite the existing values of those fields.


types/discord.ts
```diff
@@ -924,8 +924,7 @@ export interface DiscordActivitySecrets {
   match?: string;
 }
 
-// https://github.com/discord/discord-api-docs/pull/2219
-// TODO: add documentation link
+/** https://discord.com/developers/docs/topics/gateway#activity-object-activity-buttons */
```
lts20050703, line 927: Tell me if I should move this to another PR

types/shared.ts
```diff
@@ -1285,53 +1285,6 @@ export type Camelize<T> = {
...
-//   & {
-//     [K in Exclude<keyof T, KeysWithUndefined<T>>]: T[K] extends object ? Object extends Pick<T[K], keyof T[K]> ? T[K]
-//     : T[K] extends Collection<any, any> ? T[K]
-//     : T[K] extends any[] ? T[K]
-//     : Optionalize<T[K]>
-//       : T[K];
-//   }
-// )
-//   : T;
-
```
lts20050703, line 1334: tell me if I should move this to another PR

helpers/interactions/commands/createApplicationCommand.ts
```diff
@@ -89,7 +89,7 @@ export interface CreateContextApplicationCommand {
 }
 
 export function isContextApplicationCommand(
-  cmd: CreateContextApplicationCommand | CreateApplicationCommand,
-): cmd is CreateContextApplicationCommand {
+  cmd: Partial<CreateContextApplicationCommand> | Partial<CreateApplicationCommand>,
+): cmd is Partial<CreateContextApplicationCommand> {
```
itohatweb, line 93: I do not like this change

types/discord.ts
itohatweb: I think this is the answer 🙃

types/shared.ts
itohatweb: I think this is the answer 🙃

types/discord.ts
```diff
@@ -1645,7 +1644,7 @@ export interface DiscordApplicationCommand {
   options?: DiscordApplicationCommandOption[];
   /** Whether the command is enabled by default when the app is added to a guild */
   default_permission?: boolean;
-  /** The type of command. By default this is a application command(ChatInput). */
+  /** The type of command. By default this is a application command (ApplicationCommandTypes.ChatInput). */
```
itohatweb, line 1647: I do not think we need this change in this case.

types/discord.ts
```diff
@@ -1633,7 +1632,7 @@ export interface DiscordApplicationCommand {
   application_id: string;
   /** Guild id of the command, if not global */
   guild_id?: string;
-  /** 1-32 character name matching */
+  /** `ApplicationCommandTypes.ChatInput` command names must match the following regex `^[-_\p{L}\p{N}\p{sc=Deva}\p{sc=Thai}]{1,32}$` with the unicode flag set. If there is a lowercase variant of any letters used, you must use those. Characters with no lowercase variants and/or uncased letters are still allowed. `ApplicationCommandTypes.User` and `ApplicationCommandTypes.Message` commands may be mixed case and can include spaces. */
```
itohatweb, line 1635: There are more interfaces affected by this change. Please update all of them. 
Also update the permission plugin accordingly.

itohatweb (changes requested):

helpers/interactions/commands/createApplicationCommand.ts
lts20050703: @itohatweb Why? The Docs stated that **All fields are optional**, but any fields provided will entirely overwrite the existing values of those fields.

itohatweb: I feel like it's more the name of the function itself which is wrong here.

types/discord.ts
lts20050703: updated regex

helpers/interactions/commands/createApplicationCommand.ts
lts20050703: How should I fix this?

lts20050703: resolved in discord. 
> just ignore
> let this be
> what u have rn
> let this be how it is now
> yes
> like i said
> leave this

Skillz4Killz (approved):

PR #2217, opened 2022-05-13
fix(types): use snowflake transformer instead of direct `BigInt` call

meister03:


itohatweb (approved):

lts20050703 (approved):
LGTM!

lts20050703:
@meister03 Did you deno fmt?
image

PR #2218, opened 2022-05-13
[Proposal] for set Bot Status for specific Shard

meister03:
Open for other ideas....


itohatweb:
This is already implemented in a slightly different approach (using functions) in the gateway rewrite.

PR #2222, opened 2022-05-14
fix(rest,types): `getArchivedThreads` `before` option and wrong return type

lts20050703:
# Check #2219 first!


types/discord.ts
```diff
@@ -1368,6 +1368,11 @@ export interface DiscordListActiveThreads {
   members: DiscordThreadMember[];
 }
 
+export interface DiscordListArchivedThreads extends DiscordListActiveThreads {
+  /** whether there are potentially additional threads that could be returned on a subsequent call */
```
itohatweb, line 1372:
```suggestion
  /** Whether there are potentially additional threads that could be returned on a subsequent call */
```

helpers/channels/threads/getArchivedThreads.ts
```diff
@@ -42,14 +39,14 @@ export async function getArchivedThreads(
...
 
-// TODO: add docs link
-export interface ListPublicArchivedThreads {
-  // TODO: convert unix to ISO9601 timestamp
-  /** Returns threads before this timestamp. UNIX or ISO8601 timestamp */
-  before?: number | string;
+/** https://discord.com/developers/docs/resources/channel#list-public-archived-threads-query-string-params */
+export interface ListArchivedThreads {
+  /** Returns threads before this timestamp */
+  before?: number;
```
itohatweb, line 49: Has this been tested?

itohatweb (changes requested):

helpers/channels/threads/getArchivedThreads.ts
lts20050703: My bad I forgot to convert it to ISO string

itohatweb (approved):

Issue #2213, opened 2022-05-12
[api-docs] Separate GUILD_CREATE fields from Guild object (#4888)


itohatweb:
A new commit was made into the api-docs repo: https://github.com/discord/discord-api-docs/commit/78a152e67b3eb8a7f44b998d647fb66ce5831ac4

* Separate GUILD_CREATE fields from Guild object

* Change members and presences fields to be required in GUILD_CREATE

* Reword notice of moved guild object fields

Co-authored-by: Shay DeWael <swdewael@gmail.com>

Co-authored-by: Shay DeWael <swdewael@gmail.com>

This is a bot created issue.

PR #2224, opened 2022-05-14
refactor(handlers,types): separate GUILD_CREATE fields from Guild object

lts20050703:
Closes: https://github.com/discordeno/discordeno/issues/2213
Upstream: https://github.com/discord/discord-api-docs/commit/78a152e

types: Add DiscordGuildCreate interface
transformers: Create GuildCreateToggles and transformGuildCreate transformer
handlers: handleGuildCreate use the transformGuildCreate transformer


transformers/guildCreate.ts
```diff
@@ -0,0 +1,110 @@
...
+import { Bot } from "../bot.ts";
+import { Collection } from "../util/collection.ts";
+import { DiscordGuildCreate } from "../types/discord.ts";
+import { Optionalize } from "../types/shared.ts";
+import { GuildCreateToggles } from "./toggles/guild.ts";
+
+export function transformGuildCreate(bot: Bot, payload: { guild: DiscordGuildCreate } & { shardId: number }) {
+  const guildId = bot.transformers.snowflake(payload.guild.id);
+
+  const guild = {
```
itohatweb, line 11: Please do `...transformGuild(payload)` and then just add the other fields

transformers/toggles/guild.ts
```diff
@@ -94,6 +92,15 @@ export class GuildToggles extends ToggleBitfieldBigint {
...
   }
+}
+
+export class GuildCreateToggles extends GuildToggles {
+  constructor(guild: DiscordGuildCreate) {
+    super(guild);
+
+    if (guild.large) this.add(GuildToggle.large);
+    if (guild.unavailable) this.add(GuildToggle.unavailable);
+  }
```
itohatweb, line 103: Move to separate file

itohatweb (changes requested):
Please revert the unrelated format changes

itohatweb (approved):

Skillz4Killz:
Not a fan of this PR. I don't want to go this route where we end up having like 20 different transformers for a channel because each has a few different props on it.

itohatweb:
We won't do that for channels, since TypeScript does not provide anything what we could use and still maintain a good UX.
But it is different in this case. Why not say through types that eg. `channels` never exists when you use the `getGuild` helper.

Skillz4Killz:
That can be done with types without having to create multiple transformers for each type.

```ts
function getGuild() {
  return {} as SpecialGuildInterface;
}
```

Issue #2133, opened 2022-03-24
Add OAuth2


itohatweb:
OAuth2 is very cool stuff, therefore we should support it.

- [ ] endpoints
    - [ ] https://discord.com/developers/docs/resources/user#get-user-connections
    - [ ] #2750

PR #2227, opened 2022-05-15
fix(helpers,transformers,types): fix `DiscordConnection` key casing, add `transformIntegrationCreateUpdate` transformer, add `getUserConnections` helper

lts20050703:
Closes: #2133 
https://discord.com/developers/docs/resources/user#get-user-connections
fix DiscordConnection interface keys not in snake case
add transformConnection interface
add transformIntegrationCreateUpdate transformer
Fix it so that the transformIntegrationCreateUpdate transformer is used for both INTEGRATION_UPDATE and getIntegrations


Skillz4Killz:
Not a fan of this pr, too much going on here. Some make sense but others do not which makes it hard to merge this PR. Some changes require explanation like why a new transformer was added and why it is different than the integration one

PR #2232, opened 2022-05-20
Add forceSet option for collection

meister03:
Ignores maxSize
```js
collection.set('292892', {name: 'test'}, {forceSet: true})
```

### DIFF
--- a/util/collection.ts
+++ b/util/collection.ts
@@ -46,12 +46,16 @@ export class Collection<K, V> extends Map<K, V> {
     this.startSweeper({ filter: newFilter, interval: this.sweeper.interval });
   }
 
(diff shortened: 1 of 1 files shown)


itohatweb:
Elaborate when this would be used

meister03:
For instance the member collection can be capped to 0. During allowing to save just the bot guild member...

Or when you fetch something and want to keep it for further processing...

itohatweb:
For the 1st there will be a cache update which improves that behaviour.
For the 2nd I'm not sure why you would use a maxed collection for that?

meister03:
When you usually do not want to cache the resource type, but need it for some commands.

Skillz4Killz (approved):

PR #2233, opened 2022-05-20
More BigBot Fixes

Reboot-Codes:
I found an issue where the bot never revived any events from discord. Turns out the workers were making http requests to `${EVENT_HANDLER_URL}:${EVENT_HANDLER_PORT}` instead of `http://${EVENT_HANDLER_URL}:${EVENT_HANDLER_PORT}` like all other http requests. It makes sense to require the same formatting of url for all environment variables. This could possibly be done with a sort of check using `try { new URL('${EVENT_HANDLER_URL}:${EVENT_HANDLER_PORT}') } catch (e) { log.error("Check url fomatting for the EVENT_HANDLER_URL key in .env!") }` but that seems slightly repetitive and this is a smaller fix which keeps continuity.

I also added some more comments into `.env.example` and working defaults to simplify DX. 

Next on the list might be to include the automatic importing of commands. E.g. just create a new file for a new command exporting a `Command`. I found a way to do this without using the `fileloader-plugin` which is a very hacky solution. Instead it would just scan the `slash` directory for commands and add those to the commands map.


itohatweb (changes requested):
Please run `deno fmt` to make the tests pass.

Reboot-Codes:
`deno fmt ./template/bigbot` was run, apparently that doesn't make the correct changes to make the tests pass.

Reboot-Codes:
FYI, for future PRs, etc; run `deno fmt **/changed-dir/* --config ./deno.json` to make the formatting check pass. Anyways, all fixed @itohatweb.

itohatweb (approved):

itohatweb:
ig good for now

PR #2281, opened 2022-06-13
fixes intents problem

Tomato6966:


Tomato6966:
its important that examples work at least..

itohatweb:
examples are not updated by me, as long as I'm not finished with the main lib

PR #2285, opened 2022-06-13
typings: change any/unknown to void

nsylke:
There is some inconsistency on the EventHandlers interface, hoping we can use `void` instead since there is no reason to return an object on these events anyways.

### DIFF
--- a/bot.ts
+++ b/bot.ts
@@ -483,15 +483,15 @@ export function createTransformers(options: Partial<Transformers>) {
 export type RestManager = ReturnType<typeof createRestManager>;
 
 export interface EventHandlers {
-  debug: (text: string, ...args: any[]) => unknown;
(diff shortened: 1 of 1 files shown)


Skillz4Killz:
Big fan of `unknown`. Not a fan of `void`. When they return void, the code in end users events become much worse. For example:

```ts
// Will error with void
messageCreate: (message) => {
  if (!message.guildId) return await bot.helpers.sendTextMessage("I only work in guilds.");
}

// void requires it to be written like this
messageCreate: (message) => {
  if (!message.guildId) {
    await bot.helpers.sendTextMessage("I only work in guilds.");
    return;
  }
}
```

I just find the code much uglier with `void`. Unknown was good choice!

Skillz4Killz (approved):

PR #2310, opened 2022-06-21
Update discord.ts

meister03:
Add disabled prop


Tomato6966:
you forgot to add it for interactions too...

Tomato6966:
> > you forgot to add it for interactions too...
> > They use the reverse transformers

wdym by that, it only works for me if i change it also in the interactions ..

PR #2314, opened 2022-07-01
Update minimal template

meister03:
Important Note: CachePlugin has to be updated on deno.land


itohatweb:
I'm not sure about the PR title 🤔

meister03:
> I'm not sure about the PR title 🤔

my terminal was lagging :|

Yaikava:
Why did you change it dotenv to use `.env.example`?
That is wrong imo.

meister03:
> Why did you change dotenv to use `.env.example`? That is wrong imo.

it worked with that and people can change it in a sec, when they want  a other name for their env

Yaikava:
> Important Note: CachePlugin has to be updated on deno.land

I think you are supposed to import it from main module.

Yaikava:
> > Why did you change dotenv to use `.env.example`? That is wrong imo.
> 
> it worked with that and people can change it in a sec, when they want a other name for their env

This way secrets can get into git easily by accident. They can change it yeah, but it is bad idea imo.

template/minimal/configs.ts
```diff
@@ -1,6 +1,5 @@
 import { dotEnvConfig } from "./deps.ts";
 
-dotEnvConfig({ export: true });
-
+dotEnvConfig({ export: true, path: './.env.example' });
```
lts20050703, line 3:
```suggestion
dotEnvConfig({ export: true });
```

lts20050703: Don't use .env.example

lts20050703:

Skillz4Killz (approved):

PR #2315, opened 2022-07-01
Fix beginner template

meister03:


lts20050703:

PR #2388, opened 2022-08-23
Fix cache plugin doesn't update voice state update

lts20050703:
Thanks @Yaikava  for the code


Skillz4Killz (approved):

Skillz4Killz:
This seems like it is only adding to cache and never removing them. We need to remove them when channelId is null i imagine and also when the channel is deleted or the member has left the server or banned etc....

Yaikava:
> This seems like it is only adding to cache and never removing them. We need to remove them when channelId is null i imagine and also when the channel is deleted or the member has left the server or banned etc....

We can just delete it if channelId is null

lts20050703:
@Skillz4Killz Can you review my delete from cache code?
In my testing when I delete a voice channel that I'm in discord fire/emit channel delete event AND voice state update event, with channel_id undefined

Skillz4Killz (approved):

PR #2395, opened 2022-08-27
feat!: Add return types to helper functions.

vxern:
Additional changes:

- Make helper function code more consistent: The variables storing the result of an API endpoint being contacted have all been renamed to 'result' for a single result or 'results' for an array of results.
- Use standard naming techniques: Several files were missing the 'get' keyword. The 'remove' keyword has been replaced with 'delete' to correspond to both the Discord API and the rest of the code.


helpers/channels/deleteChannel.ts
```diff
@@ -1,9 +1,8 @@
 import type { Bot } from "../../bot.ts";
-import { DiscordChannel } from "../../types/discord.ts";
 
 /** Delete a channel in your server. Bot needs MANAGE_CHANNEL permissions in the server. Bot needs MANAGE_THREADS permissions in the server if deleting thread. */
-export async function deleteChannel(bot: Bot, channelId: bigint, reason?: string) {
-  await bot.rest.runMethod<DiscordChannel>(
+export async function deleteChannel(bot: Bot, channelId: bigint, reason?: string): Promise<void> {
+  return void await bot.rest.runMethod(
```
Skillz4Killz, line 5: What is the reasoning to add return void here?

helpers/discovery/editDiscovery.ts
```diff
@@ -16,7 +21,7 @@ export async function editDiscovery(bot: Bot, guildId: bigint, data: ModifyGuild
 
   return {
     guildId,
-    primaryCategoryId: result.primary_category_id,
+    primaryCategoryId: BigInt(result.primary_category_id),
```
Skillz4Killz, line 24: Are u sure this is a snowflake and not a categor name?

helpers/discovery/getDiscovery.ts
```diff
@@ -11,7 +21,7 @@ export async function getDiscovery(bot: Bot, guildId: bigint) {
 
   return {
     guildId,
-    primaryCategoryId: result.primary_category_id,
+    primaryCategoryId: BigInt(result.primary_category_id),
```
Skillz4Killz, line 24: Confirm this is indeed a snowflake and if it is use the transformers.snowflake

helpers/emojis/getEmojiUrl.ts
```diff
@@ -1,6 +1,6 @@
 import { Bot } from "../../bot.ts";
 
 /** Creates a url to the emoji from the Discord CDN. */
-export function emojiUrl(bot: Bot, id: bigint, animated = false) {
+export function getEmojiURL(_bot: Bot, id: bigint, animated = false): string {
```
Skillz4Killz, line 4: Hmm renaming this is a bit weird as there is no fetching involved here. The get makes me think there was some fetching going on here

helpers/guilds/getBan.ts
```diff
@@ -1,16 +1,22 @@
 import type { Bot } from "../../bot.ts";
+import { User } from "../../transformers/member.ts";
 import { DiscordBan } from "../../types/discord.ts";
 
+export type Ban = {
+  reason: string | undefined;
```
Skillz4Killz, line 6:
```suggestion
  reason?: string;
```

helpers/guilds/getVanityUrl.ts
```diff
@@ -1,16 +1,15 @@
...
+export type VanityUrl = {
+  code?: string;
+  uses: number;
+};
 
 /** Returns the code and uses of the vanity url for this server if it is enabled else `code` will be null. Requires the `MANAGE_GUILD` permission. */
-export async function getVanityUrl(bot: Bot, guildId: bigint) {
-  const result = await bot.rest.runMethod<Partial<DiscordInviteMetadata>>(
+export async function getVanityUrl(bot: Bot, guildId: bigint): Promise<VanityUrl> {
+  return bot.rest.runMethod<VanityUrl>(
```
Skillz4Killz, line 10:
```suggestion
  return await bot.rest.runMethod<VanityUrl>(
```

helpers/integrations/getIntegrations.ts
```diff
@@ -1,35 +1,19 @@
...
-        expire_grace_period: res.expire_grace_period,
-        user: res.user,
-        account: res.account,
-        synced_at: res.synced_at,
-        subscriber_count: res.subscriber_count,
-        revoked: res.revoked,
-        application: res.application,
-      });
+    results.map((result) => {
+      const integration = bot.transformers.integration(bot, { guild_id: guildId.toString(), ...result });
```
Skillz4Killz, line 16: Undo this change, we do not use spread operator in dd, please write them manually

helpers/interactions/commands/editInteractionResponse.ts
```diff
@@ -101,7 +103,7 @@ export async function editInteractionResponse(
   );
 
   // If the original message was edited, this will not return a message
-  if (!options.messageId) return result as undefined;
+  if (!options.messageId) return undefined;
```
Skillz4Killz, line 106:
```suggestion
  if (!options.messageId) return;
```

helpers/interactions/sendInteractionResponse.ts
```diff
@@ -36,7 +36,7 @@ export async function sendInteractionResponse(
 
   // A reply has never been send
   if (bot.cache.unrepliedInteractions.delete(id)) {
-    return await bot.rest.sendRequest<undefined>(bot.rest, {
+    return <undefined> await bot.rest.sendRequest(bot.rest, {
```
Skillz4Killz, line 39: why?

helpers/members/editBotNickname.ts
```diff
@@ -1,15 +1,19 @@
...
+): Promise<string | undefined> {
+  const result = await bot.rest.runMethod<{ nick?: string }>(
     bot.rest,
     "PATCH",
     bot.constants.routes.USER_NICK(guildId),
     options,
   );
 
-  if (!response?.nick) return;
+  if (!result?.nick) return undefined;
```
Skillz4Killz, line 16:
```suggestion
  if (!result?.nick) return;
```

helpers/webhooks/sendWebhookMessage.ts
```diff
@@ -32,7 +38,7 @@ export async function sendWebhook(bot: Bot, webhookId: bigint, webhookToken: str
       components: options.components?.map((component) => bot.transformers.reverse.component(bot, component)),
     },
   );
-  if (!options.wait) return;
+  if (!options.wait) return undefined;
```
Skillz4Killz, line 41:
```suggestion
  if (!options.wait) return;
```

helpers/channels/deleteChannel.ts
vxern: It is a personal preference. The code works just fine without it, but I feel that it makes the code more consistent since a value is always returned, even if its void.

helpers/discovery/editDiscovery.ts
vxern: I do not know. These endpoints are not documented.

helpers/discovery/getDiscovery.ts
vxern: On it.

vxern: After some digging: It is not a snowflake. Just a number. Fixed in latest commit.

helpers/interactions/sendInteractionResponse.ts
vxern: The generic annotation in `bot.rest.sendRequest()` is redundant, since the value is irrelevant, however the function requires that an `undefined` be returned if there is no value, so cast Promise return type to `undefined`.

helpers/webhooks/sendWebhookMessage.ts
vxern: Updated in latest commit.

helpers/members/editBotNickname.ts
vxern: Updated in latest commit.

helpers/interactions/commands/editInteractionResponse.ts
vxern: Updated in latest commit.

helpers/emojis/getEmojiUrl.ts
vxern: I thought that adding the get keyword to the file/function names that were missing them would make them more consistent.

helpers/discovery/editDiscovery.ts
vxern: Updated type.

helpers/channels/deleteChannel.ts
Skillz4Killz: In that case id prefer return await bot.rest.runMethod<void>()

helpers/emojis/getEmojiUrl.ts
Skillz4Killz: Consistency is good but not if it causes confusion. `get` should be reserved for prefixing on functions that make a request to discord. Otherwise its just a util function and imo something like this can in fact be removed from `helpers` and moved to utils folder

helpers/channels/deleteChannel.ts
vxern: Sure, that works too. Going to update.

helpers/interactions/sendInteractionResponse.ts
Skillz4Killz: I'm not a fan of this style. I prefer the older way its much better on the eyes and readability. the generic undefined passed to sendRequest tells typescript that it is returning undefined which is exactly what we need imo. No?

vxern: Indeed, and yes, perhaps it is indeed better for readability. I'll update the code.

helpers/guilds/getBan.ts
vxern: I believe I used `undefined` here because the TS analyser had issues with inferring the type in `getBans.ts`. Updated in next commit.

helpers/integrations/getIntegrations.ts
vxern: Done.

Skillz4Killz (approved):

PR #2415, opened 2022-09-01
add resume_gateway_url to DiscordReady

### LINKED ISSUE #2400: [api-docs] [docs] add resume_gateway_url documentation (#5282)
A new commit was made into the api-docs repo: https://github.com/discord/discord-api-docs/commit/30b37e576d3457f5401c50dea96e8bc5bfdd675b

* [docs] add resume_gateway_url documentation

* Update docs/topics/Gateway.md

Co-authored-by: Florian Spieß <business@minnced.club>

* Update docs/topics/Gateway.md

Co-authored-by: snek <me@gus.host>

* Update docs/topics/Gateway.md

Co-authored-by: snek <me@gus.host>

Co-authored-by: Florian Spieß <business@minnced.club>
Co-authored-by: snek <me@gus.host>

This is a bot created issue.

lts20050703:
closes #2400 
upstream https://github.com/discord/discord-api-docs/commit/30b37e576d3457f5401c50dea96e8bc5bfdd675b


Skillz4Killz (changes requested):
This is a major change in how gateway works, this PR needs to also touch up on the resume file to use this url, it also needs to save this url somewhere on the shard so it can use this url when it gets the READY event

gateway/manager/resharder.ts
```diff
@@ -197,6 +197,7 @@ export async function tellWorkerToPrepare(resharder: Resharder, shardId: number)
 
       if (messageData.t === "READY") {
         const payload = messageData.d as DiscordReady;
+        shard.resumeGatewayUrl = payload.resume_gateway_url;
```
Skillz4Killz, line 200: I think this should also be done in the handleMessage wherever the READY payload is being handled

gateway/shard/connect.ts
```diff
@@ -8,8 +8,7 @@ export async function connect(shard: Shard): Promise<void> {
   }
   shard.events.connecting?.(shard);
 
-  // Explicitly setting the encoding to json, since we do not support ETF.
-  const socket = new WebSocket(`${shard.gatewayConfig.url}/?v=${shard.gatewayConfig.version}&encoding=json`);
+  const socket = new WebSocket(shard.resumeGatewayUrl);
```
Skillz4Killz, line 11: This should be like a ternary that checks if shard.state is Resuming and only then it uses this resume gateway url otherwise it should use the old one, also either way we need to keep the v=${version}&encoding=json

gateway/shard/createShard.ts
```diff
@@ -91,6 +90,10 @@ export function createShard(
     state: ShardState.Offline,
     /** The total amount of shards which are used to communicate with Discord. */
     totalShards: options.totalShards,
+    // Explicitly setting the encoding to json, since we do not support ETF.
+    resumeGatewayUrl: `${options.gatewayConfig.url ?? "wss://gateway.discord.gg"}/?v=${
```
Skillz4Killz, line 94: by default this should be ""

Skillz4Killz: the config url is something else not for resume

itohatweb (changes requested):
1. Please note that the gateway should support custom URLs, means if someone decides to not use Discords default gateway but something else, that URL should be used and not the custom URL from Discord. This means that we should not use the resume URL if the connection URL has been changed to a non Discord one.

2. Make `resumeGatewayUrl` optional and use a check (eg. `resumeGatewayUrl ?? default url` note you need to take the 1st point into account too).

3. You have only implemented the resume URL handling for resharding and not the actual shard.

gateway/manager/resharder.ts
lts20050703: sorry what
there's only one place in the handleMessage where the READY payload is being handled, which is right here

lts20050703: OH YOU MEAN THE gateway/shard/handleMessage.ts file

gateway/shard/connect.ts
```diff
@@ -8,8 +8,11 @@ export async function connect(shard: Shard): Promise<void> {
   }
   shard.events.connecting?.(shard);
 
-  // Explicitly setting the encoding to json, since we do not support ETF.
-  const socket = new WebSocket(`${shard.gatewayConfig.url}/?v=${shard.gatewayConfig.version}&encoding=json`);
+  const socket = new WebSocket(
```
Skillz4Killz, line 11:
```ts
  const socket = new WebSocket(`${shard.gatewayConfig.url === "wss:gateway.discord.gg" ? shard.state === ShardState.Resuming ? shard.resumeGatewayUrl : shard.gatewayConfig.url}/?v=${shard.gatewayConfig.version}&encoding=json`);
```

gateway/shard/createShard.ts
```diff
@@ -91,6 +90,7 @@ export function createShard(
     state: ShardState.Offline,
     /** The total amount of shards which are used to communicate with Discord. */
     totalShards: options.totalShards,
+    resumeGatewayUrl: undefined as undefined | string,
```
Skillz4Killz, line 93:
```suggestion
    resumeGatewayUrl: "",
```

gateway/shard/connect.ts
```diff
@@ -8,8 +8,16 @@ export async function connect(shard: Shard): Promise<void> {
...
+    `${
+      shard.gatewayConfig.url === "wss://gateway.discord.gg"
```
Skillz4Killz, line 13: @itohatweb do u mind double checking this real quick, everything else in this pr is good just double check this line imo

gateway/manager/resharder.ts
itohatweb, line 200: This is not necessary.
```suggestion
```

gateway/shard/createShard.ts
```diff
@@ -91,6 +90,7 @@ export function createShard(
     state: ShardState.Offline,
     /** The total amount of shards which are used to communicate with Discord. */
     totalShards: options.totalShards,
+    resumeGatewayUrl: "",
```
itohatweb, line 93:
```suggestion
    resumeGatewayUrl: undefined as string | undefined,
```

gateway/shard/connect.ts
itohatweb: That one looks good.

itohatweb (changes requested):

gateway/manager/resharder.ts
Skillz4Killz: I think its necessary for the way resharding works, a new connection can have a different resume url so we need to listen to the ready events and not ignore them

gateway/shard/createShard.ts
lts20050703: @itohatweb skillz made me change that from `undefined as string | undefined` to `""`
@Skillz4Killz why did you mark this as resolved

itohatweb: why

PR #2438, opened 2022-09-08
Fixes getReations by removing the double URI encode

Burn-E99:
[util/routes.ts](https://github.com/discordeno/discordeno/blob/aca0e3cf1b74db1b39ff30897ae18ab198aab18b/util/routes.ts#L74) already takes care of URI encoding the emoji in the request.  This PR removes the extra URI encode to let the getReactions function work.

This fix was discussed on discord [here](https://discord.com/channels/785384884197392384/1016114488958791751/1016114491953532968).


lts20050703 (approved):
LGTM!

Issue #2405, opened 2022-08-31
[api-docs] Update explanation of command character limit, including localization (#5343)


itohatweb:
A new commit was made into the api-docs repo: https://github.com/discord/discord-api-docs/commit/5582b28c5c7c39055e09e4559bcbe9dfc8dd12ca
No details given.

This is a bot created issue.

PR #2441, opened 2022-09-08
plugins/permissions: slash commands max 4000 chars

lts20050703:
Closes: #2405
Upstream: https://github.com/discord/discord-api-docs/commit/5582b28c5c7c39055e09e4559bcbe9dfc8dd12ca
TL;DR: Slash commands can have a maximum of 4000 characters for combined name, description, and value properties for each command, its options (including subcommands and groups), and choices. When localization fields are present, only the longest localization for each field (including the default value) is counted towards the size limit.

THE CODE IS CURSED PLEASE HELP ME [SEMANTIC COMPRESSION](https://caseymuratori.com/blog_0015) IT


lts20050703:
Some context / elaboration:
`options.nameLocalizations ? Math.max(options.name, ...Object.values(options.nameLocalizations)) : options.name`
image

Issue #2457, opened 2022-09-11
[api-docs] Riot Games connection (#5306)


itohatweb:
A new commit was made into the api-docs repo: https://github.com/discord/discord-api-docs/commit/ed87ef81687747da13f26da68b618a326aa039bf

* Riot Games connection

* fix order

Co-authored-by: Vitor <vito7.dev@gmail.com>

Co-authored-by: Vitor <vito7.dev@gmail.com>

This is a bot created issue.

Issue #2456, opened 2022-09-11
[api-docs] Add PayPal connection (#5355)


itohatweb:
A new commit was made into the api-docs repo: https://github.com/discord/discord-api-docs/commit/922f789e861d0f1e5d682b5c04950856001c1bf1

Co-authored-by: Ian Webster <ianw_github@ianww.com>

This is a bot created issue.

PR #2458, opened 2022-09-11
add union type DiscordConnectionServices

lts20050703:
Closes: #2457 
Upstream: https://github.com/discord/discord-api-docs/commit/ed87ef81687747da13f26da68b618a326aa039bf
Closes: #2456 
Upstream: https://github.com/discord/discord-api-docs/commit/922f789e861d0f1e5d682b5c04950856001c1bf1

### DIFF
--- a/types/discord.ts
+++ b/types/discord.ts
@@ -102,24 +102,23 @@ export interface DiscordConnection {
 }
 
 /** https://discord.com/developers/docs/resources/user#connection-object-services */
-export enum DiscordConnectionServices {
-  "Battle.net" = "battlenet",
-  "eBay" = "ebay",
-  "Epic Games" = "epicgames",
-  "Facebook" = "facebook",
-  "GitHub" = "github",
-  "League of Legends" = "leagueoflegends",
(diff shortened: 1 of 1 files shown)


types/discord.ts
```diff
@@ -85,22 +85,42 @@ export interface DiscordConnection {
...
+  show_activity: boolean;
   /** Visibility of this connection */
   visibility: VisibilityTypes;
 
   /** An array of partial server integrations */
   integrations?: DiscordIntegration[];
 }
 
+/** https://discord.com/developers/docs/resources/user#connection-object-services */
+export enum DiscordConnectionServices {
```
Skillz4Killz, line 105: Why is this an enum? Generally, an enum should be used for numbers imo, since numbers are harder to read and know it's meaning. In this case it is a string `ebay` is clear where as type `0` for channel is not clear that it is a text channel.

Skillz4Killz: This should be a `export type DiscordConnectionServices = "ebay" | "x" | "y"......;`

Skillz4Killz (approved):

Issue #2455, opened 2022-09-11
[api-docs] Add back nonce (#5078)


itohatweb:
A new commit was made into the api-docs repo: https://github.com/discord/discord-api-docs/commit/62fb60ced0f514414004d4a73b00b1aaa869b20b

* Add back nonce

* Update docs/resources/Channel.md

Co-authored-by: Austin Keener <keeneraustin@yahoo.com>

This is a bot created issue.

PR #2459, opened 2022-09-11
add nonce to sendMessage options

lts20050703:
Closes: #2455 
Upstream: https://github.com/discord/discord-api-docs/commit/62fb60ced0f514414004d4a73b00b1aaa869b20b

### DIFF
--- a/helpers/messages/sendMessage.ts
+++ b/helpers/messages/sendMessage.ts
@@ -39,6 +39,7 @@ export async function sendMessage(bot: Bot, channelId: bigint, options: CreateMe
     bot.constants.routes.CHANNEL_MESSAGES(channelId),
     {
       content: options.content,
+      nonce: options.nonce,
       tts: options.tts,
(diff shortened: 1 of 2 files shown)


helpers/messages/sendMessage.ts
```diff
@@ -125,6 +126,8 @@ export async function sendMessage(bot: Bot, channelId: bigint, options: CreateMe
 export interface CreateMessage {
   /** The message contents (up to 2000 characters) */
   content?: string;
+  /** Can be used to verify a message was sent (up to 25 characters). Value will appear in the Message Create event. */
```
Skillz4Killz, line 129: Does this exist in the DiscordMessage interface and the message transformer?

Skillz4Killz (dismissed):

helpers/messages/sendMessage.ts
lts20050703: DiscordMessage interface already has nonce.
Added nonce to (Discordeno)Message (return type of message transformer)

Skillz4Killz (approved):

Issue #2434, opened 2022-09-07
Split the `getWebhooks()` helper into `getWebhooksChannel()` and `getWebhooksGuild()`.


vxern:

vxern:
Update: `getChannelWebhooks()` is found in `/helpers/channels`. This file should be moved to `/helpers/guilds/webhooks` and renamed to `getWebhooksChannel`, while the `getWebhooks` file in `/helpers/guilds/webhooks` should be renamed to `getWebhooksGuild`.

lts20050703:
https://discord.com/developers/docs/resources/webhook#get-channel-webhooks

PR #2465, opened 2022-09-14
BREAKING: Rename getWebhooks to getGuildWebhooks

lts20050703:
Rename getWebhooks to getGuildWebhooks
Move getChannelWebhooks from `helpers/channels` to `helpers/webhooks`
Closes: #2434


itohatweb (changes requested):
Hey, please resolve the conflicts.
image

Skillz4Killz:
@lts20050703

lts20050703:
@itohatweb Merge conflicts resolved but @Skillz4Killz doesn't like this PR
image

itohatweb (commented):
Only reason I could see for the name change would be that `getWebhooks` could cause confusion since users might want to get the webhooks out of a channel with that function.
Although TS would give a small hint that there is also `getChannelWebhooks`.

As of the file moving ig it would be fit better if those functions are in the `helpers/webhooks` directory.

Skillz4Killz:
@lts20050703 needs fixing tests

Issue #2479, opened 2022-09-16
[api-docs] Document connection two_way_link (#5259)


itohatweb:
A new commit was made into the api-docs repo: https://github.com/discord/discord-api-docs/commit/dc4e41baa10a58cfabf16c39236619e5cd0252a9

* Document connection two_way_link

* Fix description

Co-authored-by: Will Nelson <will.nelson@discordapp.com>

This is a bot created issue.

PR #2481, opened 2022-09-17
feat(types)!: add `two_way_link` to `DiscordConnection`

itohatweb:
When the connection has a corresponding thrid party OAuth2 token `two_way_link` will be set to `true`.

Reference: https://github.com/discordeno/discordeno/issues/2479

Closes: #2479

### DIFF
--- a/types/discord.ts
+++ b/types/discord.ts
@@ -99,6 +99,8 @@ export interface DiscordConnection {
 
   /** An array of partial server integrations */
   integrations?: DiscordIntegration[];
+  /** Whether this connection has a corresponding third party OAuth2 token. */
+  two_way_link: boolean;
 }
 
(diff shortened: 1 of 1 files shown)


types/discord.ts
```diff
@@ -99,6 +99,8 @@ export interface DiscordConnection {
 
   /** An array of partial server integrations */
   integrations?: DiscordIntegration[];
+  /** Whether this connection has a corresponding third party OAuth2 token. */
+  two_way_link: boolean;
```
Skillz4Killz, line 103: Does this need a change in a transformer?

Skillz4Killz (approved):

types/discord.ts
itohatweb: No, we currently do not have a transformer for that type.

PR #2485, opened 2022-09-22
tests: rate limit test

lts20050703:


tests/misc/rateLimit.test.ts
```diff
@@ -0,0 +1,9 @@
+import { loadBot } from "../mod.ts";
+import { CACHED_COMMUNITY_GUILD_ID } from "../utils.ts";
+
+Deno.test("[Misc] Rate Limit Test", async () => {
+  const bot = loadBot();
+  const channel = await bot.helpers.createChannel(CACHED_COMMUNITY_GUILD_ID);
+  for (let i = 0; i < 15; i += 1) await bot.helpers.sendMessage(channel.id, { content: "Rate Limit Test" });
```
Skillz4Killz, line 7: I don't think the await is good here, i think this needs to be a promise.all like 
```ts
await Promise.all(Array(10).map(() => bot.helpers.sendMessage(channel.id, { content: "Rate Limit Test" })))
```

PR #2493, opened 2022-09-30
fix(transformers): transform disabled field for InputText

Gary50613:
the disabled field didn't get transformed to DiscordComponent


Gary50613:
sorry for typo

Issue #2298, opened 2022-06-15
Fix benchmarks


Skillz4Killz:
benchmark tests have been broken since last gateway rewrite. Needs to be regenerated.

lts20050703:
is it still broken?

lts20050703:
@Skillz4Killz @itohatweb is benchmark tests still broken?

Skillz4Killz:
Yes

PR #2546, opened 2022-10-23
CI: fix/run benchmark

H01001000:
My idea is based on the ~~correctness and update speed of the [discord/discord-api-docs](https://github.com/discord/discord-api-docs) repo, using its example data as the date we run the benchmark on~~ (benchmark repo)[https://github.com/H01001000/benchmarks] (my fork). Currently, it gets the guild every time it runs, but instead, we can cache it in the benchmark and create PR or commit it directly when there is a difference with the last cache, which can run on schedule with GitHub action. This should be relatively easy to do/maintain and can run on every commit.

Then we can use  [github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark) to check the result/comment on the result and put it on site. 

Any feedback is appreciated


itohatweb (changes requested):
The problem with that is, that Discord doc examples are pretty unreliable most of the time. 
What I mean is that it's not given that the examples get updated accordingly on a change, which would mean that our gh action would break because of that.

Skillz4Killz:
No, using stale and manually maintained things is not viable. The performance folder must run on request to the live api! I can discuss the goals of this in depth in discord @ me there

Also to clarify if this is about #2298 then this pr does not solve that.

PR #2555, opened 2022-10-26
Test: clean up CI and test

H01001000:
No Breaking Change, every test will work fine

This PR adds ignore to test if there env var TEST_ENV set to UNIT or INTEGRATION, which will skip the integration test if TEST_ENV == UNIT and visa versa, test work as before is TEST_ENV is not set.
Also, I combine tests into one test workflow, with different job type tests, format tests, integration tests, and unit tests
With this, the integration test added concurrency and the unit test added the parallel flag

The test matrix is removed for now, it could add back later

Tho why codecov disable on PR?


itohatweb:
Note that there was actually a reason for why tests are only run for pushes to specific folders

H01001000:
@itohatweb gonna do that in next pr

H01001000:
@itohatweb what path the unit/integration test should not run on?

PR #2559, opened 2022-10-27
performance bench: add bench cache 

H01001000:
This PR is the foundation of #2546, which add a script that will cache a guild's object and run on a schedule (per day for now)

But need to discuss and modify the actual guild and channel to fetch

This also remove all other file


itohatweb:
Closing since this would delete the repo.

H01001000:
oh wait wrong branch

PR #2564, opened 2022-10-28
Add cache with proxy to main repo as a plugin.

deepsarda:
Uses the codebase at:

https://github.com/discordeno/cache-proxy/pull/11

Certain extras files are removed such as gitignore, package.json etc


itohatweb:
We do not want this proxy to be in the main repository

PR #2569, opened 2022-11-01
fix: add members collection to guild in guild transformer if available

AwesomeStickz:


transformers/guild.ts
```diff
@@ -65,6 +65,12 @@ export function transformGuild(bot: Bot, payload: { guild: DiscordGuild } & { sh
         return [result.id, result];
       }),
     ),
+    members: new Collection(
+      payload.guild.members?.map((member) => {
```
Skillz4Killz, line 69: Can you check if voice states come with guild members and then also run the transformer for them

AwesomeStickz: Wouldn't that be member transformer's job?

PR #2593, opened 2022-11-19
Fix sending files through a rest proxy

danny-may:
When the rest manager detects that it has a proxy to call, it [converts all file blobs to data uris](https://github.com/discordeno/discordeno/blob/main/rest/runMethod.ts#L35-L47). When the proxy receives this request, the data uri isnt converted back to a blob, so an error is thrown by undici in a node environment.

This PR adds a conversion step before any files are added to a request to ensure their blob property is a blob instance.

The code for converting from a string to arraybuffer was taken from the same gist as the already existing code to convert an arraybuffer to a string, I just moved them to their own file (/util/base64.ts)


lts20050703:
@danny-may Please run `deno fmt`

danny-may:
I missed that, fixed!

Skillz4Killz (approved):

PR #2682, opened 2022-12-10
deleted validators

Larsundso:


Larsundso:
@Skillz4Killz hope this is what you wanted

PR #2703, opened 2022-12-27
Clean idea for migration

Skillz4Killz:


packages/rest/package.json
```diff
@@ -24,9 +24,7 @@
     "test:test-type": "tsc --project tsconfig.test.json"
   },
   "dependencies": {
-    "@discordeno/constant": "18.0.0-alpha.1",
     "@discordeno/transformer": "18.0.0-alpha.1",
-    "@discordeno/utils": "18.0.0-alpha.1",
     "dotenv": "^16.0.3"
   },
   "devDependencies": {
```
H01001000, line 30: Missing types as dev dependency

packages/rest/src/manager.ts
```diff
@@ -0,0 +1,136 @@
+/* eslint-disable @typescript-eslint/restrict-template-expressions */
+import TRANSFORMERS from '@discordeno/transformer'
+import type { BigString, Camelize, DiscordUser } from '@discordeno/types'
+
+// TODO: make dynamic based on package.json file
```
H01001000, line 5: Import package json as json

packages/rest/tests/utils.ts
```diff
@@ -0,0 +1,12 @@
+import type { Camelize, DiscordGuild } from '@discordeno/types'
+import { createRestManager } from '../src/manager.js'
+import { token } from './constants.js'
```
H01001000, line 3: This file should not import token, since this will be used for unit test only

Issue #2730, opened 2023-01-18
Application command missing nsfw prop


Yaikava:
**Describe the bug** A clear and concise description of what the bug is.
Both current version and node migration are missing this property. 
https://discord.com/developers/docs/interactions/application-commands#application-command-object-application-command-structure

**Expected behavior** A clear and concise description of what you expected to happen.

**Screenshots** If applicable, add screenshots to help explain your problem.

**Version details (please complete the following information):**

- Discordeno version: main branch or node-migration-clean
- Deno version: irrelevant

PR #2731, opened 2023-01-18
🐛 - fix: types - slashcommands - add nsfw prop

Yaikava:
I do not think implementation needs to be changed, but did not test it since I am not really sure how this turborepo and node stuff works.


Yaikava:
Fixes discordeno/discordeno#2730

Skillz4Killz (approved):

PR #2738, opened 2023-01-22
test/ci: add bun test

### LINKED ISSUE #2386: Bun Compatibility
**Is your feature request related to a problem? Please describe.** A clear and concise description of what the problem
is. Ex. I'm always frustrated when [...]

Work with bun

**Describe the solution you'd like** A clear and concise description of what you want to happen.

Work with bun

**Describe alternatives you've considered** A clear and concise description of any alternative solutions or features
you've considered.

Work with bun

**Additional context** Add any other context or screenshots about the feature request here.

Work with bun

H01001000:
close #2386


scripts/fixBunTestExtension.js
```diff
@@ -0,0 +1,21 @@
...
+const dirs = ['']
+try {
+  fs.rmSync('bunTestsDist', { recursive: true, force: true })
+} catch {
+  //
+}
+for await (const dir of dirs) {
+  await Promise.all(
+    fs.readdirSync(`tests${dir}`).map(async (file) => {
+      if (!file.endsWith('.ts') && !file.includes('.')) {
```
Jarred-Sumner, line 12: jeez i'm sorry you had to write this code

H01001000: It's fine, tho I'm not sure why but it seems "describe" and variable define in beforeEach hook don't seem to work consistently. What do I think is, it should be a drop-in replacement for mocha and jest or do I need to change any code?

P.S. it seems docs/guide for testing is missing

Jarred-Sumner: bun:test is intended as a jest replacement but we haven't considered mocha as much. There are some important things still missing (such as snapshots and mocks)

This is the current docs for tests https://bun.sh/docs/cli/test, but we need to add more and explain things like `expect`, `describe`, `beforeAll` etc

> seems "describe" and variable define in beforeEach hook don't seem to work consistently

Bun will automatically import `describe` and `beforeEach` and `expect`, but it's scoped to the file (unique per file, not global). If this turns out that it causes compatibilty issues, we can make it a global in the next release

H01001000: mainly it seems to have a problem with the following (just an example), it seems the object does not exist
```ts
describe('this is a test', () => {
  let object

  beforeEach(() => {
    object = {}
  })

  it('will test', () => {
    expect(object).exist()
  })
})
```

H01001000: we use this heavily for creating new object for test on each test

Jarred-Sumner: .exist isn't a Jest matcher

Instead .toBeTruthy() or .not.toBeUndefined()

H01001000: just an example, we used chai and did not use bun:test's expect

PR #2778, opened 2023-02-12
Add and update select menu types

bibeoaba:


types/discord.ts
```diff
@@ -1173,7 +1173,15 @@ export interface DiscordActionRow {
   type: 1;
   /** The components in this row */
   components:
-    | [DiscordSelectMenuComponent | DiscordButtonComponent | DiscordInputTextComponent]
+    | [
```
Skillz4Killz, line 1176: I don't understand the purpose of this change, do u mind explaining what the difference is between

DiscordSelectMenuChannelsComponent
DiscordSelectMenuUsersComponent
....

bibeoaba: Sorry for the late reply!

There are some differences among them: `options` is only available for `DiscordSelectMenuComponent` and `channel_types` can only be used for  `DiscordSelectMenuChannelComponent`. The others don’t have both of them.

The Discord API docs also mention that: https://discord.com/developers/docs/interactions/message-components#select-menu-object-select-menu-structure

PR #2907, opened 2023-03-29
fix(utils): leaky bucket slow refill test

itohatweb:


H01001000 (approved):

packages/utils/tests/bucket.spec.ts
```diff
@@ -143,13 +151,21 @@ describe('bucket.ts', () => {
...
+        expect(await promiseState(acquired)).to.equal('fulfilled')
+      }
+
+      await clock.tickAsync(499)
+
+      {
+        const acquired = bucket.acquire()
+        expect(await promiseState(acquired)).to.equal('pending')
+      }
+
```
H01001000, line 166:
```suggestion
      const acquired1 = bucket.acquire()
      const acquired2 = bucket.acquire()

      // js event loop
      await (async () => {})()

      expect(await promiseState(acquired1)).to.equal('fulfilled')
      expect(await promiseState(acquired2)).to.equal('pending')

      await clock.tickAsync(499)
      expect(await promiseState(acquired2)).to.equal('pending')

      await clock.tickAsync(1)
      expect(await promiseState(acquired2)).to.equal('fulfilled')

```

H01001000 (changes requested):

PR #2911, opened 2023-03-30
ci: move tests into one single workflow file

itohatweb:


H01001000:
The reasons i split are having separate status badge for each package and it’s easier to see what package goes wrong at pr check

itohatweb:
Do we really need a status badge for every separate package?
Like ideally we should never have it failing tbh.

Also the reason for me moving it to one single file is that I find the separate stuff less friendly to see.

imo that's far better:
Screenshot_20230330_170231

H01001000:
Yea Ik, also should we only trigger push to main?

H01001000 (approved):

H01001000 (dismissed):
Also remove the badge from readme

itohatweb:
> Yea Ik, also should we only trigger push to main?

I would also run the tests on PRS.

itohatweb:
Yea sounds valid. Have added that.

Skillz4Killz (dismissed):

Skillz4Killz (approved):

H01001000 (approved):

PR #2916, opened 2023-03-31
fix(transformers): message transformer using different style

Skillz4Killz:
Changes From This Style:

- `1:1 shape`
  - `message.author` now exists
  - Other things similar to this are now available. 
- `desiredProps`
  - This now is set by default so that no props are cached by default. Users need to opt-into each prop they want. This makes it similar to intents which will help prevent devs from using unneeded memory.
  - This also allows a much simpler solution to remove properties as opposed to have to override an the entire transformer just to remove `topic` from channel.
- `Customizers`
  - This is the a handler for each transformer which users can override to easily add new properties or customize something without major overriding. For example, say a user wants to add a new property on `message` like `args`. A customizer can be written:
```ts
bot.transformers.customizers.message = (_, payload, message) {
  message.args = "";
  return message;
}
```
- `getters & setters`
  - To make it use `1:1` form of discord message, things like `tts` and `pinned` would be made booleans so much more expensive in memory. This way we can keep them as bitfields and just make getters to access it so the API is the same and `message.tts` works.
  - Optimize storage even more by things like `message.timestamp` can be a getter that converts id to timestamp instead of storing a timestamp as well as id on each message.
  - Because we are back to use Object.create(), the getters and setters will only be in memory once no matter how many messages we create. Also this allows us should we desire to add methods as well for example `message.reply()` or `message.delete()` without increasing memory

### DIFF
--- a/packages/benchmark/src/benchmarks/transformers.ts
+++ b/packages/benchmark/src/benchmarks/transformers.ts
@@ -32,11 +32,8 @@ bot.transformers.desiredProperties.message.application = true
 bot.transformers.desiredProperties.message.applicationId = true
 bot.transformers.desiredProperties.message.attachments = true
 bot.transformers.desiredProperties.message.author = true
-// @ts-expect-error idk
 bot.transformers.desiredProperties.message.channelId = true
-// @ts-expect-error idk
 bot.transformers.desiredProperties.message.components = true
-// @ts-expect-error idk
 bot.transformers.desiredProperties.message.content = true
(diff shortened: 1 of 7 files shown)


itohatweb:
What's the motivation on having that on bot?

itohatweb:
TBH I'm not a fan of having this here.

meister03 (approved):
Mentioned my opinion in the DD Server
--> when someone is willing to maintain it, why not

AwesomeStickz (approved):

PR #2954, opened 2023-04-02
feat(bot): allow passing of `applicationId`

itohatweb:


packages/bot/src/bot.ts
```diff
@@ -85,6 +85,8 @@ export function createBot(options: CreateBotOptions): Bot {
 export interface CreateBotOptions {
   /** The bot's token. */
   token: string
+  /** Application Id of the bot incase it is an old bot token */
```
itohatweb, line 88:
```suggestion
  /** Application Id of the bot incase it is an old bot token. */
```

PR #2962, opened 2023-04-03
refactor(rest): add header constants

itohatweb:


packages/rest/src/manager.ts
```diff
@@ -54,6 +54,17 @@ import { createRoutes } from './routes.js'
...
+const DISCORD_API_VERSION = 10
+const DISCORD_API_URL = 'https://discord.com/api'
+
+const AUDIT_LOG_REASON_HEADER = 'x-audit-log-reason'
+const RATE_LIMIT_REMAINING_HEADER = 'x-ratelimit-remaining'
+const RATE_LIMIT_RESET_AFTER_HEADER = 'x-ratelimit-reset-after'
+const RATE_LIMIT_GLOBAL_HEADER = 'x-ratelimit-global'
+const RATE_LIMIT_BUCKET_HEADER = 'x-ratelimit-bucket'
+const RATE_LIMIT_LIMIT_HEADER = 'x-ratelimit-limit'
+const RATE_LIMIT_SCOPE_HEADER = 'x-ratelimit_scope'
```
itohatweb, line 66:
```suggestion
const RATE_LIMIT_SCOPE_HEADER = 'x-ratelimit-scope'
```

packages/rest/src/manager.ts
```diff
@@ -54,6 +54,17 @@ import { createRoutes } from './routes.js'
 // TODO: make dynamic based on package.json file
 const version = '19.0.0-alpha.1'
 
+const DISCORD_API_VERSION = 10
```
Skillz4Killz, line 57: export all of these

Skillz4Killz (changes requested):

Skillz4Killz (approved):

PR #2968, opened 2023-04-04
fix: finish bucket file coverage

Skillz4Killz:


packages/utils/tests/bucket.spec.ts
```diff
@@ -169,5 +169,28 @@ describe('bucket.ts', () => {
...
+    it("Don't process queue twice", () => {
+      const bucket = new LeakyBucket({
+        max: 1,
+        refillInterval: 500,
+        refillAmount: 1,
+      })
+      // fake processing
+      bucket.processing = true;
+      // request when already processing
+      bucket.processQueue();
```
H01001000, line 193: missing a check here

packages/utils/tests/bucket.spec.ts
```diff
@@ -169,5 +169,28 @@ describe('bucket.ts', () => {
...
+    it("remaining is 0 when used too many", () => {
+      const bucket = new LeakyBucket({
+        max: 1,
+        refillInterval: 500,
+        refillAmount: 1,
+      })
+      // max is < used
+      bucket.used = 2;
+      expect(bucket.remaining).equals(0);
+    })
```
H01001000, line 182:
```suggestion
  describe('remaining', () => {
    it("should be 0 even used too many", () => {
      const bucket = new LeakyBucket({
        max: 1,
        refillInterval: 500,
        refillAmount: 1,
      })
      // max is < used
      bucket.used = 2;
      expect(bucket.remaining).equals(0);
    })
  })
```

H01001000 (changes requested):

packages/utils/tests/bucket.spec.ts
Skillz4Killz: what check?

H01001000: how do you check if it processes twice?

Skillz4Killz: it doesn't process twice, the second request when already processing is true should just cancel out, no way to check that

PR #2971, opened 2023-04-05
fix(site): remove default image

### LINKED ISSUE #2963: Remove Docusaurus Embed default image
Eg. when sharing this Link on Discord: https://discordeno.js.org/docs/bigbot/step-4-bot#server-splitting
A thumbnail gets showed which is not about Discordeno but Docusaurus. This needs to be changed

H01001000:
close #2963


Skillz4Killz:
Would be nice to have some image there, maybe 
image

PR #3007, opened 2023-04-18
fix(types): MessageActivityTypes.JoinRequest

AwesomeStickz:
PS: Should we do like this or leave it as it is to prevent future changes unintentionally breaking it?
image


itohatweb (changes requested):
Please use explicit values for all changed enums.

AwesomeStickz:
> Please use explicit values for all changed enums.

Done 👍

Skillz4Killz (approved):

PR #3025, opened 2023-05-07
Make sure we actually try to get nonce from options first

TotoTheDragon:


itohatweb:
Need to think about a better fix for this issue.

PR #3029, opened 2023-05-09
fix(gateway): Deno WebSocket

Endy3032:
Use Deno's ws via eval

The old method imports `fixGateway` from url in `bot` package to overwrite gateway which doesnt work (at least from my testing)


itohatweb:
Thanks for the PR but we do not use eval in the libs code.

PR #3030, opened 2023-05-09
fix(gateway): Use Deno's WebSocket

Endy3032:
Check for Deno namespace and use its ws instead of node ws (without using eval like #3029)

### DIFF
--- a/packages/gateway/src/Shard.ts
+++ b/packages/gateway/src/Shard.ts
@@ -17,8 +17,7 @@ import type { RequestMemberRequest } from './manager.js'
 import type { BotStatusUpdate, ShardEvents, ShardGatewayConfig, ShardHeart, ShardSocketRequest, StatusUpdate, UpdateVoiceState } from './types.js'
 import { ShardSocketCloseCodes, ShardState } from './types.js'
 
-declare var Deno: any
-declare var WebSocket: any
+declare let WebSocket: any
 
 export class DiscordenoShard {
   /** The id of the shard */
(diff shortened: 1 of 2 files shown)


packages/gateway/src/Shard.ts
```diff
@@ -129,13 +133,19 @@ export class DiscordenoShard {
     url.searchParams.set('v', this.gatewayConfig.version.toString())
     url.searchParams.set('encoding', 'json')
 
-    const socket = new WebSocket(url.toString())
+    let socket: NodeWebSocket
+
+    // @ts-expect-error Deno
+    if (isDeno) socket = new WebSocket(url.toString())
+    else if (isNode) socket = new NodeWebSocket(url.toString())
+    else throw new Error('Unsupported runtime')
```
Skillz4Killz, line 141: Should use node still. We do support other runtimes like bun which support ws. Its just deno that doesnt support ws

Skillz4Killz: Just default to node if not deno like socket = isDeno ? Websocket : Nodewebsocket

Skillz4Killz (approved):

packages/gateway/src/Shard.ts
Endy3032: This was what ITOH suggested to me in DM since he said bun is not fully compatible yet

Skillz4Killz: im pretty sure bun works with ws but if it doesnt we need to find a lib for it. @Tricked-dev any thoughts?

Skillz4Killz (approved):

PR #3033, opened 2023-05-14
fix camelToSnakeCase and snakeToCamelCase function

Anonymous6214:
- fix snakeToCamelCase function should return lowercase even with there is no _ inside the string. 
  Ex: snakeToCamelCase("READY")   soulld return "ready" not "READY"

- fix snakeToCamelCase function should convert to lowercase when there's uppercase alphabet inside the string, and the first letter of the string should not be uppercase
  Ex: snakeToCamelCase("HELLO_WORLD")   sould return "helloWorld" not "HELLOWORLD"

- fix camelToSnakeCase function should not return "_" while it is the first letter
  Ex: camelToSnakeCase("HelloWorld")   should return "hello_world" not "_hello_world"

### DIFF
--- a/packages/utils/src/casing.ts
+++ b/packages/utils/src/casing.ts
@@ -33,33 +33,30 @@ export function snakelize<T>(object: T): Snakelize<T> {
 }
 
 export function snakeToCamelCase(str: string): string {
-  if (!str.includes('_')) return str
-
-  let result = ''
-  for (let i = 0, len = str.length; i < len; ++i) {
-    if (str[i] === '_') {
-      result += str[++i].toUpperCase()
-
-      continue
+  if (!str.includes('_')) return str.toLowerCase()
+    let result = ''
+    for (let i=0; i<str.length; ++i){
+        if (str[i] === '_') {
+            result += str[++i].toUpperCase()
+            continue
+        }
+        result += str[i].toLowerCase()
     }
-
-    result += str[i]
-  }
-
-  return result
+    return result
 }
 
 export function camelToSnakeCase(str: string): string {
   let result = ''
-  for (let i = 0, len = str.length; i < len; ++i) {
-    if (str[i] >= 'A' && str[i] <= 'Z') {
-      result += `_${str[i].toLowerCase()}`
-
-      continue
+    for (let i = 0; i < str.length; ++i){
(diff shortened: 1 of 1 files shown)


packages/utils/src/casing.ts
```diff
@@ -33,33 +33,30 @@ export function snakelize<T>(object: T): Snakelize<T> {
...
-
-  let result = ''
-  for (let i = 0, len = str.length; i < len; ++i) {
-    if (str[i] === '_') {
-      result += str[++i].toUpperCase()
-
-      continue
+  if (!str.includes('_')) return str.toLowerCase()
+    let result = ''
+    for (let i=0; i<str.length; ++i){
```
Skillz4Killz, line 38: have u checked the performance impact on this, the reason it was written like the other way was cause it was faster

packages/utils/src/casing.ts
```diff
@@ -33,33 +33,30 @@ export function snakelize<T>(object: T): Snakelize<T> {
...
 }
 
 export function camelToSnakeCase(str: string): string {
   let result = ''
-  for (let i = 0, len = str.length; i < len; ++i) {
-    if (str[i] >= 'A' && str[i] <= 'Z') {
-      result += `_${str[i].toLowerCase()}`
-
-      continue
+    for (let i = 0; i < str.length; ++i){
```
Skillz4Killz, line 50: same as above comment

packages/utils/src/casing.ts
```diff
@@ -33,33 +33,30 @@ export function snakelize<T>(object: T): Snakelize<T> {
...
+        if (str[i] >= 'A' && str[i] <= 'Z') {
+            if (i === 0) {
```
Skillz4Killz, line 52: remove this if

packages/utils/src/casing.ts
```diff
@@ -33,33 +33,30 @@ export function snakelize<T>(object: T): Snakelize<T> {
...
+                result += str[i].toLowerCase()
+                continue
+            }
+            result += `_${str[i].toLowerCase()}`
```
Skillz4Killz, line 56:
```suggestion
            result += `${i === 0 ? '' : '_'}${str[i].toLowerCase()}`
```

itohatweb:
Those functions are not intended for that usage. Please add new converter functions in that case.

PR #3035, opened 2023-05-17
feat: add support for new username system

AwesomeStickz:
This is a draft until [docs](https://github.com/discord/discord-api-docs/pull/6130) are finalized.

### Changes:
- Added `globalName` to `transformers.desiredProperties.user`
- Added `globalName` to `User` object in `transformUser`
- Calculate as `(user_id % 22) % 5` in `avatarUrl` of `utils` package for default avatars

### Note:
In `avatarUrl()`, I had to use `BigInt(X)` because using `Xn` directly gives the error: `BigInt literals are not available when targeting lower than ES2020. ts(2737)`, we might need to look into this, as `Xn` is probably better compared to `BigInt(X)` performance wise

### DIFF
--- a/packages/bot/src/transformers.ts
+++ b/packages/bot/src/transformers.ts
@@ -334,6 +334,7 @@ export interface Transformers {
     }
     user: {
       username: boolean
+      globalName: boolean
       locale: boolean
       flags: boolean
       premiumType: boolean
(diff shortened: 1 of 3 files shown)


AwesomeStickz:
Docs is now changed to [this PR](https://github.com/discord/discord-api-docs/pull/6218) instead, further changes includes:

- Calculate as `(user_id % 22) % 5` in `avatarUrl` of utils package for default avatars for users that migrated to new username system, and `discriminator % 5` for the ones that haven't

PR #3037, opened 2023-05-21
feat(cli): add cli

H01001000:
This PR add a cli to discordeno, just a help command and generate command


Skillz4Killz:
Should this not be its own package?

Skillz4Killz (approved):

PR #3091, opened 2023-07-29
fix(bot): Transform hashes correctly

NeKzor:
This fixes audit log entries which contain changes to guild icon, webhook icon etc. The problem was that hashes were incorrectly interpreted as snowflake IDs.


Skillz4Killz (approved):

Skillz4Killz:
LGTM

PR #3092, opened 2023-07-29
fix(gateway): update deno detection

Endy3032:
1.35.2 removed `window` from node's `globalThis` (https://github.com/denoland/deno/pull/19307)


Skillz4Killz (approved):

Skillz4Killz:
LGTM

PR #3097, opened 2023-08-03
Bug(utils) Fix bucket refill after the first time

### LINKED ISSUE #2841: bug(utils): bucket only refill once

Fleny113:
Closes #2841

### DIFF
--- a/packages/utils/src/bucket.ts
+++ b/packages/utils/src/bucket.ts
@@ -1,5 +1,5 @@
-import logger from './logger.js';
-import { delay } from './utils.js';
+import logger from './logger.js'
+import { delay } from './utils.js'
 
 export class LeakyBucket implements LeakyBucketOptions {
   max: number
--- a/packages/utils/src/bucket.ts
+++ b/packages/utils/src/bucket.ts
@@ -1,5 +1,5 @@
-import logger from './logger.js'
-import { delay } from './utils.js'
+import logger from './logger.js';
+import { delay } from './utils.js';
 
 export class LeakyBucket implements LeakyBucketOptions {
   max: number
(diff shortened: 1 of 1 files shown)


Skillz4Killz:
I need more details on the bug and what the bug does and why this change fixes it so i can make sure this change doesnt have any accidental side effects

Fleny113:
> I need more details on the bug and what the bug does

When a bucket is created it doesn't have a timer attached to so [this check](https://github.com/discordeno/discordeno/blob/0aeadbb27558d223c0631f854c982ad181a82d5a/packages/utils/src/bucket.ts#L64) fires and creates it. The problem is that when the `refillBucket` method is called it clears the `refillsAt` but the timeout doesn't get cleared so in the variable there is still a "timed out" timer that doesn't do anything useful, so when the code that created the timeout in the `processQueue` method checks for a timeout it finds the expired one and don't create a new one, blocking the event loop in the while when the `this.used` get's to the `this.max` as there isn't a remaining nor does the refillsAt value

> why this change fixes it so i can make sure this change doesnt have any accidental side effects

Since the problem is the expired instance of the timeout, I clear the timeout (mark it as expired basically) and then set it to undefined, so if the [`this.used > 0`](https://github.com/discordeno/discordeno/blob/0aeadbb27558d223c0631f854c982ad181a82d5a/packages/utils/src/bucket.ts#L39C8-L39C22) needs to create the new timeout, it does but if not the `this.timeoutId` is kept as undefined so that the check that created the first timeout can re-create it to refill the bucket once again.

Also, the `clearTimeout` is needed in the situation when the `refillBucket` is called manually[^1] so that the `refillBucket` is not called again by a possibly created timeout

[^1]: I don't know if this should be guarded against, in the doubt i added it

Skillz4Killz (approved):

Skillz4Killz:
LGTM

PR #3098, opened 2023-08-03
Bug(gateway) Fix fetch of discord gateway session information

Fleny113:
* Fix for the `bot.start()` method where the `if (!options.gateway?.connection)` always return false since the create `createGatewayManager` function created the value if non-existing.

* Add checks to try to avoid changing user options when doing the fetching


Skillz4Killz (approved):

Skillz4Killz:
LGTM

PR #3104, opened 2023-08-24
fix(rest): invert unauthorized check

Fleny113:
Fix bug where the only way to not attach the REST authorization header was setting unauthorized to false. Now the name and behavior match


itohatweb (approved):
Good catch!

PR #3114, opened 2023-08-24
feat: Implement OAuth2 endpoints

### LINKED ISSUE #2133: Add OAuth2

### LINKED ISSUE #2750: Add helper to add guild member
Docs: https://discord.com/developers/docs/resources/guild#add-guild-member

Fleny113:
Implement the REST endpoints and ratelimit handling for OAuth2 endpoints

Closes #2133
Closes #2750


Fleny113:
There are still somethings that I'm not sure about, like how to manage correctly the multiple queues and don't display duplicates in the debug logs, but most of the pr is done

packages/bot/src/helpers.ts
```diff
@@ -109,14 +121,14 @@ export function createBotHelpers(bot: Bot): BotHelpers {
     createForumThread: async (channelId, options, reason) => {
       return bot.transformers.channel(bot, { channel: snakelize(await bot.rest.createForumThread(channelId, options, reason)) })
     },
-    createGlobalApplicationCommand: async (command) => {
-      return bot.transformers.applicationCommand(bot, snakelize(await bot.rest.createGlobalApplicationCommand(command)))
+    createGlobalApplicationCommand: async (command, bearer) => {
```
Skillz4Killz, line 124: How would you feel about making this bearer into an options?

```ts
interface XXXX {
  bearerToken: string;
}
```

The reasoning/motivation for this is to make it so that for the future any changes would be non-breaking. For example, if tomorrow discord adds another option for this function, it would make this a breaking change, since u may end up having `createGlobalApplicationCommand(command, undefined, xyz)`

I would prefer we start making new things using objects to prevent breaking changes in future.

packages/bot/src/helpers.ts
```diff
@@ -232,11 +244,23 @@ export function createBotHelpers(bot: Bot): BotHelpers {
...
+    exchangeToken: async (options) => {
+      return await bot.rest.exchangeToken(options)
+    },
+    revokeToken: async (options) => {
+      return await bot.rest.revokeToken(options)
+    },
+    getApplicationCommandPermission: async (guildId, commandId, options) => {
+      return bot.transformers.applicationCommandPermission(
+        bot,
+        snakelize(await bot.rest.getApplicationCommandPermission(guildId, commandId, options)),
```
Skillz4Killz, line 259: Not sure why the snakelize is here. I dont mean that its unnecessary just trying to understand the logic/reasoning here. Perhaps ur right and it is necessary

Skillz4Killz (commented):
Looks great! Few questions

packages/bot/src/helpers.ts
Fleny113: I added it as the `rest.get` method calls the camelize method, and in the [transformer for the application command](<https://github.com/discordeno/discordeno/blob/6ea1360a07e72aa7d7f7742ec1a196b6c391d800/packages/bot/src/transformers/applicationCommandPermission.ts>) it uses the `application_id` as snake_case, so it would get undefined

Skillz4Killz: Doesn't createBot enable the snake case option by default now? The reason i ask is looking at this code, im thinking is this wasted CPU by running snakelize on something thats already snakelized

Fleny113: I will check at runtime what actually happens later

Fleny113: It is in fact needed, 
image

for context this is res and snakedRes
```ts
    getApplicationCommandPermission: async (guildId, commandId, options) => {
      const res = await bot.rest.getApplicationCommandPermission(guildId, commandId, options)
      const snakedRes = snakelize(res)

      return bot.transformers.applicationCommandPermission(bot, snakedRes)
    },
``` and the debugger is after the snakelize

\* i have censored the ids

File: packages/rest/src/manager.ts [hunk 3]
```diff
@@ -868,18 +931,78 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
...
+      const restOptions: MakeRequestOptions = {
+        body,
+        headers: {
+          'content-type': 'application/x-www-form-urlencoded',
+        },
+        unauthorized: true,
+      }
+
+      if (body.grantType === 'client_credentials') {
+        const basicCredentials = Buffer.from(`${body.clientId}:${body.clientSecret}`)
```
Fleny113, line 989: just noticed that `Buffer.from` does not work with deno, i have to find another solution

packages/bot/src/helpers.ts
Fleny113: I have made the change, let me know if i should also do this for all the other methods that accept a bearer token

PR #3116, opened 2023-08-28
fix(bot): properly handle thread list sync event

livelove1987:


packages/bot/src/bot.ts
```diff
@@ -144,6 +141,7 @@ export interface EventHandlers {
   automodActionExecution: (payload: AutoModerationActionExecution) => unknown
   threadCreate: (thread: Channel) => unknown
   threadDelete: (thread: Channel) => unknown
+  threadListSync: (payload: { guildId: bigint; channelIds: bigint[] | unknown; threads: Channel[]; members: ThreadMember[] }) => unknown
```
itohatweb, line 144: just one small thing
```suggestion
  threadListSync: (payload: { guildId: bigint; channelIds?: bigint[]; threads: Channel[]; members: ThreadMember[] }) => unknown
```

itohatweb (changes requested):

itohatweb (approved):

packages/bot/src/bot.ts
livelove1987: oh , ye that actually much better

PR #3117, opened 2023-08-30
Fix(rest) Fix error when serializing payload with a BigInt

Fleny113:
Fix `TypeError: Do not know how to serialize a BigInt` from `JSON.stringify()` when the options contains both a file, so the body requires to be a form data, and has a body with a BigInt somewhere in the body.

### DIFF
--- a/packages/rest/src/manager.ts
+++ b/packages/rest/src/manager.ts
@@ -175,6 +175,7 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
           form.append(`file${i}`, options.files[i].blob, options.files[i].name)
         }
 
+        // Have to use changeToDiscordFormat or else JSON.stringify may throw an error for the presence of BigInt(s) in the json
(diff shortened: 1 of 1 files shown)


packages/rest/src/manager.ts
```diff
@@ -175,7 +175,7 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
           form.append(`file${i}`, options.files[i].blob, options.files[i].name)
         }
 
-        form.append('payload_json', JSON.stringify(snakelize({ ...options.body, files: undefined })))
+        form.append('payload_json', JSON.stringify(rest.changeToDiscordFormat({ ...options.body, files: undefined })))
```
Skillz4Killz, line 178: Would like to have a comment above this line, explaining why we use rest.changeToDiscordFormat and not snakelize to avoid someone accidentally changing to snakelize here

Fleny113: I have added a comment for it, let me know if it gives an explanation that is understandable

Skillz4Killz (approved):

Skillz4Killz:
LGTM

PR #3118, opened 2023-09-02
Fix: added getActiveThreads return value transformer

livelove1987:


packages/bot/src/helpers.ts
```diff
@@ -227,7 +226,11 @@ export function createBotHelpers(bot: Bot): BotHelpers {
       return await bot.rest.followAnnouncement(sourceChannelId, targetChannelId)
     },
     getActiveThreads: async (guildId) => {
-      return await bot.rest.getActiveThreads(guildId)
+      const result = await bot.rest.getActiveThreads(guildId)
```
Skillz4Killz, line 229: thoughts on making these a collection

livelove1987: I think `threads` should be treated same as channels returned by `getChannels` function. Also there are no helper functions which return Collection

Skillz4Killz (approved):

Issue #3119, opened 2023-09-04
[v19] communicationsDisabledUntil wrong typing


MatthewSH:
According to the [current typings](https://github.com/discordeno/discordeno/blob/main/packages/types/src/discordeno.ts#L1070) the expected value of `communicationDisabledUntil` is going to be a number, however that's not correct according to my tests.

I was testing a timeout command, for testing I took current date and added 5 minutes and converted back to ISO and then to milliseconds via `date-fns` (I'm terrible with dates so I take extra precautions.
```typescript
const now = new Date()
const addedTime = add(now, {
  minutes: 5,
})
const formatted = getTime(new Date(formatISO(addedTime)))
```

However, when I run it through this:
```typescript
bot.helpers.editMember(
  guildId,
  userToTimeout,
  { communicationDisabledUntil: formatted },
  'Timeout'
)
```

I get this as the error:
```
{
  ok: false,
  status: 400,
  body: '{"message": "Invalid Form Body", "code": 50035, "errors": {"communication_disabled_until": {"_errors": [{"code": "DATE_TIME_TYPE_PARSE", "message": "Could not parse 1693807660000. Should be ISO8601."}]}}}'
}
```

I got curious so I ran 2 versions to test:
<img width="644" alt="image" src="https://github.com/discordeno/discordeno/assets/3768988/4b63e457-5da7-4d30-b1fa-74aa29a77521">

Well, to my surprise I noticed the second version worked without fail:
<img width="425" alt="image" src="https://github.com/discordeno/discordeno/assets/3768988/4002e301-2e54-45ae-9bea-09bf8fa6e7d3">
_Worth noting that I cut off the error in the screenshot, but it's the exact error as above_

I believe this is just a typing issue, unless someone else sees something different. I can fix the typing right now to just expect a string.
EDIT: Went ahead and opened a PR: #3120 

Current version installed: `19.0.0-next.ceb3d4d`

PR #3120, opened 2023-09-04
fix(types): change communicationDisabledUntil to string or null

MatthewSH:
Reported in #3119, this fixes the typing for `communicationDisabledUntil`.


packages/types/src/discordeno.ts
```diff
@@ -1079,7 +1079,7 @@ export interface ModifyGuildMember {
   /** Id of channel to move user to (if they are connected to voice). Requires the `MOVE_MEMBERS` permission */
   channelId?: BigString | null
   /** when the user's timeout will expire and the user will be able to communicate in the guild again (up to 28 days in the future), set to null to remove timeout. Requires the `MODERATE_MEMBERS` permission */
```
Skillz4Killz, line 1081:
```suggestion
  /** when the user's timeout will expire and the user will be able to communicate in the guild again (up to 28 days in the future), set to null to remove timeout. Requires the `MODERATE_MEMBERS` permission. The date must be given in a ISO string form. */
```

Skillz4Killz (approved):
LGTM! I added a small note to help users know the string is a ISO form of a date string.

Issue #3121, opened 2023-09-18
Header injection for discordeno/rest


8auu:
I am currently in the process of developing custom instances for my Discord bot. I'd like to a global rest process for all of the custom instances to send requests too. 

Currently, this is only possible by using the `authorization` header that is configurable in `createRestManager`. However, this is a very janky way of doing things as it means we have to use the property meant for authenticating a request to also include some data, which has no relevance to authentication...

The solution I'd like implemented is not just for custom bots, but it would also allow for custom dynamic headers to be injected into requests, which would allow developers to know exactly where the requests are coming from. For example, I have 2 bot processes, a gateway and a bot dashboard - I'd love to be able to have the following custom header:  `process: "bot-dashboard"` to know where the request originated from.

There are two possible solutions which I have came up with.

1. Add a new property to the `createRestManager` class
This solution would allow for static headers to be injected into the payload.

```ts
export const REST = createRestManager({
  token: process.env.PUBLIC_BOT_TOKEN as string,
  proxy: {
    baseUrl: process.env.PUBLIC_BOT_REST_URL as string,
    authorization: process.env.PUBLIC_BOT_REST_AUTHORIZATION as string,
  },
  customHeaders: {
    processName: "Bot-dashboard",
    botId: "123...",
  }
});
```

2. Add a function which runs on every request called `headerInjection`.

This function could be overridden to inject dynamic headers onto the request payload, such as a datetime, counter or something else...

```ts
const rest = createRestManager()

rest.injectHeaders= (...) => {
  // do some magic here...
}
```

PR #3122, opened 2023-09-18
feat: rest header injection 

8auu:
created a function to allow users to inject custom headers into a rest request. See issue #3121 (Solution 2)

Example usage:

```ts
const REST = createRestManager({
  token: process.env.PUBLIC_BOT_TOKEN as string,
})

REST.injectHeaders = () => {
  return {
    botId: '123',
  }
}
```


itohatweb:
Hey, just a quick question, is there a reason for it being a function?

8auu:
> Hey, just a quick question, is there a reason for it being a function?

A function would allow a developer to use dynamic values that need to be changed on each request, a property inside of the createRestManager class would not, unless you were overriding it every time you wanted a value to change which could be a little tedious.

A dynamic date header which changes based on when the request was made - Although, now I think about this example its not entirely useful unless you want to know how long a request takes to resolve and log it. As you could just create a new Date on the rests server instead.

I cannot come up with any other ideas, but I'm sure someone would have something down the line. Adds a bit more flexibility to what can be customized.

Skillz4Killz (changes requested):

Skillz4Killz (approved):

Issue #3124, opened 2023-09-18
[Feat]: Add builders for components, modals, and embeds


MatthewSH:
There is already a draft PR (#2931) but hasn't been touched since April and was failing last build.

I was wanting to open up an issue to track the feature and see if anyone has any feedback on how they'd like to implement it. If there's nothing and the draft PR is still valid then I'll take that PR and just rewrite it to ensure it works on today's branch and split up some PRs so it's one builder at a time.

```[tasklist]
### Builder List
- [x] Embeds
- [ ] Components
- [ ] Modals
```

PR #3145, opened 2023-10-01
feat(utils): add embed builder

MatthewSH:
This is a PR implementing the embed builder from #3124, based on PR #2931.

# Differences from other PR
- This is only embed builders, others will come later.
- Uses DD Discord interfaces and uses `Omit` to pull out any key we use as required.
- Uses [ECMAScript private getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields#private_methods) to get the current embed instead of a method.
- Simplified code a bit, the modifying the getter directly vs setting new variable and modifying. This is more preference.
- Using `set` before method names to make it more descriptive of intent of method
- Custom `EmbedBuilderError`
- Use spread operations where possible.
- Optional `validate` method that check all embeds to ensure they're in the [documented embed limits](https://discord.com/developers/docs/resources/channel#embed-object-embed-limits) for added safety if desired.

----

```[tasklist]
# TODO
- [x] Complete `TODO` in code
- [x] Add method documentation
- [x] Finish the `validate` method
- [x] Testing
```

### DIFF
--- a/packages/utils/src/builders/embeds.ts
+++ b/packages/utils/src/builders/embeds.ts
@@ -47,7 +47,7 @@ export class EmbedsBuilder extends Array<DiscordEmbed> {
    */
   newEmbed(): EmbedsBuilder {
     if (this.length >= 10) {
-      throw new EmbedsBuilderError("Maximum embed count exceeded. You can not have more than 10 embeds.")
+      throw new Error("Maximum embed count exceeded. You can not have more than 10 embeds.")
     }
 
     this.push({})
@@ -106,7 +106,7 @@ export class EmbedsBuilder extends Array<DiscordEmbed> {
     }
 
     if (index >= this.length || index < 0) {
-      throw new EmbedsBuilderError("Can not set the current embed to a index out of bounds.")
+      throw new Error("Can not set the current embed to a index out of bounds.")
     }
 
     this.#currentEmbedIndex = index;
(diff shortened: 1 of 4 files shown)


packages/utils/src/builders/embeds.ts
```diff
@@ -0,0 +1,387 @@
...
+}
+
+/**
+ * Custom EmbedsBuilder error.
+ *
+ * @class EmbedsBuilderError
+ * @typedef {EmbedsBuilderError}
+ * @extends {Error}
+ */
+class EmbedsBuilderError extends Error {
```
Skillz4Killz, line 381: Why is this class needed?

MatthewSH: @Skillz4Killz mainly just experience. Setting a custom error allows you to check, if desired, what the error name is and allows you to handle that error differently than others.

There's no real downside to not throwing generic errors vs custom errors. I don't feel strongly either way, I can just throw a generic error instead if that's what we decide.

Skillz4Killz: I think we should keep it generic please. You can compare error message as well.

MatthewSH: Resolved in 38e1a5ba0678c5db45cb117a605da37903fd6dc1

Skillz4Killz (approved):

PR #3185, opened 2023-11-13
fix(rest)!: Fix Basic OAuth2 queues and add method for updating the queues and ratelimit paths after a refresh

### LINKED ISSUE #3137: [Rest] Full support for using multiple bot tokens
Currently, you can have a rest proxy and have your bot send requests to Discord API through that, and with #3122, you can now have a single rest proxy for multiple bots. This is great, however, there's a few things to be addressed to make this truly reliable for multiple bots:

## Ratelimit handling:
Current behavior is assuming we use only one token, thus ratelimit handling is done in one place only. But we need a way for ratelimits to be stored and handled individually for different bot tokens used by rest requests, so that they all can utilize their individual limits fully. As an example, with the current implementation, if there are 5 bots trying to send a message in the same channel using the same rest proxy, they can only send 5 messages per 5 secs but if the ratelimit handling is split individually for as many bot tokens used, they can send 25 messages per 5 secs (5 per bot). So they'll basically have to have different ratelimit data stored for different tokens and **also global ratelimit should only affect a single bot and not all bots**.

## Invalid Requests handling:
I believe current implementation might not need much changes regarding this but while making a PR to do ratelimit handling individually, it should be made sure that it doesn't end up having different invalid requests tracker, this must continue to remain global, as invalid requests for the cloudflare ban is based on IP and not token.

### LINKED ISSUE #3165: [Rest] OAuth2 bearer tokens and refresh of tokens
**Describe the bug**:
Given a Bearer token got after a code exchange with discord the token gets used to do requests and discordeno creates queues mapped in the restManager.queues with the following key: `Bearer <token here>/<url here>`[^url]. The problem is when the access token gets refreshed (even if it's before it expires) the token changes but the ratelimit is still the same, also the old queue that was created might still have pending request to send but since refreshing a token implies invaliding the old one the pending requests will fail with a `401 & Unauthorized` error and if a new request gets sent before the ratelimit for an endpoint is refreshed[^ratelimit-example] it will fail with a `429 & Too many requests`

[^url]: The / is in fact part of the url, it is explicit in to separate the token and url
[^ratelimit-example]: For example: a token is used to do a request on `/users/@me/guilds` and then the token refreshes and another request is sent to that same url it will fail

**To Reproduce**:
```ts
import { createRestManager } from "@discordeno/rest";

const token = "<bot token>";
const secret = "<client secret for the OAuth2 exchange>";
const id = "<your bot id>"
const code = "<the code given by an oauth2 code exchange flow>"

const manager = createRestManager({
    token,
});

let token = await manager.exchangeToken(id, secret, {
    grantType: "authorization_code",
    redirectUri: "<redirect url of the application>",
    code,
});

await manager.getGuilds(token.accessToken);

token = await manager.exchangeToken(id, secret, {
    grantType: "refresh_token",
    refreshToken: token.refreshToken,
    accessToken: token.accessToken,
});

await manager.getGuilds(token.accessToken);
```

**Expected behavior**:
There shouldn't be an Invalid request performed for `429 & Too many requests`

**Screenshots**:
image
*The censored lines in the log are an input for the code to exchange*

**Version details**:

Discordeno version: v19 dev, [commit eec292728571d60bfe04c87260279d969a328be2](https://github.com/discordeno/discordeno/commit/eec292728571d60bfe04c87260279d969a328be2)

Fleny113:
This pull request addresses 3 issues:
- Fixes #3165, with the `updateTokenQueues` method to be called by the user-code to update the queues with the new token
- Fixes a bug where Basic request got stuck in an infinite ratelimit
- Fixes #3137 as with every ratelimit information (queue and rateLimitedPaths) being prefixed by the authorization they refer to a rest proxy can handle more then 1 bot token at once


Skillz4Killz (dismissed):

Skillz4Killz (dismissed):
The PR looks good, please have this tested properly before merging but the general idea looks good to me

Fleny113:
> The PR looks good, please have this tested properly before merging but the general idea looks good to me

I'm currently running this pr in production, atm i do not have something that uses OAuth ready to test this on (but the test scripts i made while making the code changes are working fine). 
If someone wants to test the pr for themself (with or without OAuth2 related code/requests) that would be appreciated

packages/rest/src/manager.ts
```diff
@@ -338,10 +402,9 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
 
       const loggingHeaders = { ...payload.headers }
 
-      const authenticationScheme = payload.headers.authorization?.split(' ')[0]
-
       if (payload.headers.authorization) {
-        loggingHeaders.authorization = `${authenticationScheme} tokenhere`
+        const authorizationScheme = payload.headers.authorization?.split(' ')[0]
+        loggingHeaders.authorization = `${authorizationScheme} tokenhere`
```
lts20050703, line 407: Is the tokenhere intentional?

Fleny113: yes. thay will be logged with the depth set to1 and the log level to debug but it would be best to avoid printing the token for security

lts20050703 (approved):
LGTM! You mentioned that you disabled auto-merge so this can be tested a bit more before merge, it's been a month so what do you think?

Fleny113:
> LGTM! You mentioned that you disabled auto-merge so this can be tested a bit more before merge, it's been a month so what do you think?

Didn't had the time to setup the stuff i need for the OAuth2 testing, but the bot auth is working on my prod bot. If i can today i will test it and enable auto merge accordingly

Fleny113:
Re enabling the auto-merge as from my testing I did today it seems to be fine. It does carry over the ratelimit from the previous queue and it does change the auth header preventing the 401

Skillz4Killz (approved):

PR #3186, opened 2023-11-15
role flags

rosefighter:


Skillz4Killz (changes requested):
Good stuff! Just missing the addition in the transformer file

PR #3188, opened 2023-11-26
fix(bot): Review checks to prevent JS from excluding values that might be useful

Fleny113:
In the transformer before there were the nullability checks that were using JS implicit conversion from anything to a boolean, this caused values like 0 to be converted to false and not put in the transformed object where the 0 could actually indicate a value that could have been useful.

---

Reported by @vxern on discord


packages/bot/src/transformers/channel.ts
```diff
@@ -142,7 +141,7 @@ export interface Channel extends BaseChannel {
   /** The type of channel */
   type: ChannelTypes
   /** The id of the guild */
-  guildId?: bigint
+  guildId?: BigString
```
MatthewSH, line 144: this should be bigint

MatthewSH (changes requested):

MatthewSH (approved):

MatthewSH (approved):

H01001000 (approved):

H01001000:
@Fleny113 need you update branch

PR #3195, opened 2023-11-28
chore(deps)!: upgrade all deps to new major versions

MatthewSH:
This goes through every package we can and upgrade it to the latest possible version without breaking the website or packages.


H01001000 (changes requested):
docusaurus build failed

MatthewSH:
> docusaurus build failed

reverting back to 2.x of Docusaurus due to many issues.

H01001000:
yea that's what i did b4

MatthewSH (commented):
Just did a final review of changes. Should be good. Ran through type test and unit test locally. All passing. Release build going through as well. No issues with formatting.

H01001000 (approved):

PR #3196, opened 2023-11-28
chore: move rest proxy example to examples

MatthewSH:
Originally it sat in the root, this just moves it to examples.


MatthewSH:
Didn't realize there's docker image built and published around this. Is this docker image still needed? I was under the impression this was an example. I could be wrong.

PR #3255, opened 2023-12-03
feat: cli draft

H01001000:
The PR generates a message type file from the .discordenorc.json config file.
The type file is located at `node_modules/.discordeno/transformer/src`


.discordenorc.json
```diff
@@ -0,0 +1,5 @@
+{
```
Skillz4Killz, line 1: should this be in the repo? also i thought we wanted to go with discordeno.config.js file name?

packages/discordeno/src/bin/generators/messageTypeGenerator.ts
```diff
@@ -0,0 +1,142 @@
+const messageTypeProps = {
```
Skillz4Killz, line 1: is this autogenerated or manually created? if autogenerated it shouldnt be in the repo, if manual, im not a fan of having this written twice here.

packages/discordeno/src/bin/types/DiscordenoConfig.ts
```diff
@@ -0,0 +1,5 @@
+export interface DiscordenoConfig {
+  desiredProps: {
+    message: string[]
```
Skillz4Killz, line 3: whats this for?

PR #3257, opened 2023-12-04
feat: api docs catchup part 1

MatthewSH:
I'm working on catching up on some of the more straight forward API docs. Each commit is tied to an issue. 
A second set of eyes is appreciated to make sure I don't miss anything.


Fleny113 (approved):

H01001000 (approved):
I trust Matt and Fleny

PR #3269, opened 2023-12-07
feat!: rewrite proxy app with fastify

MatthewSH:
This is a pretty big change so I would like some feedback.

This not only will remove the old rest proxy application but will provide the next major version (v2.0) of the Docker image built with Fastify. I've moved it under a folder called `docker-apps` to help separate it form the rest of the project and hopefully be a bit more descriptive. 

Why Fastify? On paper, in their [benchmarks](https://fastify.dev/benchmarks/) Fastify laps Express in terms of requests per second. While this is not everything in terms of API development, in my experience Express does provide a lot of overhead which is useful for full applications but not so useful for our use case. 

I also added [Helmet](https://github.com/fastify/fastify-helmet) which will just help secure the proxy just a bit more without adding too much overhead.

```[tasklist]
### To-Do
- [x] Update & review workflow
- [x] Add CHANGELOG generation and versioning
- [x] Add Dockerfile
- [x] Remove old rest proxy files
```


docker-apps/rest-proxy/src/server.ts
```diff
@@ -0,0 +1,52 @@
...
+      message: 'Credentials not valid.',
+    })
+  }
+})
+
+app.all('/*', async (request, reply) => {
+  const url = request.originalUrl
+
+  try {
+    const result = await discordRestManager.makeRequest(request.method as RequestMethods, url, {
```
MatthewSH, line 28: In the [current rest proxy](https://github.com/discordeno/discordeno/blob/d81b28a24bb449ebc0aac169cfb53ce098f6986c/proxies/rest/src/index.ts#L24) we have a section that determines if a version is passed through. will this still be required for the proxy to function?

MatthewSH: Tested this and it doesn't seem like it matters.

MatthewSH:
Removed Influx from the todo list. It requires a bit too much configuration for me to add it in the initial release.

Fleny113 (changes requested):
At the moment it causes 404 as the version is not removed from the url

MatthewSH:
> At the moment it causes 404 as the version is not removed from the url

What causes the version in the URL? I was testing locally with DD Next and had no issues? 

I can lift and shift the version detection in the other one over. Should be a quick fix.

Fleny113:
> > At the moment it causes 404 as the version is not removed from the url
> 
> What causes the version in the URL? I was testing locally with DD Next and had no issues?
> 
> I can lift and shift the version detection in the other one over. Should be a quick fix.

I didn't debugged much tbh, i used an already exiting testing bot i had, changed the url to point to the local instance i was running of the proxy and logging the url had the `/v10/` prefixed causing a 404, i do need to check where the version is actually added tho

MatthewSH:
> > > At the moment it causes 404 as the version is not removed from the url
> > 
> > 
> > What causes the version in the URL? I was testing locally with DD Next and had no issues?
> > I can lift and shift the version detection in the other one over. Should be a quick fix.
> 
> I didn't debugged much tbh, i used an already exiting testing bot i had, changed the url to point to the local instance i was running of the proxy and logging the url had the `/v10/` prefixed causing a 404, i do need to check where the version is actually added tho
> 
> EDIT: found the code that is adding the version, [packages/rest/src/manager.ts:475](https://github.com/discordeno/discordeno/blob/388f5fd5d24e6911de65930042b3709e7923ed75/packages/rest/src/manager.ts#L475C52-L475C69)

It's fixed now, I added the old version removal code.

Fleny113 (dismissed):
Seems to work fine now

Fleny113 (approved):

meister03 (approved):
Looks good

H01001000 (approved):

PR #3271, opened 2023-12-07
chore: mark bot, utils, types as beta

H01001000:


AwesomeStickz (approved):
I trust H0 will never make a mistake so this gets an approval

PR #3332, opened 2023-12-20
fix: docker app workflow

MatthewSH:


H01001000 (approved):

Fleny113 (approved):

H01001000 (changes requested):
Keep the path filter for push, don’t want release every push

H01001000 (approved):

Fleny113 (approved):

PR #3383, opened 2024-01-21
feat(bot, rest, gateway, utils): Add support for custom loggers

### LINKED ISSUE #3367: [Feat]: Pass a logger into components, rather than relying on console.log
### Tell us what problem the feature would solve

The usecase is primarily to redirect logging messages into any structured logger, rather than relying on capturing the console output.

### Describe the solution you'd like

The top-level object creators should allow passing a logger like so:

```ts
interface Logger {
  debug: (...args: unknown[]) => void,
  log: (...args: unknown[]) => void,
  warn: (...args: unknown[]) => void,
  error: (...args: unknown[]) => void,
  fatal: (...args: unknown[]) => void,
}

// redirecting all logs of REST manager to directly submit into some logging implementation
createRestManager({ token: ..., logger: createLogger({ prefix: "DD: REST Client" });
```

### Describe alternatives you've considered

There are currently no alternatives, except capturing the console like so, which then lacks the ability to discern the source of the log (unless analyzing the stack):

```ts
{
    // capturing console and redirecting it into winston
    const logger = loggerFactory("BASE:!Console");
    console.log = (...args: Parameters<typeof console.log>) => logger.info(args);
    console.warn = (...args: Parameters<typeof console.warn>) => logger.warn(args);
    console.error = (...args: Parameters<typeof console.error>) => logger.error(args);
    console.debug = (...args: Parameters<typeof console.debug>) => logger.debug(args);
}
```

### Additional Info

_No response_

Fleny113:
Allows to pass a custom logger object with at least the `debug`, `info`, `warn`, `error`, `fatal` to prevent Typescript error as much as possible to every class/object that uses the logger from `@discordeno/utils`

The `createBot` accepts a factory to create the `REST` named, `GATEWAY` named and `BOT` named loggers so the user can differentiate them based on the need of LogLevel for example

closes #3367


H01001000 (dismissed):
seems gd

Skillz4Killz (dismissed):

Skillz4Killz:
Fix Conflicts please

Skillz4Killz (approved):

vxern (approved):
Looks good to me!

PR #3355, opened 2024-01-07
chore(website): Cleanup docusaurus setup

Fleny113:
- Fixes eslint (and lint-staged) setup for website
- Migrates the TS setup from docusaurus v2 to v3
- Updates dependencies
- Adds Automatic JSX runtime (no more `import React from 'react'`)
- Add typecheck workflow job as docusaurus does not use types for the build
- Update typedoc setup to fix warns

PR #3385, opened 2024-01-21
docs: Add examples for Node.js and Bun, update Deno example

Fleny113:
- Add a minimal example for Node.js and Bun along with the Deno example.
- Update the Deno example to remove the reference to the gateway fix that is no longer necessary

depends on #3355


website/docs/beginner/env.md
```diff
@@ -3,51 +3,128 @@ sidebar_position: 4
...
+#### `dotenv` package
+
+If you don't want or can't use the `--env-file` option you can do it with the `dotenv` package
+
+First of all you need to install the `dotenv` package along it's type if you are using TypeScript
+
+`npm install dotenv @types/dotenv --save-dev`
+_or using yarn, pnpm or bun_
+
+We now need to load the `.env` from our code so let's add the following code to your typescript file:
```
MatthewSH, line 52: you can also use dotenv on command line:
```
node -r dotenv/config script.js
```

This will load the `.env` from where-ever you're running it.

website/docs/examples/deno.md
```diff
@@ -1,36 +1,37 @@
...
 sidebar_label: Using with Deno
 ---
 
 # Using with Deno
 
-To be able to use Discordeno with Deno there is one workaround needed if you are using the gateway package.
-We need to make Deno use the websocket from Deno instead of the one which is used in node, because the npm support in Deno cannot use our websocket correctly.
+Discordeno supports Deno by using the npm: specifier for your import.
 
-To do it you can use a workaround provided here: https://nest.land/package/katsura/files/src/discordenoFixes/gatewaySocket.ts
```
MatthewSH, line 11: Is this fix no longer required?

MatthewSH (changes requested):

website/docs/examples/deno.md
Fleny113: i tried with latest and it was connecting fine without it. If it has long term i have no clue

website/docs/beginner/env.md
Fleny113: i know but importing the dotenv package is more known way to do this, i guess it could be mentioned tho

website/docs/examples/deno.md
MatthewSH: sounds good to me.

website/docs/beginner/env.md
MatthewSH: Just wondering if it's worth mentioning a non-code change based alternative? This is, funny enough, how I started using dotenv a while ago so I don't have to use `.env` files on prod and only for dev

Fleny113: If we want to be sincere: before i started using --env-file from node (what i do prefer now) i was using `import 'dotenv/config'` that is basically both. So you can argue for all 3, but the import with side effects i would avoid in a beginned scoped docs, and using preloading is adding yet another option to dotenv and --env-file that it is not that big of a deal honestly and in dotenv readme itself you need you need to scroll a bit to find it

MatthewSH: fine by me! just wanted to throw another option out there, but your stance makes sense. I haven't upgraded to 20 yet, can't wait to not have to use dotenv.

MatthewSH (approved):

MatthewSH (approved):

H01001000 (approved):

PR #3422, opened 2024-02-11
refactor(bot, gateway, rest)!: Improve the interface of the `createBot()` function.

vxern:
This PR seeks to improve the interface of the `createBot()` function by doing the following:
- `options.events` is now optional (previously the user had to pass in an empty object).
- `options.gateway.token` is now also optional since it can be inherited from `options.token`.
- `options.gateway.cache.requestMembers.pending` is now optional given that it gets set to an empty `Collection` object anyway by `createGatewayManager()`.

### DIFF
--- a/packages/bot/src/bot.ts
+++ b/packages/bot/src/bot.ts
@@ -27,6 +27,7 @@ import type { Sticker } from './transformers/sticker.js'
 import type { ThreadMember } from './transformers/threadMember.js'
 import type { User } from './transformers/user.js'
 import type { VoiceState } from './transformers/voiceState.js'
+import type { BotGatewayHandlerOptions } from './typings.js'
 
 /**
  * Create a bot object that will maintain the rest and gateway connection.
@@ -36,7 +37,10 @@ import type { VoiceState } from './transformers/voiceState.js'
  */
 export function createBot(options: CreateBotOptions): Bot {
   if (!options.rest) options.rest = { token: options.token, applicationId: options.applicationId }
-  if (!options.gateway) options.gateway = { token: options.token, events: {} }
+  if (!options.rest.token) options.rest.token = options.token
+  if (!options.gateway) options.gateway = { token: options.token }
+  if (!options.gateway.token) options.gateway.token = options.token
+  if (!options.gateway.events) options.gateway.events = {}
   if (!options.gateway.events.message) {
     options.gateway.events.message = async (shard, data) => {
       // TRIGGER RAW EVENT
@@ -50,7 +54,6 @@ export function createBot(options: CreateBotOptions): Bot {
     }
   }
 
-  options.rest.token = options.token
   options.gateway.intents = options.intents
   options.gateway.preferSnakeCase = true
 
@@ -59,8 +62,8 @@ export function createBot(options: CreateBotOptions): Bot {
   const bot: Bot = {
     id,
     applicationId: id,
-    transformers: createTransformers({}, { defaultDesiredPropertiesValue: options.defaultDesiredPropertiesValue ?? false }),
-    handlers: createBotGatewayHandlers({}),
+    transformers: createTransformers(options.transformers ?? {}, { defaultDesiredPropertiesValue: options.defaultDesiredPropertiesValue ?? false }),
+    handlers: createBotGatewayHandlers(options.handlers ?? {}),
     rest: createRestManager(options.rest),
     gateway: createGatewayManager(options.gateway),
     events: options.events ?? {},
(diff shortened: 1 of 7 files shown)


packages/gateway/src/manager.ts
```diff
@@ -50,7 +50,7 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
     cache: {
       requestMembers: {
         enabled: options.cache?.requestMembers?.enabled ?? false,
-        pending: new Collection(),
+        pending: options.cache?.requestMembers?.pending ?? new Collection(),
```
Fleny113, line 53:
```suggestion
        pending: new Collection(),
```

*Just my opinion*, i don't think having this option has much of a point at all, i known it was already there so would be a breaking change but i don't see the value. If someone does please tell me

AwesomeStickz (approved):

packages/bot/src/bot.ts
```diff
@@ -101,11 +104,15 @@ export interface CreateBotOptions {
...
   /** Any options you wish to provide to the gateway manager. */
-  gateway?: CreateGatewayManagerOptions
+  gateway?: CreateGatewayManagerOptions & Partial<Pick<CreateGatewayManagerOptions, 'token'>>
   /** The event handlers. */
-  events: Partial<EventHandlers>
+  events?: Partial<EventHandlers>
+  /** The functions that should transform discord objects to discordeno shaped objects. */
+  transformers?: Partial<Transformers>
+  /** The handler functions that should handle incoming discord payloads from gateway and call an event. */
+  handlers?: Partial<BotGatewayHandlerOptions>
```
Fleny113, line 115:
```suggestion
  /** The functions that should customize the discordeno shaped objects */
  customizers?: Partial<Transformers["customizers"]>
```

I usually think of transformers and handlers of a more internal part of discordeno rather then something that should be exposed, especially in the createBot, for example the handlers 99% of the time you are not going to edit them because if you do it's not something common so overring the function after the creation is not that big of deal, same for transformers, they tend to be code that you don't want to mained as they require a lot of if checks for both proprieties that need to exist and desired proprieties that need to be enabled or even worse if the transformer has a base template (eg. Message, Interaction ecc) as those would be impossibile to change in here without coping that entire part of the code

Fleny113 (changes requested):
*The request change is for the transformer comment, the comment on the pending collection for member collection is just a comment*

Also: CI is failing

packages/gateway/src/manager.ts
vxern: Yeah, I guessed as much, but I'd rather that deciding to remove this property be left up to somebody with greater contributor experience with the library, perhaps it could be discussed in the dev chat at some point

Fleny113: I only bring this up because you made me remember of it, so we will talk for this on discord as it might be better to give this a pr separately

packages/bot/src/bot.ts
vxern: For handlers, I could agree that maybe those are "internal enough" to not be exposed, but I feel like there are still very valid cases for overriding/plugging into transformers, the immediately obvious use for it being caching.

On the other hand, I believe that even if these are/could be considered internals, whether that's transformers or events or handlers, allowing the developer to override them in a consistent interface if they need to is still helpful. After all, Discordeno was designed for big bot developers who are more or less expected to know the outcomes of what they plug into or change in the library.

vxern:
Whoops, sorry about that failing test. I removed it now given it was made redundant by `rest.token` now being required.

Fleny113:
> Whoops, sorry about that failing test. I removed it now given it was made redundant by `rest.token` now being required.

packages/bot/src/handlers/members/GUILD_MEMBERS_CHUNK.ts is failing for the nullability in the bot.gateway.caching.requestMembers.pending collection, other then the tests

packages/bot/src/bot.ts
Fleny113: > the immediately obvious use for it being caching.

I do not see the value of caching the parsing of structure at all, caching and merging the objects can be done in the customizers


> After all, Discordeno was designed for big bot developers who are more or less expected to know the outcomes of what they plug into or change in the library.

You do have a point, although i still personally still feel they don't belong, so for now i will leave the conversation as "unresolved" to see if anyone else as an opinion on this

packages/bot/src/handlers/members/GUILD_MEMBERS_CHUNK.ts
```diff
@@ -11,7 +11,7 @@ export async function handleGuildMembersChunk(bot: Bot, data: DiscordGatewayPayl
   // If this request has no nonce, skip checks.
   if (!payload.nonce) return
 
-  const pending = bot.gateway.cache.requestMembers.pending.get(payload.nonce)
+  const pending = bot.gateway.cache.requestMembers.pending!.get(payload.nonce)
```
Fleny113, line 14: rather than suppressing the null errors, it would probably be best to update the GatewayManager interface to remove the nullable from cache.requestMembers.pending

Skillz4Killz (approved):

packages/bot/src/bot.ts
Skillz4Killz: After hearing both sides, I prefer to lean towards the side that DD should be flexible and built for those who know how to work with it.

PR #3423, opened 2024-02-11
fix(bot, types): `description` being a required property on `CreateSlashApplicationCommand`.

vxern:
The `description` is not required as indicated here:
https://discord.com/developers/docs/interactions/application-commands#create-guild-application-command


Fleny113 (changes requested):
While the api documentation does agree with you, the actual api implementation does not, when trying out this with curl i do get the following api response:
```json
{
  "message": "Invalid Form Body",
  "code": 50035,
  "errors": {
    "description": {
      "_errors": [
        {
          "code": "BASE_TYPE_REQUIRED",
          "message": "This field is required"
        }
      ]
    }
  }
}
```

Fleny113:
Looking on this further this is already according to the docs by using a discriminated unions for type 2 and type 3 commands (user and message commands respectivly)

PR #3424, opened 2024-02-11
feat: Add voice state to the desired properties object.

vxern:


packages/bot/src/transformers/voiceState.ts
```diff
@@ -3,16 +3,17 @@ import type { Bot } from '../index.js'
...
-
-    channelId: payload.voiceState.channel_id ? bot.transformers.snowflake(payload.voiceState.channel_id) : undefined,
-    guildId: payload.guildId || (payload.voiceState.guild_id ? bot.transformers.snowflake(payload.voiceState.guild_id) : 0n),
-    userId: payload.voiceState.user_id ? bot.transformers.snowflake(payload.voiceState.user_id) : 0n,
-  } as VoiceState
+  if (props.requestToSpeakTimestamp && payload.voiceState.request_to_speak_timestamp)
+    voiceState.requestToSpeakTimestamp = Date.parse(payload.voiceState.request_to_speak_timestamp)
+  if (props.channelId && payload.voiceState.channel_id) voiceState.channelId = bot.transformers.snowflake(payload.voiceState.channel_id)
+  if (props.guildId)
+    voiceState.guildId = payload.guildId || (payload.voiceState.guild_id ? bot.transformers.snowflake(payload.voiceState.guild_id) : 0n)
```
Fleny113, line 13:
```suggestion
    voiceState.guildId = payload.guildId
```

First of all, using `||` is not a great idea because of javascript and it's faulsy values (even tho it does not apply in this situation (or at least shouldn't)), second `payload.guild` is always present based on the typings for it

Fleny113 (changes requested):

packages/bot/src/transformers/voiceState.ts
vxern: I kept `||` to keep the functionality the same after the PR, even if it could have been wrong. I think converting these "otherwise if falsey" checks would deserve its own PR if we're going in that direction.

Fleny113: Uh, i did not see at all that it was already in that way, so yeah that could have it's own pr, sorry

Fleny113 (approved):

packages/bot/src/transformers/voiceState.ts
vxern: All good, thank you

MatthewSH (approved):

PR #3453, opened 2024-02-22
feat(cli)!: CLI generated types

### LINKED ISSUE #2983: [Bot] Customize Typings Based on DesiredProperties
**Is your feature request related to a problem? Please describe.** 
It would be better to have typings be generated automatically based on the desired properties the user provides.

**Describe the solution you'd like** 
- Possibly a CLI like `npx prisma generate` where we do `npx discordeno generate`
- Typescript generic that user provides on `createTransformers<>()` which can provide custom ones

**Describe alternatives you've considered**

**Additional context** 

With current solution it is not typesafe as a desiredPropety could be set to false and yet the typings say it exists. We could make all properties optional but that isn't exactly a good UX either.

### LINKED ISSUE #3190: [CLI] npx discordeno generate
I've created a basic mockup of how this code should work. Need to finish this up with 1 type before we start implementing for all types. 

Try to see if we can get the Message.id type created.

Fleny113:
Add Autogenerated types based on Desired Proprieties.

This will also remove the nested objects inside other structures and base objects (`BaseMember`, `BaseUser`, ...), this allows a simpler code on both the information collection and the generation of the interfaces

> [!WARNING]
> This is breaking because due to several reasons:
> - The removal of nested object
> - `ApplicationCommandPermission` got renamed to `GuildApplicationCommandPermission` and `ApplicationCommandPermission` was removed from the `@discordeno/type` package and instead we are now using `CamelizedDiscordApplicationCommandPermission`
> - The removal of nested object in desired proprieties
> - Some desired proprieties got removed, some got added

> [!CAUTION]
> Deno will **NOT** be supported, it may be looked into the future how to support Deno.
> The current plan is to use `node_modules` to change the types, making Deno impossibile to support with this approch
>
> The only way Deno is supported is if the user tells deno to use `node_modules` for npm packages

## TODO

- [x] Use the TS compiler api to get the discordeno types
  - [x] Merge all the interfaces into one single file
- [x] Create the generator for the interfaces
  - [x] Generate interface with type and name
  - [x] Handle optional types (`propName?: propType`)
  - [x] Handle array types
  - [x] Handle TSDoc / JSDoc comments
- [x] Create the cli code
  - [x] Unify TS Compiler Api and emitter
  - [x] Source path for input file from fs
  - [x] Add tests to make sure future pr(s) don't break the cli
  - [x] Implement the type replacer for when a prop is not desired
  - [x] Source the configuration from a config file

#### Other Changes

Changed the module resolution to `Node16`. This shouldn't break anything as we require node >= 18. `Node16` is the recommended module resolution and the `node` alias for `Node10` is deprecated

#### Issues

- fixes #3190
- fixes #2983
- closes #3255


packages/discordeno/src/bin/generate/typescript.ts
```diff
@@ -0,0 +1,193 @@
+import assert from 'node:assert'
+import { createWriteStream } from 'node:fs'
+import ts from 'typescript'
+
+// TODO: change this fileName to be based on the filesystem
+const fileName = process.argv.at(2)!
```
lts20050703, line 6: 1. I think it's better to throw an error if the user forgot to provide a file name, maybe something like this?
```suggestion
// TODO: change this fileName to be based on the filesystem
const fileName = process.argv.at(2)
if (!fileName) throw new Error("You need to provide a file name!")
```

2. What does "change this fileName to be based on the filesystem" mean?

scripts/TSApiTest.ts
```diff
@@ -0,0 +1,119 @@
...
+    return {
+      type: typeText,
+      isOptional: (member.getFlags() & ts.SymbolFlags.Optional) === ts.SymbolFlags.Optional,
+      documentation: member.getDocumentationComment(checker),
+      jsDoc: member.getJsDocTags(checker),
+    }
+  }
+}
+
+generateDocumentation(process.argv.at(2)!, {
```
lts20050703, line 111:
```suggestion
const fileName = process.argv.at(2)
if (!fileName) throw new Error("You need to provide a file name!")
generateDocumentation(fileName, {
```

packages/discordeno/src/bin/generate/typescript.ts
Fleny113: the script is not meant to be run manually, the code will attempt to get the .d.ts file from the node_modules (currently it uses a .ts so that needs work too)

packages/discordeno/tests/generate.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    const typesFile = await findUp('packages/bot/dist/types/transformers/types.d.ts', {
+      allowSymlinks: true,
+    })
+
+    expect(typesFile).to.exist
+
+    if (!typesFile) {
+      throw new Error("This shouldn't happen!")
```
lts20050703, line 24:
```suggestion
    if (!typesFile) {
      throw new Error("Transformers types file not found!")
```

lts20050703 (approved):
LGTM!

Fleny113:
The only missing thing is to allow `createBot()` to use the desired proprieties from the configuration to be the used at runtime.

This can be difficult for some reasons, including:
- the file is/isn't compiled based on the compiler option
- in a monorepo, the file will be (provably) outside the typescript `rootDir`, resulting in an error

So the user will have the issue with loading this file

however currently if you manage to import it, you can just do this:
```ts
// or whatever are you using to get this object
import config from "../discordeno.config.js"

bot.transformers.desiredProperties = config.desiredProperties.properties
```

PR #3536, opened 2024-04-11
fix(build): Fix CJS extension script to avoid replacing .json

Fleny113:
Currently the script uses a RegExP to change all the `.js` references to `.cjs` to avoid runtime errors as since we have `"type": "module"` in our package.json all the `.js` files are getting interpreted as ESModules and we need to use the `.cjs` extension for CommonJS. The script just checked for the presence of `.js` anywhere in the code and replaced with `.cjs` but this caused errors in the REST package as the `Response` object that `fetch()` returns has a `.json` method to get the JSON body but `.json` starting with `.js` was being matched by the RegExP and was creating unexpected runtime errors. This pr adds a RegEx Negative LookAhead to make sure we aren't replacing any `.json` instance.


MatthewSH (approved):
LGTM! Nice catch!

lts20050703 (approved):

PR #3550, opened 2024-05-05
fix(bot,utils): Add poll transformer export & fix typedoc warn

Fleny113:
Typedoc was warning on the generation about not being able to resolve the `scopes` (*it's a typo of `scope`*) in the `CreateOAuth2LinkOptions` and the fact that the Poll & PollMedia types were not exported out from the package. This pr fixes both issues, the second one affecting the imports of `@discordeno/bot` when trying to import the Poll types


lts20050703 (approved):
LGTM!

AwesomeStickz (approved):

PR #3552, opened 2024-05-06
chore: Code changes & formatting and linting

Fleny113:
I will explain pretty much everything in the PR, but a lot of the changes are the result of running prettier on the entire project and ESLint --fix on most of the files.

The formatting is taking most of the file changes with very minimal ones often but might as well run it one on everything than everytime we need to touch that file (even if by a merge) eslint/prettier edits them and then we got to reverse that change because it's unrelated.

- `.devcontainer/devcontainer.json`: Add a `yarn install` after the dev container is created. Currently if you create a GitHub CodeSpace for example the VSCode task for the build:watch fails as it is missing the node_modules folder.
- `packages/benchmark`: This does affect some other files, like `.github/workflows/benchmark.yml`, the reason this got changed is because while I was testing the benchmarks node (v22 at least) was failing to resolve the `benchmark` package (the npm package), because it was actually trying to resolve the package itself, so I renamed the package to `benchmarks` in the `package.json`, but at that point I have just renamed the folder.
- `.npmignore`: On the npm package we have the `.tsbuildinfo` and we don't need it in a package, so we can just remove it.
- `README.md`: Add the bun unit test script, since there was already the deno unit script, might as well add the bun unit script.
- `packages/benchmarks/package.json`: The `main` & `exports` values were expecting the dist folder to also have the cjs files, but the scripts are to build only with esm, so to make it run i have fixed the values.
- `packages/benchmarks/src/benchmarks/casting.ts`: The `Camelize 1 event` was actually running snakelize on an already snakelized event, so it didn't make any sense.
- `packages/benchmarks/src/benchmarks/memory.ts`: The old code was achiving the same thing as `snakeToCamelCase`, so no reason to use that code instead of the function we already have.
    - also applies to `packages/bot/tests/e2e/resetguilds.spec.ts`
- Varius `tsconfig.json`: 
    - `tsconfig.test.json`: was not getting picked up by typescript (at least in vscode, tested also with a clean codespace) so it got moved to the test directory with the `tsconfig.json` name and it seems to be getting picked up by typescript. 
    - The source tsconfig got edited to explicly state what is the `rootDir` for the source, use an include pattern instead of an exclude one and set explicitly what is the path for the `.tsbuildinfo` file.
    - `packages/tsconfig/base.json`: The exclude pattern now uses `${configDir}` since this PR includes the TS 5.5 bump. It also moves the rest of the configs to this base (`rootDir`, `outDir` and `tsBuildInfoFile`) to the base using `${configDir}` as well, making the packages tsconfig just the `extends`. The `skipDefaultLib` is just skipping the check for typescript `.d.ts` files.
- `packages/gateway/tests/integration/connection.spec.ts`: ESLint warnings/errors mainly
- `scripts/checkCpuModel.js`: ESLint error
- `packages/rest/tests/e2e/utils.ts`: This affected most if not all e2e files, just a rename to use camelCase instead of all lowercase.
- `scripts/finalizeTypedocs.js`: Node 22 compatibility, node 22 removed support for `assert` in imports statements and the import function, so it broke when run with node 22. This does not affect node 18+ as it seems to be working under a built of the latest LTS of node 18.
- Varius `package.json`:
    - `packages/benchmarks/package.json`: `microtime` is not used, `node-fetch` is not necessary considering we support only node 18+ where the built-in fetch got stable.
    -  `packages/rest/package.json`: We do use `dotenv` for our tests (to use the `.env` file in the root of the directory in e2e tests) but we don't need to specifiy it in the "prod" dependencies. `why-is-node-running` is not used, and doesn't seem to have a reason to be there either
    - `packages/types/package.json`: In the types package we just have an "import correctly" test, so chai is not necessary, also `@types/simon` without `simon` installed is not doing anything, just creating clutter in the `package.json`.
    - `packages/utils/package.json`: While we might want to readd the public key validation for HTTP interactions (from what I understand, at some point we had the code to do support this), currently we don't need `tweetnacl` to be installed.

If there is something that isn't clear why it got edited, i will do my best to respond with a reason, reason that might just prettier/eslint that edited or why i did edit it.


packages/benchmarks/src/generateMessage.ts
```diff
@@ -4,7 +4,6 @@ const benchmarkData = await fetch(`https://raw.githubusercontent.com/discordeno/
   .then(async (res) => await res.text())
   .then((text) => JSON.parse(text.slice(24)))
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars, no-unused-vars
 // const commitSha = await fs.readFile('./sha', 'utf-8')
```
lts20050703, line 7:
```suggestion
```

This feels like debug code to me, I think we should just remove all of this?

packages/gateway/tests/integration/connection.spec.ts
```diff
@@ -120,7 +119,7 @@ const createUws = async (options: {
         }
         // retrieve listening port
         uwsToken = token
```
lts20050703, line 121:
```suggestion
```

packages/gateway/tests/integration/connection.spec.ts
```diff
@@ -120,7 +119,7 @@ const createUws = async (options: {
...
-        port = uWS.us_socket_local_port(token)
+        port = uWS.us_socket_local_port(uwsToken)
```
lts20050703, line 122:
```suggestion
        port = uWS.us_socket_local_port(token)
```

packages/gateway/tests/integration/connection.spec.ts
```diff
@@ -120,7 +119,7 @@ const createUws = async (options: {
...
         resolve({ port, uwsToken })
```
lts20050703, line 123:
```suggestion
        resolve({ port, uwsToken: token })
```

packages/gateway/tests/integration/connection.spec.ts
lts20050703: For some reason GitHub doesn't allow me to select three lines and make one suggestion, so please check the other two below

packages/gateway/tests/integration/connection.spec.ts
lts20050703: For some reason GitHub doesn't allow me to select three lines and make one suggestion, so please check the one above and the one below

packages/gateway/tests/integration/connection.spec.ts
lts20050703: For some reason GitHub doesn't allow me to select three lines and make one suggestion, so please check the other two above

packages/gateway/tests/integration/connection.spec.ts
```diff
@@ -162,7 +161,7 @@ describe('gateway', () => {
     const timeout = setTimeout(() => {
       throw new Error('Not heartbeat in time')
     }, 1017)
```
lts20050703, line 163: Why is the duration so specific??? Like I would expect it to be 1000ms or something like that

packages/gateway/tests/integration/connection.spec.ts
Fleny113: This causes an issue with ESLint. Since token is any (really good typings) eslint will error with: "Unsafe argument of type `any` assigned to a parameter of type `us_socket`." But with an `as` it can be used

packages/gateway/tests/integration/connection.spec.ts
Fleny113: I have no clue why is it this much specific

vxern: At first glance, it looks like something you would get as a factor of some equation, though I couldn't guess where it would come from either.

Fleny113: Seems to be a value to account for the jitter, but the uWS uses the same value as hearbeat_interval so i don't understand what is going on this this value

Fleny113: The value turns out to be a value to make `Shard.calculateSafeRequests` return something that isn't 0. `calculateSafeRequests` is there for determinate how many messages we can send before a shard gets its ratelimit reset & send enough heartbeats for it to not get closed by discord. The previous value (1017) is the lowest value for which the calculation `120 - (Math.ceil(60_000 / <heartbeat_interval>) * 2)`[^calc] returns 2. I have changed the value to 1050 to make it seem a little less odd and add a comment to explain what is that value and to make aware anyone editing the test that it has the `1017` constrain as a minimum value.

[^calc]: I have inlined some constants, both 120 & 60_000 are from the Shard class.

lts20050703: For future references on why this number was chosen

image

lts20050703 (approved):
LGTM!

Skillz4Killz:
@Fleny113 conflicts need fixing plz b4 mergr

PR #3557, opened 2024-05-09
feat(docs): Add getting-started with installation

Fleny113:
Add a getting-started page to the docs. The getting started will include the installation and desired proprieties explanation as they tend to be very confusing at first


lts20050703 (dismissed):
LGTM!

I see that's there's a "TODO: Add a link to the page on how to use OAuth2 in Discordeno.". I think it's okay to merge this and do the TODO in a new PR.

As for "<!-- TODO: Do we want to keep this info box? Or rewrite a bit this --", I'm thinking of something like "`Thank you for choosing Discordeno! Discordeno is designed to easily make scalable bots, so it might not be for you. However, we hope that you at least give it a try and don't hesitate to join and ask for help in our Discord server!`"

If it sounds too "begging people" please let me know lol, I'm fine with the current text (also fine if the text was removed)

Fleny113:
> I think it's okay to merge this and do the TODO in a new PR.

It would be a bit too much for one pr, i don't want to forget about it, so i added a todo, that is it

> I'm thinking of something like ...

I wanted to ask because i don't know what the best solution is, discordeno does have some not completely beginner friendly things (for example Desired Proprieties or the non-exiting caching by default) but it is still valuable imho to know, and i thought it sounded too "hard"

website/docs/getting-started.md
```diff
@@ -0,0 +1,153 @@
...
+## Minimal bot that logins into Discord
+
+You can use `@discordeno/bot` as follows:
+
+```ts
+import { createBot, Intents } from '@discordeno/bot'
+
+const bot = createBot({
+  token: 'YOUR BOT TOKEN',
+  intents: Intents.Guilds | GuildMessages, // Or other intents that you might needs.
```
AwesomeStickz, line 36:
```suggestion
  intents: Intents.Guilds | Intents.GuildMessages, // Or other intents that you might needs.
```

Fleny113: Whoops, i forgot the Intents while copy pasting stuff around

website/docs/getting-started.md
```diff
@@ -0,0 +1,153 @@
...
+
+As well as managing your connection to the Discord gateway, Discordeno also deals with handling the required handling for HTTP/REST requests too. Discordeno will manage for you the ratelimit of the Discord endpoints, these ratelimit includes the invalid request hard limit for the [Cloudflare ban](https://discord.com/developers/docs/topics/rate-limits#invalid-request-limit-aka-cloudflare-bans).
+
+In Discordeno you can use the `bot.helpers` object to have access to all the function that will perform some kind of actions on Discord. These methods will also give you some objects that will be slightly different from the one from Discord, a notable example are the IDs. Discord will give you a string with a number inside, this is to avoid some rounding errors in some languages (Javascript included), however this issue can be avoided when using [`BigInt`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) in Javascript. Discordeno will transform most (if not every) id that you get from a string to a `BigInt` format, however most methods in the `bot.helpers` object will accept both a `BigInt` or a normal string where content is the id.
```
AwesomeStickz, line 71:
```suggestion
In Discordeno, you can use the `bot.helpers` object to have access to all the function that will perform some kind of actions on Discord. These methods will also give you some objects that will be slightly different from the one from Discord, a notable example are the IDs. Discord will give you a string with a number inside, this is to avoid some rounding errors in some languages (Javascript included), however this issue can be avoided when using [`BigInt`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) in Javascript. Discordeno will transform most (if not every) id that you get from a string to a `BigInt` format, however most methods in the `bot.helpers` object will accept both a `BigInt` or a normal string where content is the id.
```

website/docs/getting-started.md
```diff
@@ -0,0 +1,153 @@
...
+// Change the id below with one of the channels of one of the server your bot is in
+const message = await bot.helpers.sendMessage(123123123123123123n, {
+  content: 'Hello world. This is test message from Discordeno.',
+})
+```
+
+:::tip[Naming of functions in Discordeno]
+Most of the functions in Discordeno tend to be very explicit in what they will do and be similar to how discord calls their endpoints in the documentation.
+
+Also helpers methods from Discordeno will always perform a single action. They will never call multiple Discord endpoints.
```
AwesomeStickz, line 88:
```suggestion
Also, helper methods from Discordeno will always perform a single action only. They will never call multiple Discord endpoints.
```

website/docs/getting-started.md
```diff
@@ -0,0 +1,153 @@
...
+:::
+
+## Understanding Desired Proprieties in Discordeno
```
AwesomeStickz, line 91:
```suggestion
## Understanding Desired Properties in Discordeno
```

website/docs/getting-started.md
```diff
@@ -0,0 +1,153 @@
...
+
+If we ran the code from above, for example on the ready event, the bot will send a message in the channel you specified, as long as it has the permissions to do so, with the content of "Hello world. This is test message from Discordeno.", however if we log to the console the message object we are getting we won't see many, if any, values on it. This is also the case from the events such as `MessageCreate`, but why is that? Well, this is a Discordeno feature called `Desired proprieties`.
```
AwesomeStickz, line 93:
```suggestion
If we ran the code from above, for example on the ready event, the bot will send a message in the channel you specified, as long as it has the permissions to do so, with the content of "Hello world. This is test message from Discordeno.". However, if we log to the console the message object we are getting, we won't see many, if any, values on it. This is also the case from the events such as `MessageCreate`, but why is that? Well, this is a Discordeno feature called `Desired proprieties`.
```

website/docs/getting-started.md
```diff
@@ -0,0 +1,153 @@
...
+
+Desired proprieties are a feature to allow for a smaller memory impact on your application for proprieties that you don't use. An example may be the channel topic, you might not be interested in knowing the topic of a channel, however Discord will always give it to you making your code have to deal with values that you will not use but still consume memory. This is why Discordeno requires you to explicitly set ahead of time the proprieties that you want/use.
```
AwesomeStickz, line 95:
```suggestion
Desired properties is a feature to allow for a smaller memory impact on your application for properties that you don't use. An example may be the channel topic, you might not be interested in knowing the topic of a channel, however, Discord will always give it to you making your code have to deal with values that you will not use but still consume memory. This is why Discordeno requires you to explicitly set ahead of time the properties that you want/use.
```

website/docs/getting-started.md
```diff
@@ -0,0 +1,153 @@
...
+
+You can set these from the `bot.transformers.desiredProprieties` object. This will contain an object for every type that supports this feature, such as `Message` or `Interaction` for example where inside you will find the single proprieties with a boolean value, or an object with more proprieties inside. Discordeno will default everything to false and you can set the values you need to true manually like this:
```
AwesomeStickz, line 97:
```suggestion
You can set these from the `bot.transformers.desiredProperties` object. This will contain an object for every type that supports this feature, such as `Message` or `Interaction` for example, inside which you will find the single properties with a boolean value, or an object with more properties inside. Discordeno will default everything to false and you can set the values you need to true manually like this:
```

website/docs/getting-started.md
```diff
@@ -0,0 +1,153 @@
...
+
+```ts
+bot.transformers.desiredProprieties.message.id = true
```
AwesomeStickz, line 100:
```suggestion
bot.transformers.desiredProperties.message.id = true
```

website/docs/getting-started.md
```diff
@@ -0,0 +1,153 @@
...
+bot.transformers.desiredProprieties.message.content = true
```
AwesomeStickz, line 101:
```suggestion
bot.transformers.desiredProperties.message.content = true
```

website/docs/getting-started.md
```diff
@@ -0,0 +1,153 @@
...
+bot.transformers.desiredProprieties.message.channelId = true
```
AwesomeStickz, line 102:
```suggestion
bot.transformers.desiredProperties.message.channelId = true
```

website/docs/getting-started.md
```diff
@@ -0,0 +1,153 @@
...
+```
+
+With the above 3 lines of code we will get the message ID, the channel ID and the message content\* for a specific message, and if we now see the object from both our sendMessage helper method and/or the messageCreate we will find the `id`, the `channelId` and the `content` proprieties to be present in the object.
```
AwesomeStickz, line 105:
```suggestion
With the above 3 lines of code, we will get the message ID, the channel ID and the message content\* for a specific message, and if we now see the object from both our `sendMessage` helper method and/or the `messageCreate` event, we will find the `id`, the `channelId` and the `content` proprieties to be present in the object.
```

website/docs/getting-started.md
```diff
@@ -0,0 +1,153 @@
...
+
+We could also the proprieties using the Javascript Object Syntax (`{ id: true, content: true, channelId: true }`), but that would require us to specify all others keys in the object with false, and doing something like that would get extremely annoying.
```
AwesomeStickz, line 107:
```suggestion
We can also set the properties using the Javascript Object Syntax (`{ id: true, content: true, channelId: true }`), but that would require us to specify all others keys in the object with false, and doing something like that would get extremely annoying.
```

website/docs/getting-started.md
```diff
@@ -0,0 +1,153 @@
...
+
+\*: As long as the required privileged intent is enabled.
+
+:::danger[Changing the default for Desired Proprieties]
```
AwesomeStickz, line 111:
```suggestion
:::danger[Changing the default for Desired Properties]
```

website/docs/getting-started.md
```diff
@@ -0,0 +1,153 @@
...
+THIS IS NOT RECOMMENDED IF YOU WANT TO SHIP YOUR BOT TO PRODUCTION.
+
+While not recommended, in the `createBot` we do allow for a value: `defaultDesiredPropertiesValue`. This if set to true every desired propriety by default, you can still disable some if you need. The reason why this is not recommended and marked as deprecated from the documentation is because while Desired Proprieties DO slow you down during development needing to make sure you aren't using something that you won't have a runtime they have a significant performance impact based on the proprieties you disable on both a CPU side and memory side. Also we are currently working on a solution to the current Typescript issues (see box below) and that may remove the `defaultDesiredPropertiesValue` value all together or move it to a new place.
```
AwesomeStickz, line 114:
```suggestion
While not recommended, in the `createBot` function, we do allow for a value: `defaultDesiredPropertiesValue`. This, if set to true will set every desired property to true by default, you can still disable some if you need. The reason why this is not recommended and marked as deprecated from the documentation is because while Desired Properties DO slow you down during development (needing to make sure you aren't using something that you won't have at runtime), they have a significant performance impact based on the properties you disable on both a CPU side and memory side. Also, we are currently working on a solution to the current TypeScript issues (see box below) and that may remove the `defaultDesiredPropertiesValue` value altogether or move it to a new place.
```

website/docs/getting-started.md
```diff
@@ -0,0 +1,153 @@
...
+:::
+
+:::warning[Typescript and Desired Proprieties]
```
AwesomeStickz, line 117:
```suggestion
:::warning[Typescript and Desired Properties]
```

website/docs/getting-started.md
```diff
@@ -0,0 +1,153 @@
...
+We are aware that Typescript, at this time has no idea that those propriety will be missing when we execute our code. We are working on fixing this issue but it is not yet ready at this time.
```
AwesomeStickz, line 118:
```suggestion
We are aware that TypeScript, at this time has no idea that those properties will be missing when we execute our code. We are working on fixing this issue but it is not yet ready at this time.
```

website/docs/getting-started.md
```diff
@@ -0,0 +1,153 @@
...
+
+Discordeno includes some utilities (such as `createOAuth2Link` to help create a link for the OAuth2 authorization link) and also the methods to do the exchange of the token and refresh of your exiting token.
+
+<!-- TODO: Add a link to the page on how to use OAuth2 in Discordeno. -->
+
+You can read more about OAuth2 in Discord on the [documentation](https://discord.com/developers/docs/topics/oauth2).
+
+### Use individual features from Discordeno separably
+
+While we do provide a `@discordeno/bot` package on npm with all the feature, you might just need a subset of those. For this reason Discordeno is split across multiple packages which are the following:
```
AwesomeStickz, line 141:
```suggestion
While we do provide a `@discordeno/bot` package on npm with all the features, you might just need a subset of those. For this reason, Discordeno is split across multiple packages which are the following:
```

website/docs/getting-started.md
```diff
@@ -0,0 +1,153 @@
...
+
+- `@discordeno/bot`: Groups all other packages and adds transformers and other functionalities on top.
+- `@discordeno/rest`: Will only provide the ratelimit features on the Discord REST Api. This will provide the same object as [`bot.rest`](#raw-rest-methods) when you create a `RestManager`.
```
AwesomeStickz, line 144:
```suggestion
- `@discordeno/rest`: Will only provide the methods you could use to send requests to Discord REST API, and ratelimit handling feature. This will provide the same object as [`bot.rest`](#raw-rest-methods) when you create a `RestManager`.
```

website/docs/getting-started.md
```diff
@@ -0,0 +1,153 @@
...
+- `@discordeno/gateway`: Will only provide the gateway connection to Discord. This will provide the same object as `bot.gateway` when you create a `GatewayManager`
```
AwesomeStickz, line 145:
```suggestion
- `@discordeno/gateway`: Will only provide the methods you could use to establish and manage gateway connections to Discord. This will provide the same object as `bot.gateway` when you create a `GatewayManager`
```

website/docs/getting-started.md
```diff
@@ -0,0 +1,153 @@
...
+- `@discordeno/utils`: Will only provide some utilities that are independent from other Discordeno features.
+- `@discordeno/types`: Will only provide the Discord & Discordeno types used in all other packages.
+
+The use-case for the separate packages may be more advanced compared to the use-cases where it makes more sense to simply use `@discordeno/bot`.
+
+### Raw REST methods
+
+While we recommend using `bot.helpers` you might find yourself in a situation where avoiding the transformers and desired proprieties can be beneficial, for that you can use the `bot.rest` methods, in this object there are the same methods from `bot.helpers` but with a more raw return object. The return objects will simply be the discord raw response but with the keys being camelCase instead of snake_case.
```
AwesomeStickz, line 153:
```suggestion
While we recommend using `bot.helpers`, you might find yourself in a situation where avoiding the transformers and desired properties can be beneficial. For that, you can use the `bot.rest` methods, in this object, there are the same methods from `bot.helpers` but with a more raw return object. The return objects will simply be the raw response from Discord, but with the keys being camelCase instead of snake_case.
```

website/docs/intro.md
```diff
@@ -24,17 +24,14 @@ Discordeno is actively maintained to guarantee **excellent performance, latest f
 - Freedom from 1 hour downtimes due to invalid requests
   - Prevent your bot from being down for an hour, by lowering the maximum downtime to 10 minutes.
 - Freedom from global rate limit errors
-  - As a bot grows, you need to handle global rate limits better. Shards don't communicate fast enough to truly
-    handle it properly. With one point of contact to discords API, you will never have issues again.
+  - As a bot grows, you need to handle global rate limits better. Shards don't communicate fast enough to truly handle it properly. With one point of contact to discords API, you will never have issues again.
```
AwesomeStickz, line 27:
```suggestion
  - As a bot grows, you need to handle global rate limits better. Shards don't communicate fast enough to truly handle it properly. With one point of contact to Discord API, you will never have issues again.
```

AwesomeStickz (changes requested):

website/docs/getting-started.md
```diff
@@ -0,0 +1,153 @@
...
+
+You can set these from the `bot.transformers.desiredProperties` object. This will contain an object for every type that supports this feature, such as `Message` or `Interaction` for example, inside which you will find the single properties with a boolean value, or an object with more properties inside. Discordeno will default everything to false and you can set the values you need to true manually like this:
+
+```ts
+bot.transformers.desiredProperties.message.id = true
+bot.transformers.desiredProperties.message.content = true
+bot.transformers.desiredProperties.message.channelId = true
+```
+
+With the above 3 lines of code, we will get the message ID, the channel ID and the message content\* for a specific message, and if we now see the object from both our `sendMessage` helper method and/or the `messageCreate` event, we will find the `id`, the `channelId` and the `content` proprieties to be present in the object.
```
AwesomeStickz, line 105:
```suggestion
With the above 3 lines of code, we will get the message ID, the channel ID and the message content\* for a specific message, and if we now see the object from both our `sendMessage` helper method and/or the `messageCreate` event, we will find the `id`, the `channelId` and the `content` properties to be present in the object.
```

AwesomeStickz (dismissed):

AwesomeStickz (approved):

lts20050703 (commented):
I'm reviewing this at 3AM so don't trust me lol, feel free to comment on any changes that doesn't make sense

website/docs/getting-started.md
```diff
@@ -0,0 +1,151 @@
...
+
+\*: As long as the required privileged intent is enabled.
+
+:::danger[Changing the default for Desired Properties]
+THIS IS NOT RECOMMENDED IF YOU WANT TO SHIP YOUR BOT TO PRODUCTION.
+
+While not recommended, in the `createBot` function, we do allow for a value: `defaultDesiredPropertiesValue`. This, if set to true will set every desired property to true by default, you can still disable some if you need. The reason why this is not recommended and marked as deprecated from the documentation is because while Desired Properties DO slow you down during development (needing to make sure you aren't using something that you won't have at runtime), they have a significant performance impact based on the properties you disable on both a CPU side and memory side. Also, we are currently working on a solution to the current TypeScript issues (see box below) and that may remove the `defaultDesiredPropertiesValue` value altogether or move it to a new place.
```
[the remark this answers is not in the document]
Fleny113: Why the ``` on the end? I don't see what they would be doing at least rn

lts20050703 (approved):
LGTM!

AwesomeStickz (approved):

PR #3558, opened 2024-05-10
fix(gateway): Remove bot token from the shard debug logs

Fleny113:
Currently we add to the debug logs for the shards the bot token when resume is called. In the rest package we do censor it with "tokenhere" while still showing the HTTP header, however in this situation there isn't the need to show what token is in use and it is better to not leak the bot token in a log file or something like that


lts20050703 (approved):
LGTM!

MatthewSH (approved):
LGTM2.

PR #3561, opened 2024-05-19
fix(rest,type): getPollAnswerVoters return type

kane50613:
According to [Get Answer Voters](https://discord.com/developers/docs/resources/poll#get-answer-voters), `getPollAnswerVoters` should return `users` object with an array of users.


Fleny113 (approved):

lts20050703 (approved):
LGTM!

PR #3614, opened 2024-05-25
feat(bot,types): Add invite types

### LINKED ISSUE #3603: [api-docs] Added invite type (#6568)
A new commit was made into the api-docs repo: https://github.com/discord/discord-api-docs/commit/4c6b571f3c0b7dcb1a6da13b48a0714aebdb0ee2

This is a bot created issue.

Fleny113:
Add invite types to the `DiscordInvite` object

- Upstream: https://github.com/discord/discord-api-docs/pull/6568
- fixes #3603


packages/bot/src/transformers/invite.ts
```diff
@@ -1,11 +1,12 @@
-import type { DiscordApplication, DiscordInviteCreate, DiscordInviteMetadata } from '@discordeno/types'
+import type { DiscordApplication, DiscordInviteCreate, DiscordInviteMetadata, DiscordInviteType } from '@discordeno/types'
 import { isInviteWithMetadata, type Application, type Bot, type ScheduledEvent, type User } from '../index.js'
 import type { InviteStageInstance } from './stageInviteInstance.js'
 
 export function transformInvite(bot: Bot, payload: { invite: DiscordInviteCreate | DiscordInviteMetadata; shardId: number }): Invite {
   const props = bot.transformers.desiredProperties.invite
   const invite = {} as Invite
 
+  if (props.type && 'type' in payload.invite) invite.type = payload.invite.type
```
lts20050703, line 9: Just in case other people wonder like I did: `payload.invite.type` is not used because invite type can be the number 0 which is falsy.

lts20050703 (dismissed):
LGTM!

packages/bot/src/transformers/invite.ts
Fleny113: That is not even the issue, i would have added !== undefined.

The issue is that DiscordInviteCreate does not have `type`
DiscordInvite does so also DiscordInviteMetadata does, the issue is that i couldn't think of a way to discriminate the union on `DiscordInvite` (with AND without the metadata) from the DiscordInviteCreate

lts20050703 (approved):

Skillz4Killz (approved):

Issue #3612, opened 2024-05-25
[Feat]: Use Flags enums instead of `number` in typings


Fleny113:
### Tell us what problem the feature would solve

Currently in a bunch of places, such as `DiscordUser` we use `number` on the `flags` and `public_flags`. However, discord does document what are the possibile flags, for the user flags (public and not) it is on the [User object user flags](https://discord.com/developers/docs/resources/user#user-object-user-flags)

### Describe the solution you'd like

Use/create the enums for the flags for there values

### Describe alternatives you've considered

Manually use the enum that exists with the exiting `number` type, however this does not guarantee the use of the currenct enum

lts20050703:
We already have ToggleBitfield (reference: https://github.com/discordeno/discordeno/tree/main/packages/bot/src/transformers/toggles), so yeah a flags and public_flags toggles would be nice

Fleny113:
> We already have ToggleBitfield (reference: https://github.com/discordeno/discordeno/tree/main/packages/bot/src/transformers/toggles), so yeah a flags and public_flags toggles would be nice

From what i understand the toggleBitfield are used to avoid having to store a considerable amount of booleans, not for flags. Correct me if i'm wrong

lts20050703:
Oh wait I'm dumb smh

But yeah a enum, something like the Intents enum would be nice

lts20050703:
Wait aren't we already have that? https://github.com/discordeno/discordeno/blob/ad7e74c4cd3f704442acea38dd890018ab0d1f86/packages/types/src/shared.ts#L19

Fleny113:
> Wait aren't we already have that? https://github.com/discordeno/discordeno/blob/ad7e74c4cd3f704442acea38dd890018ab0d1f86/packages/types/src/shared.ts#L19

In some cases we don't use them (user) in some we don't have it (member, idr if we merged that but still)

Fleny113:
We will add the enums, but we will keep the number typing, so this issue can be closed

PR #3617, opened 2024-05-25
fix(bot,types): guild members flags

### LINKED ISSUE #3568: [api-docs] Update permissions needed for PATCH member.flags field (#6817)
A new commit was made into the api-docs repo: https://github.com/discord/discord-api-docs/commit/88fc6590c933653b2391d40965086eb408944dec

This is a bot created issue.

### LINKED ISSUE #3602: [api-docs] add guild member flags to the guild member update event (#6717)
A new commit was made into the api-docs repo: https://github.com/discord/discord-api-docs/commit/db3c57ac9990fd0e5c4dd53748268be1f124d39f
This was added to the member object in #5874 and expanded upon in
 #5881 but never added to the guild member update gateway event.

This is a bot created issue.

Fleny113:
Add the previusly missing flags in the DiscordMember object
Add the flags to ModifyGuildMember to edit them for a specified member

- fixes #3602
  - Upstream: https://github.com/discord/discord-api-docs/pull/6717
- fixes #3568
  - Upstream: https://github.com/discord/discord-api-docs/pull/6817

~~Part of #3612~~


packages/bot/src/transformers/member.ts
```diff
@@ -35,6 +35,7 @@ export function transformMember(bot: Bot, payload: DiscordMember, guildId: BigSt
   if (props.deaf || props.mute || props.pending) {
     member.toggles = new MemberToggles(payload)
   }
+  if (props.flags && payload.flags) member.flags = payload.flags
```
lts20050703, line 38:
```suggestion
  if (props.flags && 'flags' in payload) member.flags = payload.flags
```
I'm pretty sure flags can be 0?

packages/bot/src/handlers/members/GUILD_MEMBER_UPDATE.ts
```diff
@@ -7,5 +7,6 @@ export async function handleGuildMemberUpdate(bot: Bot, data: DiscordGatewayPayl
   const payload = data.d as DiscordGuildMemberUpdate
 
   const user = bot.transformers.user(bot, payload.user)
+  // @ts-expect-error Flags in the update are nullable, while on the member they are be always present
```
lts20050703, line 10: Why is this ts-expect error needed?

packages/types/src/shared.ts
```diff
@@ -34,6 +34,32 @@ export enum UserFlags {
...
+   */
+  DidRejoin = 1 << 0,
+  /**
+   * Member has completed onboarding
+   *
+   * @remarks
+   * This value is not editable
+   */
+  CompletedOnboarding = 1 << 1,
+  /** Member is exempt from guild verification requirements */
```
lts20050703, line 52:
```suggestion
  /** Member is exempt from guild verification requirements, which allows a member who does not meet verification requirements to participate in a server. */
```
I don't think this is necessary, so feel free to just mark this as resolved.

Reference: https://discord.com/developers/docs/resources/guild#guild-member-object-guild-member-flags (the info box below)

packages/bot/src/transformers/member.ts
Fleny113: Oh right, i need to add !== undefined then, (see my comment on the invite type pr, i'm not going to use `in` in this case)

packages/bot/src/handlers/members/GUILD_MEMBER_UPDATE.ts
Fleny113: ~~isn't the explanation sufficent?~~

`DiscordGuildMemberUpdate` has `flags?: <idr the type name>`, while `DiscordMember` (what the transformer accepts) has `flags: <idr the type name>`, this causes a type error and for this reason there is the expect-error

packages/types/src/shared.ts
Fleny113: I feel like the discord one is already clear enough so this is isn't needed, feel free to say that i'm wrong

lts20050703 (approved):
LGTM!

lts20050703 (dismissed):

lts20050703 (approved):
LGTM! Merge when we finalized and agreed on the convention

packages/bot/src/transformers/member.ts
```diff
@@ -74,6 +75,8 @@ export interface Member extends BaseMember {
   permissions?: Permissions
   /** when the user's timeout will expire and the user will be able to communicate in the guild again (set null to remove timeout), null or a time in the past if the user is not timed out */
   communicationDisabledUntil?: number
+  /** Guild member flags */
+  flags: MemberFlag
```
lts20050703, line 79:
```suggestion
  flags: bigint
```

packages/types/src/discord.ts
```diff
@@ -329,6 +330,8 @@ export interface DiscordMember {
   permissions?: string
   /** when the user's timeout will expire and the user will be able to communicate in the guild again (set null to remove timeout), null or a time in the past if the user is not timed out */
   communication_disabled_until?: string | null
+  /** Guild member flags */
+  flags: MemberFlag
```
lts20050703, line 334:
```suggestion
  flags: string
```

packages/types/src/discord.ts
```diff
@@ -2843,6 +2846,8 @@ export interface DiscordGuildMemberUpdate {
   pending?: boolean
   /** when the user's [timeout](https://support.discord.com/hc/en-us/articles/4413305239191-Time-Out-FAQ) will expire and the user will be able to communicate in the guild again, null or a time in the past if the user is not timed out. Will throw a 403 error if the user has the ADMINISTRATOR permission or is the owner of the guild */
   communication_disabled_until?: string
+  /** Guild member flags */
+  flags?: MemberFlag
```
lts20050703, line 2850:
```suggestion
  flags?: string
```

packages/types/src/discordeno.ts
```diff
@@ -1238,8 +1239,10 @@ export interface ModifyGuildMember {
   deaf?: boolean | null
   /** Id of channel to move user to (if they are connected to voice). Requires the `MOVE_MEMBERS` permission */
   channelId?: BigString | null
-  /** when the user's timeout will expire and the user will be able to communicate in the guild again (up to 28 days in the future), set to null to remove timeout. Requires the `MODERATE_MEMBERS` permission. The date must be given in a ISO string form. */
+  /** When the user's timeout will expire and the user will be able to communicate in the guild again (up to 28 days in the future), set to null to remove timeout. Requires the `MODERATE_MEMBERS` permission. The date must be given in a ISO string form. */
   communicationDisabledUntil?: string | null
+  /** Set the flags for the guild member. Requires the `MANAGE_GUILD` or `MANAGE_ROLES` or the combination of `MODERATE_MEMBERS` and `KICK_MEMBERS` and `BAN_MEMBERS` */
+  flags?: MemberFlag
```
lts20050703, line 1245:
```suggestion
  flags?: BigString
```

lts20050703 (changes requested):
See my comment on the flag enums PR here https://github.com/discordeno/discordeno/pull/3635#pullrequestreview-2087191060

packages/types/src/discord.ts
Fleny113: i need to check but for this you might be right based on what discord uses in the response.

Skillz4Killz:
Should be moved to MemberToggles similar to Guild Flags on GuildToggles

Fleny113:
> Should be moved to MemberToggles similar to Guild Flags on GuildToggles

In guid there aren't any flags, do you mean like the guild features with then also the getter? Or you mean like with the user flags where we use the `ToggleBitfield`?

Skillz4Killz:
@Fleny113 Anything the toggles dont need a gui like u mentioned guild features. The toggles purpose is to optimize the memory needs. Storing booleans get expensive. So any boolean can be combined into the Toggles for 0 increase in memory. This is also why discord stores role/channel permissions as a bitwise flag. 

Always try to use toggles for boolean values in DD. this will help maximize efficiency while keeping the usability relativelyt the same exact thing because of getters

Fleny113:
> Always try to use toggles for boolean values in DD. this will help maximize efficiency while keeping the usability relativelyt the same exact thing because of getters

The thing that confused me is that the flags are already a bit set, not a bunch of bools.

So, it's a thing of putting the flags alongside the other 3 bool (`deaf`, `mute`, `pending`), which means adding the toggles for `didRejoin`, `compleatedOnboarding`, `bypassesVerification` and `startedOnboarding`, then change `member.flags` to use those toggles instead of storing the discord bit set?

lts20050703:
@Skillz4Killz As far as I can see there are three options

1. Keeping `member.flags` a number. Discordeno just export the `MemberFLags` enum so developers can import and do bitwise operation to check if a member has a flag. For example `if (member.flags & MemberFlags.DidRejoin) {}`
2. Merging Member Flags to the existing `MemberToggle`. For example `if (member.didRejoin) {}`
3. Making Member Flags something like the whole Permissions system. For example
```ts
const flags = new Flags(await calculateMemberFlags(guild, interaction.member))
const didRejoin = flags.has('DID_REJOIN')
```

Which one do you think is the best? I think we can all agree that option 3 is terrible, so it's between option 1 and option 2.

Personally, I'm fine with either option 1 or 2, but like Fleny said flags are already a bitset (like Permissions) and NOT a bunch of booleans, so I prefer option 1.

lts20050703:
@Skillz4Killz Currently the PR is doing option 1 (`member.flags` is a number), so we will need to wait for Fleny to update.

Just to confirm, you prefer `member.flags.didRejoin` instead of `member.didRejoin` right?

Skillz4Killz:
I don't have a preference of the getters I leave that to y'all to decide what u prefer. I just prefer minimal

Fleny113:
I have changed flags to be a `ToggleBitfield` like it is in the message and user transformers. And added the getters for the `MemberFlags`.

Is this ok or do you prefer getting the flags merged into the toggles? i'm unsure about merging them because this would be the only place where we threat flags this way (afaik)

lts20050703:
1. There are still build errors lol, maybe revert the changes, ignore the test errors for this PR, and then create another PR for the typedoc breaking changes fix?
2. If we want to save memory then getting the flags to be merged into the toggles will use less memory than yet another toggle. I'm fine with either option but in favor of merging the flags into the toggles.
TL;DR 1 toggle (1 number) uses less memory than 2 numbers (2 toggles)

Sorry, should have told you earlier. Thank you for everything you have done <3

Fleny113:
> 1. There are still build errors lol, maybe revert the changes, ignore the test errors for this PR, and then create another PR for the typedoc breaking changes fix?

The CI failure is typedoc fault, not this related to pr 

> 2. If we want to save memory then getting the flags to be merged into the toggles will use less memory than yet another toggle. I'm fine with either option but in favor of merging the flags into the toggles.
>    TL;DR 1 toggle (1 number) uses less memory than 2 numbers (2 toggles)

I'm fine with merging them, but i prefer the 2 separate. Since i don't want to argue for hours on this i will just merge them (after i try to fix the typedoc issues in another pr)

lts20050703 (dismissed):
Now that I have time to think about it, maybe we should keep them separate. I think developers will be confused if they don't see `flags` autocomplete when trying to check for a member's flag.

LGTM!

Skillz4Killz:
You use a getter to provide flags for auto complete. The toggles should be one to preserve memory please. This will be an incremental problem. With a single member object it's not a big deal but large bots have hundreds of millions of members so this needs optimizing like crazy

PR #3634, opened 2024-05-28
chore: Migrate ESLint and prettier to Biome

Fleny113:
This removes everything ESLint[^webpack] and Prettier related from the project, including the config package for eslint and the docs mention.

Reasons why we should migrate away from eslint:
- ESLint to run on the entire project will takes literal minutes (10min+ on my pc)
- ESLint to run at all even for one file will use more then 1.5 GB of RAM (2GB if we include Virtual Page on Windows, on Linux i did not test the exact memory usage) 
- The time it takes for `lint-staged` to run eslint is quite a bit (around 1min on my pc) which is a pretty big annoyance/road block for contributing

Reason to use Biome:
- It is reasonably fast (around 600ms to run on the entire project)
- It has pretty much the same rules as ESLint even if some do not are a direct copy and others are "inspired" (as per the term the biome documentation uses) so they have the same concept applied in a different thing
- Format and Lint can be a single command*
- ~~It is written in rust~~ (To be clear: this is a joke, not an actual reason)

\*: Currently the old `fmt` scripts are now `biome format` and `lint` scripts are `biome lint`. Should we use `check` on the `lint` or `fmt` (or both) scripts instead? (`check` will run both format and lint)

[^webpack]: `website/yarn.lock` still has some eslint packages, because apparently Webpack depends on some eslint packages


lts20050703 (dismissed):
LGTM! Since everything is just formatting changes I'm not really reviewing anything, though, but I'm all for reducing contribution friction (aka annoyance roadblock)

Though, I wonder if we should merge other PRs before this one or merge this PR before others first if we decided to merge this?

Fleny113:
> Though, I wonder if we should merge other PRs before this one or merge this PR before others first if we decided to merge this?

I think the only pr that will need some work to merge in if this gets merged before that one is the refactor pr since it is for a good part literally an eslint and prettier run, however i think that by doing a rebase i should be able to resolve merge conflicts fairly trivially. Others PRs, yes might have some conflicts especially for the import sorting biome does, but should be a faily easy conflict to resolve

lts20050703 (dismissed):

Fleny113:
Small issue i have found yesterday i didn't notice before: yarn and biome fight for the format of the package .json

This isn't much of an issue (i hope) however it is worth mentioning, especially if dependabot will format the file the yarn way. We can either disable biome for the package.json(s) or find a way to disable the yarn formatting of the package.json

PR #3635, opened 2024-05-29
feat(types,bot)!: Use flag Enums where possible

Fleny113:
This PR replaces some `number` type usage and replaces them with the proper enum. This means adding the `ActivityFlag` enum.

> [!IMPORTANT]
> This does NOT include the `MemberFlag` enum, that is included in #3617.

### Questions:
1. Since Enums should be singular, as it doesn't make sense that the enum is plural if it is a number effectively, it would be the same as saying `numbers` instead of `number`[^enumTSHandbook]. Should we rename the other flag enums to make them singular?
This is the case in other enums as well, like the `<Name>Types` enum(s) (or even entire types I think), those as well should be renamed based on what we decide but they are out of the scope of this PR since they are not flags, it will need a separate PR.
1. In the PR I have renamed some existing flag enums to remove the `Discord` prefix since some do not have it (eg. `ChannelFlags` and `UserFlags`). Should we keep it instead? It is somewhat inconsistent that (almost) everything out of `@discordeno/types` has the `Discord` prefix however flag enums are an exception to this.

---

Part of #3612.

> [!NOTE]
> This pull request currently contains some breaking changes, beside the type change which is a breaking change in itself:
> - Rename `DiscordSkuFlag` to `SkuFlag` for constistency[^enumsNames].
> - Rename `DiscordMessageFlag` to `MessageFlag` for constistency[^enumsNames].
> - Remove `MessageFlags` from `@discordeno/bot`, leave only the one in `@discordeno/types`

[^enumsNames]: See the questions above for more informations.
[^enumTSHandbook]: The typescript handbook itself on the enum docs uses signular names (eg. `Direction`). [TS Handbook link](https://www.typescriptlang.org/docs/handbook/enums.html)

### DIFF
--- a/packages/bot/src/transformers/sku.ts
+++ b/packages/bot/src/transformers/sku.ts
@@ -1,4 +1,4 @@
-import type { DiscordSku, DiscordSkuFlag, DiscordSkuType } from '@discordeno/types'
+import type { DiscordSku, DiscordSkuType, SkuFlag } from '@discordeno/types'
 import type { Bot } from '../index.js'
 
 export function transformSku(bot: Bot, payload: DiscordSku): Sku {
@@ -27,5 +27,5 @@ export interface Sku {
   /** System-generated URL slug based on the SKU's name */
   slug: string
   /** SKU flags combined as a bitfield */
-  flags: DiscordSkuFlag
+  flags: SkuFlag
 }
--- a/packages/bot/src/transformers/activity.ts
+++ b/packages/bot/src/transformers/activity.ts
@@ -1,4 +1,4 @@
-import type { ActivityTypes, Bot, DiscordActivity } from '../index.js'
+import type { ActivityFlag, ActivityTypes, Bot, DiscordActivity } from '../index.js'
 
 export function transformActivity(bot: Bot, payload: DiscordActivity): Activity {
   const activity = {
@@ -38,7 +38,7 @@ export function transformActivity(bot: Bot, payload: DiscordActivity): Activity 
 
 export interface Activity {
   join?: string
-  flags?: number
+  flags?: ActivityFlag
   applicationId?: bigint
   spectate?: string
   url?: string
--- a/packages/bot/src/transformers/channel.ts
+++ b/packages/bot/src/transformers/channel.ts
@@ -1,5 +1,13 @@
 /* eslint-disable @typescript-eslint/no-unsafe-argument */
-import type { BigString, ChannelTypes, DiscordChannel, DiscordThreadMember, OverwriteReadable, VideoQualityModes } from '@discordeno/types'
+import type {
+  BigString,
+  ChannelFlags,
+  ChannelTypes,
+  DiscordChannel,
+  DiscordThreadMember,
+  OverwriteReadable,
+  VideoQualityModes,
+} from '@discordeno/types'
 import { calculatePermissions, type Bot } from '../index.js'
 import { Permissions } from './toggles/Permissions.js'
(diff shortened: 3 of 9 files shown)


packages/bot/src/transformers/sku.ts
```diff
@@ -1,4 +1,4 @@
-import type { DiscordSku, DiscordSkuFlag, DiscordSkuType } from '@discordeno/types'
+import type { DiscordSku, DiscordSkuType, SkuFlag } from '@discordeno/types'
```
lts20050703, line 1: Inconsistency here

lts20050703 (changes requested):
I don't think you understand how flags work...

You can't just change everything into the Flags enum because how flags work is that it's like a permission bitfield, with each bit can be independently changed to 1 or 0.

For example, a user can be both STAFF (`1 << 0`) and PARTNER (`1 << 1`), which result in `11` binary or `3` decimal.

I think we should keep the property as number and just export the enum so the end user can do some bitwise operations, similar to how users "construct" gateway intents using `|` with the `Intents` enum

Fleny113:
> I don't think you understand how flags work...
> 
> You can't just change everything into the Flags enum because how flags work is that it's like a permission bitfield, with each bit can be independently changed to 1 or 0.
> 
> For example, a user can be both STAFF (`1 << 0`) and PARTNER (`1 << 1`), which result in `11` binary or `3` decimal.
> 
> I think we should keep the property as number and just export the enum so the end user can do some bitwise operations, similar to how users "construct" gateway intents using `|` with the `Intents` enum

Uh?? No? Typescript will resolve the type to the enum itself. The OR operator returns the enum not something else

[playground link](https://www.typescriptlang.org/play/?#code/KYOwrgtgBAYgNgQwOZQN4CgpQILagXigEYoAeUqABgBpMoAheg4sim9AX3XQGMB7EAGcALlAAmALliIUheMgB0uKAB9pixkA)

packages/bot/src/transformers/sku.ts
Fleny113: The `Discord`? If so, there is a question but you ignored it from what i can tell. For now i removed it bc other enums are this way (`UserFlags` or `ChannelFlags`)

lts20050703:
Sorry I guess I wasn't cleared what I meant

https://www.typescriptlang.org/play/?ts=5.4.5#code/KYOwrgtgBAqgzsATgMQDYEMDmcoG8BQURUAygC7oBmlUAvFAIxQA8zUADADSHEAK6iMiCR1GLNg3wBffPjIBPAA7BYCRKILEolDNgBcqpGixxpsgMYB7EHDJQwag-BH1NxHSYMBmaUA

In this example, Discord return a user that is both a staff and partner, so it will return `11` binary or `3` decimal

I think at minimum all we need to do is to export the Flags enum. Then the user can import those flags enum, and do some bitwise operation to check if the user has that flag. For example `if (user.flags & UserFlags.Staff) {}`

Alternatively we can do something like the whole Permissions thing, something like this https://github.com/discordeno/discordeno/blob/main/packages/utils/src/permissions.ts

Fleny113:
Then the issue is another one, number is not compatible with UserFlags because the enum is a narrow type compared to number. For this you can just tell TS to shutup and say as you want with a  as

Skillz4Killz:
Im okay with the name change if it adds consistency but that should be separate pr. For the concept of this PR of changingt he type number to an enum i agree with @lts20050703 that isnt doable with how the api works regarding flags.

Issue #3703, opened 2024-07-03
[Bug]: `guildEmojisUpdate`'s `emojis` field is of type `Collection<bigint, DiscordEmoji>`


tsClauz:
### Runtime

NodeJS

### Runtime Version

22.3.0

### Version

19.0.0-next.ad7e74c

### Describe the bug

Discordeno's event `guildEmojisUpdate` takes as argument an object containing `guildId` and `emojis`. The `emojis` property is a `Collection` of `DiscordEmojis` (`Collection<bigint, DiscordEmoji>`).

### What should've happened?

The `emojis` property should be of type `Collection<bigint, Emoji>`, as the emojis should be passed through the transformer before reaching the event.

### Code to reproduce the bug

```typescript
// This issue does not refer to a runtime bug.
```

PR #3704, opened 2024-07-03
fix(bot)!: Use `bot.transformers.emoji` in `handleGuildEmojisUpdate`

Fleny113:
Currently we use `Collection<bigint, DiscordEmoji>` for the `guildEmojisUpdate` event, however there is not reason (afaik) to use `DiscordEmoji` instead of the transformed object `Emoji`

Fixes: #3703 

> [!WARNING]
> This is breaking because the `DiscordEmoji` and the `Emoji` type are incompatible, this may result in Typescript type errors


lts20050703 (approved):
LGTM!

Skillz4Killz (approved):

Issue #3702, opened 2024-07-03
[Bug]: `guildStickersUpdate` missing field


tsClauz:
### Runtime

NodeJS

### Runtime Version

22.3.0

### Version

19.0.0-next.ad7e74c

### Describe the bug

Discordeno's event `guildStickersUpdate` only takes one argument which is the array of guild stickers. This is a problem because the gateway `GUILD_STICKERS_UPDATE`'s `guild_id` doesn't seem to be provided at all and it becomes impossible to determine which guild the event refers to if the array of stickers is empty.

### What should've happened?

The event should take an object which includes both `guildId` and `stickers`.

### Code to reproduce the bug

```typescript
// This issue does not refer to a runtime bug.
```

PR #3706, opened 2024-07-03
fix(bot)!: Add `guildId` to `guildStickersUpdate` event

Fleny113:
Currently there is no way to access the `guildId` on the `guildStickersUpdate` event if the events sends an empty array of stickers

Fixes: #3702 

> [!WARNING]
> This is a breaking change since it changes the argument to `guildStickersUpdate` to be a payload object with the `guildId` and the stickers as proprieties on it


lts20050703 (approved):
LGTM!

Skillz4Killz (approved):

PR #3709, opened 2024-07-05
fix(utils): Use '%20' instead of ' ' for joining scopes

Fleny113:
Currently the `createOAuth2Link` function uses a space character to join the scopes provided, this does work, however it creates an issue where if you don't pass the resulting URL in something like the `URL` constructor the space can give issues when using. For example, it makes discord error when used as the url for a link button


lts20050703 (approved):
LGTM!

Skillz4Killz (approved):

PR #3726, opened 2024-07-08
feat(types)!: Use BigInt for permission flags

Fleny113:
Discord is pretty close to the Number.MAX_SAFE_INTEGER (around 1 << 53, with 1 << 55 it will start truncating numbers), to avoid this we need to change to BigInt.

This pr also removes some BigInt constructor calls that are no longer necessary.

The reason this PR changes the enum to an object is because Typescript does not support bigint keys in objects, and enums are transformed into objects with also the values as keys. There are issues in the typescript repo, however they date back version 2.6

> [!WARNING]
> This is a breaking change for 2 reason:
> 1. BigInt is not compatibile with number, so if you used number it breaks
> 1. Since we needed to convert to an object now if you do `BitwisePermissionFlags[1]` it now gives undefined instead of the name of the permission

### DIFF
--- a/packages/types/src/shared.ts
+++ b/packages/types/src/shared.ts
@@ -649,106 +649,106 @@ export enum ApplicationCommandPermissionTypes {
 }
 
 /** https://discord.com/developers/docs/topics/permissions#permissions-bitwise-permission-flags */
-export enum BitwisePermissionFlags {
+export const BitwisePermissionFlags = {
   /** Allows creation of instant invites */
-  CREATE_INSTANT_INVITE = 0x0000000000000001,
+  CREATE_INSTANT_INVITE: 1n << 0n,
   /** Allows kicking members */
-  KICK_MEMBERS = 0x0000000000000002,
+  KICK_MEMBERS: 1n << 1n,
   /** Allows banning members */
-  BAN_MEMBERS = 0x0000000000000004,
+  BAN_MEMBERS: 1n << 2n,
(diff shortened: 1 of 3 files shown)


Fleny113:
> Why `export enum BitwisePermissionFlags` -> `export const BitwisePermissionFlags`?

(oh wait, i forgot to write it in the pr lol)

TS does not like BigInt + Enums. There are issues on the typescript repo that date back to 3.6 dev and 2.6 dev

For this reason we need to use an object, however i'm not going to do the double mapping TSC usually does (there passing the enum value would get you the enum key)

packages/types/src/shared.ts
```diff
@@ -649,106 +649,106 @@ export enum ApplicationCommandPermissionTypes {
 }
 
 /** https://discord.com/developers/docs/topics/permissions#permissions-bitwise-permission-flags */
-export enum BitwisePermissionFlags {
+export const BitwisePermissionFlags = {
```
lts20050703, line 652: Actually now that we switched to a const, why not this to make sure that next time Discord add a new permission string we don't forget to update this or the PermissionStrings type?
```suggestion
export const BitwisePermissionFlags: Record<PermissionStrings, bigint> = {
```

Fleny113: 1. the as const is useless at that point
2. PermissionStrings is `keyof typeof BitwisePermissionFlags`, so it becomes a dependency loop in the type checker

Skillz4Killz (approved):

PR #3733, opened 2024-07-10
fix(website): fix typedoc configuration

Fleny113:
Update `typedoc-markdown-plugin` to fix CI errors.

Adjust the website/`typedoc.json`/`finalizeTypedocs.js` to adjust for the new typedoc release.

This pr includes changes to the READMEs and some jsdoc comments to fix MDX invalid syntaxes that `typedoc-markdown-plugin` didn't sanitize


website/api_reference/generated/.nojekyll
lts20050703, line 1: Do we need to set the `githubPages` option to `false`?

lts20050703 (approved):
LGTM!

website/api_reference/generated/.nojekyll
Fleny113: i didn't see typedoc generate the file anymore, so i think it is no longer an option? tbf i did not check

Fleny113: According to the typedoc docs:
> When enabled, automatically add a .nojekyll file to the output directory to prevent GitHub Pages from processing your documentation site using Jekyll. If you have scoped packages, TypeDoc generates HTML files that start with _ which are ignored by Jekyll. Defaults to true.

However i didn't see the file being generated ?? so I'm a bit confused

Anyway i don't think this is an issues because we use github pages from CI, not from the static files

Skillz4Killz (approved):

PR #3735, opened 2024-07-10
build(deps-dev): Update turbo to v2

Fleny113:
Update turbo to v2. The difference between this and dependabot pr is that this pr was made using `@turbo/codemod` to also migrate the turbo.json file.

This pr also includes some configuration changes
- Remove `--no-daemon`, let turbo decide may be a better idea after all
- Change `release-build` to improve caching. Currently the `release-build` is a script that will build first with SWC then TSC, sequentially for each package, when SWC could be changed to be all in parallel, for this reason now the script (in the root, the one that uses turbo) now runs in parallel `build` and `build:type` and leaves turbo to deal with all the stuff it can parallelize, also this improves caching as running `release-build` will cache for `build` and `build:type` as well.
- Change `build:watch` to use turbo so that it can be cached more efficently as noted above for `release-build`
- Removed the dev script, it didn't work
- The `fmt` script no longer require `build:type` as a dependency (why it did? lint needs it for type aware linting, but fmt?)


lts20050703 (approved):

Skillz4Killz (dismissed):

packages/bot/src/transformers/toggles/member.ts
lts20050703, line 1: Unrelated change?

lts20050703 (approved):
LGTM!

packages/bot/src/transformers/toggles/member.ts
Fleny113: provably a change due to the switch frim eslint to biome that didn't get formatted in the pr

PR #3736, opened 2024-07-11
fix(bot): Add missing proprieties on `Channel`

Fleny113:
Currently a bunch of proprieties are missing on the `Channel` type. This PR adds them


Fleny113:
> Unrelated to this PR: Is there any reason for `Channel extends BaseChannel`?

Discord doesn't have these `Base<type>` types. This is something we do for typings on the getters, so it includes stuff that will be in the `base<type>` object there is in the transformer file

Skillz4Killz (approved):

PR #3737, opened 2024-07-11
fix(docker): Fix invalid snyk serif

Fleny113:
Add step in the CI to prevent an invalid serif file to be uploaded to the GitHub Security tab

(didn't test as it is push only and locally with `act` i wasn't able to get the action running)


lts20050703 (approved):
LGTM!

Skillz4Killz (approved):

PR #3740, opened 2024-07-11
fix(gateway): Log the `code.code` more often, incluing in the generic message

Fleny113:
Currently we only log the close code for the shard if the error requires re-identifying (as specified by Discord), however the `close.code` can be useful in other situations as well, so now the debug logs always include them. Also they are now included in the "Closed shard #n" info log since it run on the `default` switch case where knowing the error code may be useful


AwesomeStickz (approved):

TriForMine (approved):

lts20050703 (approved):
LGTM!

Skillz4Killz (approved):

PR #3741, opened 2024-07-12
fix(bot)!: Rename `channel.autoArchiveDuration` to `defaultAutoArchiveDuration`

Fleny113:
Currently we name the discord propriety `default_auto_archive_duration` on `Channel` as `autoArchiveDuration`. This may be confusing for users but this PR renames the propriety as `defaultAutoArchiveDuration`.

This also adds the desired propriety for `defaultAutoArchiveDuration`, we don't remove the previous one as that is used for the thread metadata


lts20050703 (approved):
LGTM!

PR #3745, opened 2024-07-13
fix: adapt biome formatting of package.json files to match the yarn one

Fleny113:
Currently yarn and biome fight every time you install the dependencies/run the formatter.

The PRs adds the package.json to the ignore list for biome.

> [!NOTE]
> This PR includes a change to the import sorting for the member toggles as #3617 got merged while using ESLint and it didn't get updated to use biome

### DIFF
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -41,16 +41,10 @@
       "allowTrailingCommas": true
     }
   },
-  "overrides": [
-    {
-      "include": ["package.json"],
-      "formatter": {
-        "lineWidth": 1
-      }
-    }
-  ],
   "files": {
     "ignore": [
+      // Ignore the package.json and leave yarn to format it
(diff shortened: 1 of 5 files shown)


lts20050703 (approved):
LGTM! But is there no way to ignore the `package.json` file like `.prettierignore` or something? If there is why not just tell Biome to ignore the `package.json` file?

Fleny113:
> LGTM! But is there no way to ignore the `package.json` file like `.prettierignore` or something? If there is why not just tell Biome to ignore the `package.json` file?

We can put it in the ignore list so biome ignores it

lts20050703:
Yeah I feel like telling Biome to ignore files auto-generated by other tools is better in this case

lts20050703 (approved):
LGTM!

PR #3758, opened 2024-07-16
feat(types): Add banner to `DiscordPartialGuild`

### LINKED ISSUE #3755: [api-docs] Add banner field to Get Current User Guilds (#6982)
A new commit was made into the api-docs repo: https://github.com/discord/discord-api-docs/commit/28beb3020d37ce41c0c0645820ca0e70597a85ef

This is a bot created issue.

Fleny113:
Add banner to `DiscordPartialGuild`

On a side note: Why do we even have this type? wouldn't it be the same as using `Partial<DiscordGuild>`? It would also be a better idea to do so because it doesn't seem that discord documents explicitly what are the props that this method will return

- upstream: https://github.com/discord/discord-api-docs/pull/6982
- fixes #3755

### DIFF
--- a/packages/types/src/discord.ts
+++ b/packages/types/src/discord.ts
@@ -913,6 +913,8 @@ export interface DiscordPartialGuild {
   id: string
   /** Icon hash */
   icon: string | null
+  /** Banner hash */
+  banner: string | null
   /** true if the user is the owner of the guild */
   owner: boolean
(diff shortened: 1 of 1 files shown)


lts20050703 (approved):
LGTM! Though I agree that it would be better if we use Partial<DiscordGuild>

Fleny113:
closed in favor of #3768

PR #3760, opened 2024-07-16
ci: Update actions versions

Fleny113:
Update GitHub actions versions for especially `actions/cache` and the artifacts related ones

`actions/cache@v3` uses node16 which gets force-replaced by github with node20 (as node16 is EOL) and it gives quite a few warnings in the workflow summary

It also seems to solve the issue we had with very long time to write the cache for turborepo


lts20050703 (approved):
LGTM!

PR #3761, opened 2024-07-16
fix(rest): Fix followups from getting stuck in queue

Fleny113:
Currently it is possibile for the followups to get stuck in an dead queue (that will never refresh) because of the omission of both the header and the token in the url (due to how the ratelimit buckets work)

### DIFF
--- a/packages/rest/src/manager.ts
+++ b/packages/rest/src/manager.ts
@@ -102,8 +102,8 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
       }
     },
 
-    checkRateLimits(url, requestAuthorization) {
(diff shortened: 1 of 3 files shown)


Fleny113:
For the ones that want to test the pr i have built the packages and put the `.tgz` inside the zip (github doesn't support `.tgz` files so i needed to do a .zip)

[discordeno-packages.zip](https://github.com/user-attachments/files/16247747/discordeno-packages.zip)

Fleny113:
I have changed this to a draft as when in a proxy the value isn't passed to the proxy so it might cause issues there, so while i fix i have marked the pr as draft

PR #3786, opened 2024-07-22
faet(bot): Type `commandOptionsParser`, remove any

Fleny113:
Currently we return `Record<string, any>` from `commandOptionsParser`, this can be problematic as it is not obvious that for `ApplicationCommandOptionTypes.User` and for `ApplicationCommandOptionTypes.Mentionable` (when a user is provided) we return `{ user: User, member: Member }`, this now changes the type to make it more clear.

This fixes the types for the `InteractionDataResolved.channels` and `InteractionDataResolved.members` are they were incorrect

### DIFF
--- a/packages/bot/src/commandOptionsParser.ts
+++ b/packages/bot/src/commandOptionsParser.ts
@@ -1,11 +1,11 @@
 import { ApplicationCommandOptionTypes } from '@discordeno/types'
-import type { Interaction, InteractionDataOption } from './index.js'
+import type { Attachment, Channel, Interaction, InteractionDataOption, Member, Role, User } from './index.js'
 
-export function commandOptionsParser(interaction: Interaction, options?: InteractionDataOption[]): Record<string, any> {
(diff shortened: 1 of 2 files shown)


lts20050703 (approved):
LGTM!

packages/bot/src/typings.ts
```diff
@@ -130,7 +130,7 @@ export interface DiscordInteractionDataResolved {
   /** The Ids and Role objects */
   roles?: Record<string, DiscordRole>
   /** The Ids and partial Channel objects */
-  channels?: Record<string, Pick<DiscordChannel, 'id' | 'name' | 'type' | 'permissions'>>
+  channels?: Record<string, Pick<DiscordChannel, 'id' | 'name' | 'type' | 'permissions' | 'thread_metadata' | 'parent_id'>>
```
lts20050703, line 133: Leaving this for future reference:
`** Partial Channel objects only have id, name, type and permissions fields. Threads will also have thread_metadata and parent_id fields.`
https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-object-resolved-data-structure

Skillz4Killz (approved):

PR #3787, opened 2024-07-22
feat(bot): Add `desiredProperties` to createBot

Fleny113:
Add `desiredProprieties` to `createBot` and make the transformers object in `createTransformers` partial. This allows a user to provide an object with some of the props they want to be included in the transformed objects


lts20050703 (approved):
LGTM!

PR #3792, opened 2024-07-23
fix(bot,gateway): Fix (some) resharding bugs 

Fleny113:
Currently the resharding
- Will throw an error if you don't provide a `getSessionInfo`, while we can provide a fallback value if you use `createBot` from `@discordeno/bot`
- Will attempt to reshard regardless of the % of capacity (the formula is bugged)
- Will make `gateway.resharding.shards` and `gateway.shards` be the same collections, create issues at a later point

TODO:
- [x] Find a way to get the correct capacity% shards

(@AwesomeStickz)

### DIFF
--- a/packages/gateway/src/manager.ts
+++ b/packages/gateway/src/manager.ts
@@ -61,11 +61,9 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
       shards: new Collection(),
       pendingShards: new Collection(),
       async getSessionInfo() {
-        gateway.logger.error(
+        throw new Error(
           '[Resharding] was enabled but no getSessionInfo handler was provided. Please set a handler like: gateway.resharding.getSessionInfo = async () => { // insert code here to fetch getSessionInfo from rest process. }',
         )
-
-        return null
       },
       async checkIfReshardingIsNeeded() {
         if (!gateway.resharding.enabled) {
@@ -692,13 +690,8 @@ export interface GatewayManager extends Required<CreateGatewayManagerOptions> {
     shards: Collection<number, Shard>
     /** Holds the pending shards that have been created and are pending all shards finish loading. */
     pendingShards: Collection<number, Shard>
-    /**
-     * Handler to get shard count and other session info.
-     *
-     * @remarks
-     * Should never return null. Returning null will skip the resharding
-     */
-    getSessionInfo: () => Promise<Camelize<DiscordGetGatewayBot> | null>
+    /** Handler to get shard count and other session info. */
+    getSessionInfo: () => Promise<Camelize<DiscordGetGatewayBot>>
     /** Handler to edit the shard id on any cached guilds. */
     updateGuildsShardId: (guildIds: string[], shardId: number) => Promise<void>
     /** Handler to check if resharding is necessary. */
--- a/packages/gateway/src/manager.ts
+++ b/packages/gateway/src/manager.ts
@@ -119,12 +119,6 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
         if (typeof info.lastShardId === 'number') gateway.lastShardId = info.lastShardId
         gateway.logger.warn(`[Resharding] Starting the reshard process. New Total Shards. ${gateway.totalShards}`)
 
-        if (oldShardNumber === gateway.totalShards) {
-          gateway.logger.warn('[Resharding] Aborting the reshard process. The total number of shards is the same')
-
-          return
-        }
-
         // Resetting buckets
         gateway.buckets.clear()
         // Refilling buckets with new values
--- a/packages/gateway/src/manager.ts
+++ b/packages/gateway/src/manager.ts
@@ -230,10 +230,7 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
         gateway.logger.warn(`[Resharding] Completed.`)
 
         // Replace old shards
-        gateway.shards.clear()
-        for (const shard of gateway.resharding.shards.values()) {
-          gateway.shards.set(shard.id, shard)
-        }
+        gateway.shards = new Collection(gateway.resharding.shards)
 
         // Clear our collections and keep only one reference to the shards, the one in gateway.shards
         gateway.resharding.shards.clear()
--- a/packages/gateway/src/manager.ts
+++ b/packages/gateway/src/manager.ts
@@ -198,7 +198,7 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
         gateway.logger.warn(`[Resharding] Shard #${shard.id} is now pending`)
 
         // Check if all shards are now online.
-        if (gateway.resharding.shards.size !== gateway.resharding.pendingShards.size) return
+        if (gateway.lastShardId - gateway.firstShardId >= gateway.resharding.pendingShards.size) return
 
         gateway.logger.warn(`[Resharding] All shards are now online.`)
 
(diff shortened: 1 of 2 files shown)


packages/gateway/src/manager.ts
```diff
@@ -99,6 +119,12 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
         if (typeof info.lastShardId === 'number') gateway.lastShardId = info.lastShardId
         gateway.logger.warn(`[Resharding] Starting the reshard process. New Total Shards. ${gateway.totalShards}`)
 
+        if (oldShardNumber === gateway.totalShards) {
+          gateway.logger.warn('[Resharding] Aborting the reshard process. The total number of shards is the same')
+
+          return
+        }
+
```
AwesomeStickz, line 127:
```suggestion

```

AwesomeStickz: This is not needed and shouldn't be included so you could use resharding as a way to change your gateway versions in gateway proxy (this was supposed to be one of the usecases even tho it might not be too easy to do)

packages/gateway/src/manager.ts
```diff
@@ -66,28 +66,48 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
...
+          return { needed: false, info: sessionInfo }
+        }
+
+        gateway.logger.warn('[Resharding] Able to reshard, checking whether necessary now.')
 
+        // FIXME: this calculation is clearly wrong to some degree, using 2 shards and 5000 guilds the % should be 100, but is 250
         // 2500 is the max amount of guilds a single shard can handle
         // 1000 is the amount of guilds discord uses to determine how many shards to recommend.
         // This algo helps check if your bot has grown enough to reshard.
         const percentage = ((2500 * sessionInfo.shards) / (gateway.totalShards * 1000)) * 100
```
AwesomeStickz, line 96:
```suggestion
        const percentage = (gateway.totalShards / sessionInfo.shards) * 100
```

packages/gateway/src/manager.ts
Fleny113: > as a way to change your gateway versions in gateway proxy

might be me: but that does like a dumb use-case to me lol, aside from that, fine ig?

packages/gateway/src/manager.ts
```diff
@@ -66,24 +66,43 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
...
-        // Resharding is disabled.
-        if (!gateway.resharding.enabled) return { needed: false }
-        gateway.logger.warn(`[Resharding] Resharding is enabled.`)
+        if (!gateway.resharding.enabled) {
+          gateway.logger.debug('[Resharding] Resharding is disabled.')
+
+          return { needed: false }
+        }
+
+        gateway.logger.warn('[Resharding] Checking if resharding is needed.')
```
AwesomeStickz, line 75:
```suggestion
        gateway.logger.warn('[Resharding] Checking if resharding is needed.')

        if (!gateway.resharding.enabled) {
          gateway.logger.debug('[Resharding] Resharding is disabled.')

          return { needed: false }
        }

```

packages/gateway/src/manager.ts
```diff
@@ -66,24 +66,43 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
...
+        gateway.logger.warn('[Resharding] Resharding is enabled.')
 
         const sessionInfo = await gateway.resharding.getSessionInfo()
-        gateway.logger.warn(`[Resharding] Session info retrieved.`)
+
+        gateway.logger.warn('[Resharding] Session info retrieved.')
+        gateway.logger.debug(`[Resharding] Session info details: ${JSON.stringify(sessionInfo)}`)
```
AwesomeStickz, line 81:
```suggestion
        gateway.logger.debug(`[Resharding] Session info retrieved: ${JSON.stringify(sessionInfo)}`)
```

packages/gateway/src/manager.ts
```diff
@@ -66,24 +66,44 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
...
+          return { needed: false, info: sessionInfo }
+        }
+
+        gateway.logger.warn('[Resharding] Able to reshard, checking whether necessary now.')
 
         // 2500 is the max amount of guilds a single shard can handle
         // 1000 is the amount of guilds discord uses to determine how many shards to recommend.
         // This algo helps check if your bot has grown enough to reshard.
-        const percentage = ((2500 * sessionInfo.shards) / (gateway.totalShards * 1000)) * 100
+        // While this is imprecise as discord changes the recommended number of shard every 1000 guilds it is good enough
```
AwesomeStickz, line 95: Mention this in the comment of `shardsFullPercentage` so user can know that it is not 100% accurate

packages/gateway/src/manager.ts
```diff
@@ -200,20 +218,24 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
...
+              if (message.t === 'GUILD_MEMBERS_CHUNK') {
+                oldHandler?.(shard, message)
+              }
             },
           }
         }
 
         gateway.logger.warn(`[Resharding] Shutting down old shards.`)
-        // Close old shards
         await gateway.shutdown(ShardSocketCloseCodes.Resharded, 'Resharded!', false)
```
AwesomeStickz, line 229:
```suggestion
        // Close old shards
        await gateway.shutdown(ShardSocketCloseCodes.Resharded, 'Resharded!', false)
```

packages/gateway/src/manager.ts
Fleny113: https://github.com/discordeno/discordeno/pull/3792/commits/21592257219cd5e90b93c0dee8be6c2b52ca5444
maybe it is a bit too verbose? idk tell me if you think that it is

lts20050703 (approved):
LGTM!

packages/gateway/src/manager.ts
```diff
@@ -170,24 +191,21 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
...
+
         // Check if all shards are now online.
-        if (gateway.lastShardId - gateway.firstShardId + 1 > gateway.resharding.pendingShards.size) return
+        if (gateway.lastShardId - gateway.firstShardId >= gateway.resharding.pendingShards.size) return
+
         gateway.logger.warn(`[Resharding] All shards are now online.`)
 
         // New shards start processing events
-        for (const shard of gateway.resharding.pendingShards.values()) {
+        for (const shard of gateway.resharding.shards.values()) {
```
lts20050703, line 206: @Fleny113 do we need to undo this change too? You said its not instantly populated

lts20050703 (changes requested):

packages/gateway/src/manager.ts
Fleny113: no, it isn't instantly populated, but by the time this line runs they are the same

What i meant in commit https://github.com/discordeno/discordeno/pull/3792/commits/59dcf503b6b93eacaf57d4a2be36a8741dfb6628 is that in the if above that prevents stuff like this for-of to run when that runs `shards` is not fully populated, but if the if check has passed then they are the same and this doesn't need to be reverted, it is just easier (imo) to understand with the iteration on the `shards` collection rather than the `pendingShards` collection

lts20050703 (approved):
LGTM!

AwesomeStickz (approved):

PR #3793, opened 2024-07-23
fix(gateway): Make heartbeat logging clearer

Fleny113:
Currently our debug logs in `startHeartbeating` consists of `[Gateway] Start Heartbeating Shard #<id>` and then a letter based on the order of the log call.

This pr replaces these logs with more clearer versions of these logs to help with the understanding of these logs without having a direct reference on the code on the side as remembering what each letter means is really difficult.

Also, this removes a few log calls as they were right before/after another one with no real use for both of them

### DIFF
--- a/packages/gateway/src/Shard.ts
+++ b/packages/gateway/src/Shard.ts
@@ -226,7 +226,7 @@ export class DiscordenoShard {
       },
       true,
     )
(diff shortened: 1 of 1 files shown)


lts20050703 (dismissed):
LGTM!

lts20050703 (approved):

AwesomeStickz (approved):

PR #3794, opened 2024-07-23
fix(bot)!: Remove old unprovided params from event params

Fleny113:
Currently our `bot.events.messageUpdate` and `bot.events.presenceUpdate` accepts 2 parameters, the updated message/presence and the old message/presence, however discord doesn't give the old message/presence, so this removes it as it is never called with that second parameter set to something

> [!WARNING]
> This is breaking because it is changing the `messageUpdate` and `presenceUpdate` events function signature

### DIFF
--- a/packages/bot/src/bot.ts
+++ b/packages/bot/src/bot.ts
@@ -221,7 +221,7 @@ export interface EventHandlers {
   messageCreate: (message: Message) => unknown
   messageDelete: (payload: { id: bigint; channelId: bigint; guildId?: bigint }, message?: Message) => unknown
   messageDeleteBulk: (payload: { ids: bigint[]; channelId: bigint; guildId?: bigint }) => unknown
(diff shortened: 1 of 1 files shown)


lts20050703 (approved):
LGTM! How did the method was typed that way lol

Fleny113:
> How did the method was typed that way lol

My guess is from v18 and the cache plugin (??)

AwesomeStickz:
Can you change `presenceUpdate: (presence: PresenceUpdate, oldPresence?: PresenceUpdate) => unknown` to `presenceUpdate: (presence: PresenceUpdate) => unknown` as well as it doesn't seem to [send that](https://discord.com/developers/docs/topics/gateway-events#presence-update) too, and I can't suggest that directly in files tab

AwesomeStickz (approved):

PR #3795, opened 2024-07-23
fix(rest): ignore snakelized properties in `changeToDiscordFormat`

### LINKED ISSUE #3666: [Bug]: REST `convertToDiscordFormat` incorrectly snakelizing with localization fields
### Runtime

NodeJS

### Runtime Version

22.2.0

### Version

19.0.0-next.92bf166

### Describe the bug

When using a REST proxy, we run `rest.createRequestBody` to set as the body for the `fetch` call to the rest proxy, if i had to guess this is because when proxying to something that will not do the snakelizing in the proxy code it will fail the request as Discord does not accept camelCase. 

However this causes an issue when we proxy to a Discordeno proxy, a proxy executes `rest.makeRequests` and when it reaches the `sendRequest` method it will run again `rest.createRequestBody`, however while `createRequestBody` has some check to avoid issues with some special fields that need to be threather in a special way (eg. calculating the permision bitfield, not snakelizing the object), but the method checks for the keys in camelCase form and it works fine, until the rest proxy gets the already snakelized object, so the keys are snake_case and not camelCase and those checks get compleatly ignored and the method proceds to snakelize stuff it shouldn't, for example the localized keys, so in the end you obtain `{ "en-_u_s": "...." }` instead of `{ "en-US": "...." }` resuling in a discord API error

### What should've happened?

The `createRequestBody` method should not snakelize the localization fields on the REST proxy, resulting in discord not erroring out due to the unknown locale

### Code to reproduce the bug

Assuming localhost:5000 has the REST proxy docker image that lives in [`docker-apps/rest-passthrough`](https://github.com/discordeno/discordeno/tree/main/docker-apps/rest-passthrough) and the authentication is in a ENV called `REST_AUTH`

```typescript
import { createRestManager } from '@discordeno/rest'

const rest = createRestManager({
  token: process.env.TOKEN,
  proxy: {
    baseUrl: "http://localhost:5000",
    authorization: process.env.REST_AUTH!,
  },
})

await rest.upsertGlobalApplicationCommands([{ name: "a", description: "b", nameLocalizations: { "en-US": "c" } }])
```

Fleny113:
Currently for `nameLocalizations`, `descriptionLocalizations` and some others, we have a special behavior. In the specific case of `nameLocalizations` and `descriptionLocalizations` that will skip the snakelization as that would break the format discord expects.

However, when the rest receives the objects already in the snake_case form (eg in a proxy) it will actually fail the check for `nameLocalizations` and `descriptionLocalizations` breaking the format that discord expects resulting in an error on the request

fixes #3666

### DIFF
--- a/packages/rest/src/manager.ts
+++ b/packages/rest/src/manager.ts
@@ -204,12 +204,15 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
               case 'deny':
                 newObj[key] = typeof value === 'string' ? value : calculateBits(value)
                 continue
+              case 'default_member_permissions':
               case 'defaultMemberPermissions':
                 newObj.default_member_permissions = typeof value === 'string' ? value : calculateBits(value)
                 continue
+              case 'name_localizations':
               case 'nameLocalizations':
                 newObj.name_localizations = value
                 continue
+              case 'name_localizations':
               case 'descriptionLocalizations':
                 newObj.description_localizations = value
                 continue
--- a/packages/rest/src/manager.ts
+++ b/packages/rest/src/manager.ts
@@ -212,7 +212,7 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
               case 'nameLocalizations':
                 newObj.name_localizations = value
                 continue
-              case 'name_localizations':
+              case 'description_localizations':
               case 'descriptionLocalizations':
                 newObj.description_localizations = value
                 continue


lts20050703 (dismissed):
LGTM!

packages/rest/src/manager.ts
```diff
@@ -204,12 +204,15 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
...
                 continue
+              case 'default_member_permissions':
               case 'defaultMemberPermissions':
                 newObj.default_member_permissions = typeof value === 'string' ? value : calculateBits(value)
                 continue
+              case 'name_localizations':
               case 'nameLocalizations':
                 newObj.name_localizations = value
                 continue
+              case 'name_localizations':
```
AwesomeStickz, line 215:
```suggestion
              case 'description_localizations':
```

AwesomeStickz (changes requested):

AwesomeStickz:
This or should we just do this?

```js
if (key.includes('_')) continue
```

packages/rest/src/manager.ts
Fleny113: whoops

Fleny113:
> This or should we just do this?
> 
> ```js
> if (key.includes('_')) continue
> ```

the issue is that this method will be called twice with a proxy, in the bot code and in the proxy, so this would prevent both from snakalize the key, wouldn't it?

AwesomeStickz:
> so this would prevent both from snakalize the key, wouldn't it?

Isn't the purpose to change camelized to snakelized version?

PR #3806, opened 2024-07-26
fix(gateway): Clear `offlineSendQueue` after resolving all the promises

Fleny113:
Currently when we call `send` in a shard we check if the shard is online, and if it isn't the payload is queued for when it is online, however we do not clear the promises resolve callbacks, keeping iterating on useless callback that won't do anything as the promise already resolved.

Also, we currently use the `.map` method that will return a new array with the results resulting in a waste of memory as those arrays are also never used, so `.forEach` is a better method to use for this use-case


lts20050703 (approved):
LGTM!

AwesomeStickz (approved):

PR #3808, opened 2024-07-28
feat(gateway): Use the native WebSocket implementation when available

Fleny113:
Currently we use the built-in WebSocket client implementation only if the runtime being run is Deno, however both Bun and Node (as for v22) have a WebSocket client implementation that we can use. However, after some testing the NodeJS WebSocket seems to be slightly less efficient, so for node we always use the `ws` npm package.

> [!WARNING]
> CI will NOT test this change as we run our Integration Tests with Node v18


lts20050703 (approved):
LGTM!

packages/gateway/src/Shard.ts
```diff
@@ -105,29 +110,29 @@ export class DiscordenoShard {
...
     this.events.connecting?.(this)
 
     const url = new URL(this.connectionUrl)
     url.searchParams.set('v', this.gatewayConfig.version.toString())
     url.searchParams.set('encoding', 'json')
 
-    const socket: NodeWebSocket =
-      // @ts-expect-error Deno
-      globalThis.Deno !== undefined && Reflect.has(globalThis, 'Deno') ? new WebSocket(url.toString()) : new NodeWebSocket(url.toString())
+    // We check for built-in WebSocket implementations before fallback-ing to ws as Deno and Bun have a WebSocket implementation and NodeJS as for v22 has one as well
```
AwesomeStickz, line 120:
```suggestion
    // We check for built-in WebSocket implementations before falling back to ws, as Deno and Bun have a WebSocket implementation and NodeJS as for v22 has one as well
```

packages/gateway/src/Shard.ts
```diff
@@ -105,29 +110,29 @@ export class DiscordenoShard {
...
+    // @ts-expect-error NodeWebSocket doesn't support "dispatchEvent", and while we don't use it it is required on the "WebSocket" type
```
AwesomeStickz, line 121:
```suggestion
    // @ts-expect-error NodeWebSocket doesn't support "dispatchEvent", and while we don't use it, it is required on the "WebSocket" type
```

packages/gateway/src/Shard.ts
```diff
@@ -105,29 +110,29 @@ export class DiscordenoShard {
...
-    socket.onmessage = async (message: NodeWebSocket.MessageEvent) => await this.handleMessage(message)
+    socket.onerror = (event) => this.handleError(event)
+    socket.onclose = async (closeEvent) => await this.handleClose(closeEvent)
+    socket.onmessage = async (messageEvent) => await this.handleMessage(messageEvent)
 
     return await new Promise((resolve) => {
       socket.onopen = () => {
-        // Only set the shard to `Unidentified` state,
-        // if the connection request does not come from an identify or resume action.
+        // Only set the shard to `Unidentified` state, if the connection request does not come from an identify or resume action.
```
AwesomeStickz, line 131:
```suggestion
        // Only set the shard to `Unidentified` state if the connection request does not come from an identify or resume action.
```

packages/gateway/src/Shard.ts
```diff
@@ -327,6 +335,22 @@ export class DiscordenoShard {
     }
   }
 
+  /** Handle an incoming gateway message. */
+  async handleMessage(message: MessageEvent): Promise<void> {
+    let preProcessMessage = message.data
+
+    // If message compression is enabled,
+    // Discord might send zlib compressed payloads.
```
AwesomeStickz, line 343:
```suggestion
    // If message compression is enabled, Discord might send zlib compressed payloads.
```

AwesomeStickz (approved):

Skillz4Killz:
I'd like to see a benchmark comparison between the current one we use and new built in.

Fleny113:
> I'd like to see a benchmark comparison between the current one we use and new built in.

Do you know any benchmarking tool for a WS client specifically? Aside from the Autobahn test suite to check for incompatibilities I'm not aware of any benchmarks for WS client, only for WS servers which does not apply in this case

Fleny113:
I have run the autobahn test suite on both WS and the nodejs WebSocket (under the name "undici" as that is what it uses, https://github.com/nodejs/undici) and put them in a github pages.

https://fleny113.github.io/ws-undici-autobahn/

To me it seems that ws is faster, however i wouldn't say that it is so much slower the built-in that it is bad to use

lts20050703:
> To me it seems that ws is faster

Although I agree that looking at the benchmark ws is faster than nodejs undici in most cases, I'm not sure which section are relevant to our usecases.

> however i wouldn't say that it is so much slower the built-in that it is bad to use

I'm not sure about this... Discordeno is used by lots of big bot devs so it's still better to use whichever is faster.

@Skillz4Killz Thoughts?

Fleny113:
> Discordeno is used by lots of big bot devs so it's still better to use whichever is faster.

I mean, I do agree, we would need some real-world testing tho, the test from my understanding is just sending compressed/uncompressed binary data for bitmaps (some are json too) over websocket

Fleny113:
I did a few tests, generally the node new WebSocket global uses ~~20MB more and in speeds is about the same, so i have added 2 conditions to check for Deno and Bun so Node still uses the `ws` package

(The test was a server created with `ws` that after a client connected and sent a message it would send at interval of 1 ms all the events from https://github.com/discordeno/benchmarks, the time were pretty much equal, on the order of milliseconds)

lts20050703 (approved):
LGTM!

Issue #3611, opened 2024-05-24
[api-docs] [gateway] Clarify Reconnect opcode, Document zstd-stream (#6877)


DiscordenoBot:
A new commit was made into the api-docs repo: https://github.com/discord/discord-api-docs/commit/549e656768ec7fc1e98670f16b9487c65dc17269
This clarifies that Reconnect can come at any time, even before hello. Also, document the resurrected zstd-stream gateway compression option.

---------

Co-authored-by: advaith <advaithj1@gmail.com>
Co-authored-by: val.le <81811276+valdotle@users.noreply.github.com>

This is a bot created issue.

Fleny113:
We do not support zstd compression on the gateway so this has no effect

AwesomeStickz:
But maybe we should look into it so it stays open?

lts20050703:
> In fact we do have some code to handle zstd compression

We don't have any code to handle z**std**, the code we have handle z**lib** using NodeJS's builtin standary library `node:zlib` module

If we were to suppport z**std** that would probably means yet another dependencies? or maybe a optionalDependencies?

PR #3815, opened 2024-07-29
feat(gateway): Add transport compression support

Fleny113:
Add transport compression support to shards

> [!NOTE]
> This pr depends on some of the changes from #3808

TODO:
- [x] Zlib-stream
- [x] Zstd-stream
- [x] bring back payload compression

for zstd-stream:
- upstream: https://github.com/discord/discord-api-docs/pull/6877
- fixes: #3611


Fleny113:
I am going to wait for #3808 to be merged so that i can adjust the code as needed

lts20050703:
@Fleny113 Is there any reason why this depends on https://github.com/discordeno/discordeno/pull/3808 ?

Fleny113:
> Is there any reason why this depends on #3808 ?

I need the type to be WebSocket instead of the NodeWebsocket as i set a value for WebSocket later (the `binaryType = 'arrayBuffer'`)

Fleny113:
For the ones that wonder why `fzstd` was chosen instead of another zstd library:

- [zstd-codec](https://www.npmjs.com/package/zstd-codec):
   1. After the first message it returns null instead of actually decoding.
   2. If we use `decompressChunks` instead of `decompress` it wants ALL the chunks at once, something impossibile to do while working with the discord gateway
- [zstddec](https://www.npmjs.com/package/zstddec):
   1. Returns a WASM error if we don't specify the size of the uncompressed data, which we do not have access to
- [@oneidentity/zstd-js](https://www.npmjs.com/package/@oneidentity/zstd-js):
   1. If it errors, it prints the entire code in the error as it is minified
   2. Is unable to decode after the first message: `ZSTD_ERROR: Unknown frame descriptor,  error code: -10`
- [@mongodb-js/zstd](https://github.com/mongodb-js/zstd):
   1. Does not support ZSTD streaming, proven by the `Incomplete frame` error
- [simple-zstd](https://www.npmjs.com/package/simple-zstd):
   1. Depends on the system installed `zstd` binary/library
- [fzstd](https://www.npmjs.com/package/fzstd): Works fine

(list from https://facebook.github.io/zstd)

For the code I used to test them:
```js
// import mongo from '@mongodb-js/zstd';
// import oizstd from '@oneidentity/zstd-js';
import { Decompress } from 'fzstd';
// import codec from "zstd-codec";
// import { ZSTDDecoder } from "zstddec";
// const { ZstdInit } = oizstd;
// const { decompress } = mongo;
// const { ZstdCodec } = codec

import { readFile } from "node:fs/promises";

const textDecoder = new TextDecoder();

// const zstddec = new ZSTDDecoder();
// await zstddec.init();

const fzstdDecro = new Decompress((data) => {
    console.log(textDecoder.decode(data));
});

// const { ZstdStream } = await ZstdInit();

// const zstd = await new Promise(resolve => ZstdCodec.run(resolve))
// const stream = new zstd.Streaming()

const file = await readFile("zstd-test-1.json", "utf-8");
const json = JSON.parse(file);

for (const msg of json) {
    const u8 = new Uint8Array(Buffer.from(msg, 'hex'));

    fzstdDecro.push(u8);

    // BROKEN: RuntimeError: memory access out of bounds
    // console.log(zstddec.decode(u8))

    // Has the fun quirk of spitting out ITS ENTIRE F* CODE IN THE STACK TRACE
    // BROKEN: Can't get the frame descriptor after the first message
    // try {
    //     console.log(textDecoder.decode(ZstdStream.decompress(u8)));
    // }
    // catch (e) {
    //     console.log(e.message);
    // }

    // BROKEN: Does not work after the first message
    // console.log(textDecoder.decode(stream.decompress(u8)));
    // console.log(textDecoder.decode(stream.decompressChunks([u8])))

    // BROKEN: incomplete frame
    // console.log(await decompress(Buffer.from(msg, 'hex')))
}
```

where `zstd-test-1.json` is a json with an array (`[ '...', '...' ]`) with HEX encoded Uint8Arrays with messages from the discord gateway in zstd-stream compression

packages/gateway/src/Shard.ts
```diff
@@ -117,20 +139,77 @@ export class DiscordenoShard {
     url.searchParams.set('v', this.gatewayConfig.version.toString())
     url.searchParams.set('encoding', 'json')
 
+    // Set the compress url param and initialize the decompression contexts
+    if (this.gatewayConfig.transportCompression) {
+      url.searchParams.set('compress', this.gatewayConfig.transportCompression)
+
+      if (this.gatewayConfig.transportCompression === TransportCompression.zlib) {
```
lts20050703, line 146: (preference/stylistic, feel free to ignore) maybe switch case? idk I just think it make more sense in this case

packages/gateway/src/Shard.ts
```diff
@@ -339,17 +427,74 @@ export class DiscordenoShard {
...
+    // A buffer is a Uint8Array under the hood. An ArrayBuffer is generic, so we need to create the Uint8Array that uses the whole ArrayBuffer
+    const compressedData: Uint8Array = data instanceof Buffer ? data : new Uint8Array(data)
+
+    if (this.gatewayConfig.transportCompression === TransportCompression.zlib) {
+      if (!this.inflate) {
+        this.logger.fatal('[Shard] zlib-stream transport compression was enabled. But no instance of Inflate was found.')
+        return null
+      }
+
+      // Alias, used to avoid some null checks in the Promise constructor
```
lts20050703, line 456: Why not `this.inflate?.` or `this.inflate!.`?

packages/gateway/src/Shard.ts
```diff
@@ -339,17 +427,74 @@ export class DiscordenoShard {
...
+  /**
+   * Decompress a zlib/zstd compressed packet
+   *
+   * @private
+   */
+  async decompressPacket(data: ArrayBuffer | Buffer): Promise<DiscordGatewayPayload | null> {
+    // A buffer is a Uint8Array under the hood. An ArrayBuffer is generic, so we need to create the Uint8Array that uses the whole ArrayBuffer
+    const compressedData: Uint8Array = data instanceof Buffer ? data : new Uint8Array(data)
+
+    if (this.gatewayConfig.transportCompression === TransportCompression.zlib) {
```
lts20050703, line 450: (preference/stylistic, feel free to ignore) maybe switch case? idk I just think it make more sense in this case

lts20050703 (approved):
LGTM!

packages/gateway/src/Shard.ts
Fleny113: In this case it does not work, it would be a `switch` (for transport compression) + an `if` later (for payload comperssion), might as well keep it an `if`

packages/gateway/src/Shard.ts
Fleny113: ~~`?.` is runtime and `!.` is lying to TS *tecnically*~~

If you think it is clearer fine, ig they are used more often rather than aliasing to avoid these issues

packages/gateway/src/Shard.ts
Fleny113: idk, i tend to avoid 2 cases switch statements. I generally think a few (2-4) if(s) are fine, after that a switch might be a better option

packages/gateway/src/Shard.ts
```diff
@@ -118,20 +140,77 @@ export class DiscordenoShard {
...
+            return
+          }
+
+          this.inflateBuffer = data
+        })
+      }
+
+      if (this.gatewayConfig.transportCompression === TransportCompression.zstd) {
+        const fzstd = await getFZStd().catch(() => {
+          this.logger.warn('[Shard] "fzstd" is not installed. Disabling transport compression.')
```
lts20050703, line 176:
```suggestion
          this.logger.warn('[Shard] "fzstd" is not installed. Disabled transport compression.')
```

packages/gateway/src/Shard.ts
```diff
@@ -118,20 +140,77 @@ export class DiscordenoShard {
...
     socket.onerror = (event) => this.handleError(event)
-    socket.onclose = async (closeEvent) => await this.handleClose(closeEvent)
-    socket.onmessage = async (messageEvent) => await this.handleMessage(messageEvent)
+    socket.onclose = (closeEvent) => this.handleClose(closeEvent)
+    socket.onmessage = (messageEvent) => this.handleMessage(messageEvent)
 
     return await new Promise((resolve) => {
       socket.onopen = () => {
-        // Only set the shard to `Unidentified` state if the connection request does not come from an identify or resume action.
+        // Only set the shard to `Unidentified` state, if the connection request does not come from an identify or resume action.
```
lts20050703, line 213:
```suggestion
        // Only set the shard to `Unidentified` state if the connection request does not come from an identify or resume action.
```

packages/gateway/src/Shard.ts
```diff
@@ -278,7 +357,16 @@ export class DiscordenoShard {
...
-    this.logger.debug(`[Shard] Gateway connection closed with code ${close.code} (${close.reason || '<No reason provided>'}).`)
+
+    // Clear the zlib/zstd data
+    this.inflate = undefined
+    this.zstdDecompress = undefined
+    this.inflateBuffer = null
+    this.decompressionPromisesQueue = []
+
+    this.logger.debug(
+      `[Shard] Gateway connection closed with code ${close.code} (${close.reason || '<No reason provided>'}) (${close.reason || '<No reason provided>'}).`,
```
lts20050703, line 368:
```suggestion
      `[Shard] Gateway connection closed with code ${close.code} (${close.reason || '<No reason provided>'}).`,
```

packages/gateway/src/Shard.ts
```diff
@@ -340,17 +428,74 @@ export class DiscordenoShard {
...
+   *
+   * @private
+   */
+  async decompressPacket(data: ArrayBuffer | Buffer): Promise<DiscordGatewayPayload | null> {
+    // A buffer is a Uint8Array under the hood. An ArrayBuffer is generic, so we need to create the Uint8Array that uses the whole ArrayBuffer
+    const compressedData: Uint8Array = data instanceof Buffer ? data : new Uint8Array(data)
+
+    if (this.gatewayConfig.transportCompression === TransportCompression.zlib) {
+      if (!this.inflate) {
+        this.logger.fatal('[Shard] zlib-stream transport compression was enabled. But no instance of Inflate was found.')
```
lts20050703, line 453:
```suggestion
        this.logger.fatal('[Shard] zlib-stream transport compression was enabled but no instance of Inflate was found.')
```

packages/gateway/src/Shard.ts
```diff
@@ -340,17 +428,74 @@ export class DiscordenoShard {
...
+      this.inflateBuffer = null
+
+      return JSON.parse(decodedData)
     }
 
-    // Safeguard incase decompression failed to make a string.
-    if (typeof preProcessMessage !== 'string') return
+    if (this.gatewayConfig.transportCompression === TransportCompression.zstd) {
+      if (!this.zstdDecompress) {
+        this.logger.fatal('[Shard] zstd-stream transport compression was enabled. But no instance of Decompress was found.')
```
lts20050703, line 481:
```suggestion
        this.logger.fatal('[Shard] zstd-stream transport compression was enabled but no instance of Decompress was found.')
```

lts20050703 (approved):

lts20050703 (approved):

H01001000 (approved):

PR #3820, opened 2024-07-31
fix(bot): Make token in rest and gateway for `createBot` optional

Fleny113:
from #3422:
> - `options.gateway.token` is now also optional since it can be inherited from `options.token`.

While the types have a `Parital` typescript does not actually correctly understand this, and it errors if you don't provide the token in the `rest` and `gateway` nested objects. This pr changes the type to make them actually optional


AwesomeStickz (approved):

lts20050703 (approved):
LGTM!

PR #3823, opened 2024-08-01
fix(bot,types)!: update `DiscordAuditLogChange` to be less manual

Fleny113:
Our current `DiscordAuditLogChange` is pretty much outdated, on both the `types` and `bot` package. This is because while discord has been adding properties, we haven't updated the audit log change to reflect those, as it is very easy to forget about it.

This PR removes completely the `AuditLogChange` type and refactors the `DiscordAuditLogChange` to use the actual objects as source for all the values + some manually done for handling the exceptions discord has for some objects. The reason why I chose this is the following note in the Discord Developer Docs:
> You should assume that your app may run into any field for the changed object, though none are guaranteed to be present. In most cases only a subset of the object's fields will be in the changes array.

So, while discord may never send a field, we should still have it typed in case it does send it.

For the removal of `AuditLogChange`: It is simply a mess to keep up to date, keeping it implies keeping the transformer up to date, and handing 230+ properties in the transformer is very hard to handle even with type errors, especially because the current solution is `.map` with a switch that will make the entire `.map` arrow function error, not just the single case that has an issue.

This pr also makes the values in `OptionalAuditEntryInfo` actually optional even if the type tecnically doesn't define them as such, however they will never exist all at the same time, so unless we create a discriminated union for every single case it is probably better to have the user null-check what they are going to use. Also, this adds the code for the transformation of `application_id` which for some reason was omitted, not sure exacly why, maybe the fact that it is duplicated? however, I would expect the reverse transformer to not just omit the `application_id` because it can...

> [!WARNING]
> This is breaking as removing the transformed type is going to break some code

### DIFF
--- a/packages/bot/src/transformers/auditLogEntry.ts
+++ b/packages/bot/src/transformers/auditLogEntry.ts
@@ -1,133 +1,26 @@
 import type { DiscordAuditLogEntry } from '@discordeno/types'
-import { type AuditLogEntry, type Bot, iconHashToBigInt } from '../index.js'
+import { type AuditLogEntry, type Bot } from '../index.js'
 
 export function transformAuditLogEntry(bot: Bot, payload: DiscordAuditLogEntry): AuditLogEntry {
   const auditLogEntry = {
     id: bot.transformers.snowflake(payload.id),
-    changes: payload.changes?.map((change) => {
-      switch (change.key) {
-        case '$add':
-        case '$remove':
-          return {
-            key: change.key,
-            new: change.new_value?.map((val) => ({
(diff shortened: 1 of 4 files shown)


lts20050703:
I'm not sure about this... I do think it's a pain to maintain the transformer, but it feels wrong that this *specific* (can you even call this) "transformer" (anymore?) is not transforming the very thing it should be transforming.

@Skillz4Killz Thoughts on this?

Fleny113:
> is not transforming the very thing it should be transforming.

be clear, the transformer transforms the AuditLogEntry, i have removed the one for the AuditLogChange, however i do understand what you mean, the issue is that with 230+ props keeping track of everything is hard

PR #3825, opened 2024-08-01
fix(rest): Use a `_` check instead of the manual one on the switch

Fleny113:
This reverts the recently merged #3795 and changes the approach to check for a `_` instead of the manual check we were doing before.


AwesomeStickz (approved):

lts20050703 (approved):
LGTM!

PR #3828, opened 2024-08-04
fix(rest): Add the property if it has `_` to the object

Fleny113:
In #3825 an if was added to ignore all fields that had `_` in the key as they were already in snake_case. However, the check simply ignored the key, but we do not mutate the object, we create another one so all keys with `_` provided to rest are ignored.


lts20050703 (approved):
LGTM!

PR #3848, opened 2024-08-11
fix(examples): Add commands dir import to examples register commands script

Fleny113:
Currently we do not load the commands before registering them to discord, this makes so it always sends an empty array


AwesomeStickz (approved):

lts20050703 (approved):
LGTM!

PR #3851, opened 2024-08-12
feat(gateway): Expose `makePresence` from `GatewayManager`

Fleny113:
Currently the only way to make a shard have a presence is editing it after the shard is connected, however the `DiscordenoShard` class has a `makePresence` that unless it is used manually it cannot be used as the `GatewayManager` does not accept this method.

This pr adds the `makePresence` method to the `GatewayManager` and passes it down when creating a shard


AwesomeStickz (dismissed):

lts20050703 (approved):
LGTM!

AwesomeStickz (approved):

lts20050703 (approved):

PR #3858, opened 2024-08-12
build(website): Update website deps, remove 1s load time

Fleny113:
- Update website dependencies
- Remove 1s load time on /
- Remove default docusaurus assets
- Update README


website/src/pages/index.tsx
lts20050703, line 1: If we're removing the fake 1s load time, why not remove the loading screen altogether? Like remove the `const [loading, setLoading] = useState(true)`, the useEffect hook, and the `if (loading) return`

lts20050703 (changes requested):

website/src/pages/index.tsx
Fleny113: Docusaurus will still pre-render that loading page, so when you load the page the loading is there until tge react bundle loads. Removing it creates a semi-unsyled hone page, reason for this is that while yes, the browser will wait for the CSS file before showing the HTML, since we use CSS in JS (styled-components) those styles are part of the JS bundle that did not yet load.

AwesomeStickz (approved):

TriForMine (approved):

lts20050703 (approved):

PR #3863, opened 2024-08-13
fix(bot)!: Remove direct calls to transformers

Fleny113:
Remove the direct calls to transformers in favor of the `bot.transformers` object and support for customizers

> [!WARNING]
> This is breaking, because `transformInteractionDataResolved` did not use the ""standard"" way to make a transformer and used 3 parameters, this has been changed to be 2 parameters the second one being an object however this is a breaking change


AwesomeStickz:
Could you sort the transformers and customizers by their name while you're at it?

AwesomeStickz (approved):

lts20050703 (approved):
LGTM!

PR #3864, opened 2024-08-13
test(rest:e2e): Increase timeout for send a direct message e2e test

Fleny113:
Currently the "Send a Direct Message" E2E test fails due to the 30s timeout. This increases it to 60s hoping this is long enough to make it pass

### DIFF
--- a/packages/rest/tests/e2e/member.spec.ts
+++ b/packages/rest/tests/e2e/member.spec.ts
@@ -54,7 +54,9 @@ describe('Member tests', () => {
     expect(member2.nick).to.null
   })
 
-  it('Send a direct message', async () => {
+  it('Send a direct message', async function () {
+    this.timeout(60000)
+
     // DM test only on dd unit testing bot
     if (rest.applicationId.toString() !== '770381961553510451') return
     // Itoh Alt ID


AwesomeStickz (approved):

AwesomeStickz (approved):

packages/rest/tests/e2e/member.spec.ts
```diff
@@ -54,7 +54,9 @@ describe('Member tests', () => {
     expect(member2.nick).to.null
   })
 
-  it('Send a direct message', async () => {
+  it('Send a direct message', async function () {
```
lts20050703, line 57: why change from arrow function to normal function?

lts20050703 (approved):

packages/rest/tests/e2e/member.spec.ts
Fleny113: `this`.

for a more serious explanation: An arrow function does have have a this itself, and mocha gives a this for configure the timeout

Fleny113:
Not sure how they are related but #3849 already fixed the timeout issue for the direct message e2e, so this pr is no longer needed

*not sure how it was causing a timeout not an exception but I'm not going to ask*

PR #3887, opened 2024-08-28
add .yarnrc.yml to advanced example solution

Zero-Tools:
The .yarnrc.yml file is needed or else when yarn runs it will attempt to use PnP

### DIFF
--- a/examples/advanced/.yarnrc.yml
+++ b/examples/advanced/.yarnrc.yml
@@ -1,0 +1,4 @@
+# This file is needed or else when yarn runs in `docker build` will attempt to use PnP
+compressionLevel: mixed
+enableGlobalCache: false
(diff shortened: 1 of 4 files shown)


Fleny113:
Can you add the `.yarnrc.yml` to the other examples as well?

Zero-Tools:
@Fleny113 I've added the other files and updated the comment. let me know if there is anything else I can do

Fleny113 (approved):
The comment in the files still refers to `docket build` even if these projects don't have docker configured. But for the rest seems fine

Zero-Tools:
@Fleny113 I've updated the inline comments

Fleny113 (approved):

AwesomeStickz (approved):

Issue #3519, opened 2024-03-31
[Bug]: fatal "InvalidStateError: readyState not OPEN" error after losing connection


Macil:
### Runtime

Deno

### Runtime Version

1.42.0

### Version

18.0.1

### Describe the bug

Sometimes when I leave a Discord bot running on my computer, let it fall asleep, and then come back on, I find that the bot has exited with this error output:

```
error: Uncaught (in promise) InvalidStateError: readyState not OPEN
      shard.socket?.send(
                    ^
    at WebSocket.send (ext:deno_websocket/01_websocket.js:326:13)
    at https://deno.land/x/discordeno@18.0.1/gateway/shard/startHeartbeating.ts:52:21
    at eventLoopTick (ext:core/01_core.js:203:13)
```

This doesn't always happen when my bot is running while my computer falls asleep. Sometimes it works fine.

I have not looked into Discordeno's code, but having previously developed programs using Websockets and running into the same "readyState not OPEN" error, I know that this can be caused by calling .send() on a Websocket that is still connecting to a server or when the connection has been closed already. I think it's likely that Discordeno has a heartbeat timer that can call .send() while the connection is closed and not yet reopened and/or while the connection has been previously closed and is still reconnnecting now. The fix would be to either make the heartbeat function check the current state of the Websocket before sending with it, or to make the heartbeat timer shut down entirely once the websocket has been closed and restart once a new websocket connection has been established.

### What should've happened?

Discordeno should always be able to handle a connection closing and reopening without a fatal error.

### Code to reproduce the bug

```typescript
import "https://deno.land/std@0.220.1/dotenv/load.ts";
import {
  createBot,
  Intents,
  startBot,
} from "https://deno.land/x/discordeno@18.0.1/mod.ts";

const bot = createBot({
  token: Deno.env.get("DISCORD_TOKEN")!,
  intents: Intents.Guilds | Intents.GuildMessages | Intents.DirectMessages |
    Intents.MessageContent,
  events: {
    ready(bot, payload) {
      console.log(new Date(), "ready");
    },
    messageCreate(bot, message) {
      console.log(new Date(), "messageCreate");
    },
  },
});
await startBot(bot);
```

Macil:
[...]

PR #3907, opened 2024-09-07
fix(gateway): Fix shard reconnecting in a loop

Fleny113:
In some situations, Shards are getting in a loop of resuming -> close -> resuming -> ..., this pr attempts to fix it and improves the logging for shards.

The main change is in `handleClose` under the case where we do try to resume the connection, if the shard as already resuming, we are now fallback to identify instead of trying again

This **may** also fix the bug in #3519 as by setting the socket to undefined in `handleClose` it may now not try to send a heatbeat on a closed websocket.


packages/gateway/src/Shard.ts
```diff
@@ -568,10 +582,11 @@ export class DiscordenoShard {
           this.logger.debug(`[Shard] Heartbeat not acknowledged for shard #${this.id}. Assuming zombied connection.`)
           this.close(ShardSocketCloseCodes.ZombiedConnection, 'Zombied connection, did not receive an heartbeat ACK in time.')
 
-          return await this.identify()
+          await this.resume()
```
lts20050703, line 585: Is this typo? Why was it changed from identify to resume?

lts20050703 (changes requested):

packages/gateway/src/Shard.ts
Fleny113: > If a client does not receive a heartbeat ACK between its attempts at sending heartbeats, this may be due to a failed or "zombied" connection. The client should immediately terminate the connection with any close code besides 1000 or 1001, then reconnect and attempt to Resume.

From the discord docs, https://discord.com/developers/docs/topics/gateway#heartbeat-interval-example-heartbeat-ack

vxern (dismissed):

lts20050703 (changes requested):

packages/gateway/src/Shard.ts
```diff
@@ -250,7 +250,7 @@ export class DiscordenoShard {
           properties: this.gatewayConfig.properties,
           intents: this.gatewayConfig.intents,
           shard: [this.id, this.gatewayConfig.totalShards],
-          presence: await this.makePresence?.(),
+          presence: await this.makePresence(),
```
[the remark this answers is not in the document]
Fleny113: Well, yes. By default it is a function that just returns `{}`, and the user can ovveride it either using the options passed to the constructor or by setting the func afterwards but the func should always exist

lts20050703 (approved):
LGTM!

PR #3976, opened 2024-11-09
fix(bot): role.color being undefined when value is 0

8auu:


Fleny113 (changes requested):
rather than changing the type, can you change the transformer to check explicitly for undefined? rn i belive it is using js falsy/true-y values and 0 is a falsy value.

Fleny113 (approved):

PR #3977, opened 2024-11-09
refactor(rest): make proxy.authorization optional in CreateRestManagerOptions

AwesomeStickz:
Currently dd requires `proxy.authorization` in `CreateRestManagerOptions`, and not all proxies needs a custom auth, so this can be an optional property instead


Fleny113 (changes requested):
`rest.authorizationHeader` is always present: https://github.com/discordeno/discordeno/blob/main/packages/rest/src/manager.ts#L86

We could allow `undefined` for the auth, maybe with a remark on it tho

PR #4067, opened 2024-12-29
ci: Use mocha for both deno and bun

Fleny113:
Use mocha for both Deno and Bun

We want to test our library against the runtime, we don't want to test their Testing Framework, so for this reason we now use mocha for all 3 runtimes

This however (for deno specificly) requires patching mocha (cjs package) to change the entry-point it uses for yargs (cjs + esm package) as the current one makes Deno freak out since the file it imports does not have ".cjs" and yet it is a commonjs module.

This pr also has the following changes:
- Bump Deno to v2.1.x
- Bump Bun to v1.1.42
- Remove the `build:types` requirements for Deno and Bun Unit Tests CI
- Remove all the nested `.mocharc.json` in favor of `.mocharc.cjs` and `.mocharc.base.cjs` in the root of the repository
    - This is needed as for node we need to specify the ts-node loader, however Deno and bun don't need it and will make them crash for invalid flags given to the runtime
- `yarn patch mocha` to change the `yargs` entrypoint from `yargs/yargs` to `yargs` (will resolve to `yargs/index.js` which still does enough work to make mocha work)
- Remove `denoImportMap.json` in favor of `deno.json`
- `deno.json`:
    - Remove the lock file (we don't save it anyway)
    - Specify the node_modules dir is created manually
    - Allow sloppy-imports (stuff related to how deno resolves `.ts` and `.js` files in imports)
- remove the `before` & `after` logs from `packages/utils/bucker.spec.ts`
- use `globalThis` instead of `typeof global !== 'undefined' ? global : window` for mocking the `fetch` function in `packages/utils/tests/urlToBase64.spec.ts`
- Remove both scripts that were used to support `deno test` and `bun test`
- `turbo.json`: remove the `build:type` dependency from the Unit Testing scripts


Skillz4Killz:
@Fleny113 conflicts

PR #4077, opened 2024-12-30
feat(utils): add EmbedsBuilder.addFields()

mezotv:
This implements a simple addFields function, similar to setFields, but instead of overwriting the previous fields it just adds them to the array. I also added a test for the addField function and fixed a typo in the tsdoc from earlier.


mezotv:
Ill try to fix the tests later moving this to a draft until that is figured out!

itohatweb (changes requested):
Would it be possible to move the renaming to `EmbedsBuilder` to a separate PR? So we can avoid needing to bump a major version a bit longer.

mezotv:
> Would it be possible to move the renaming to `EmbedsBuilder` to a separate PR? So we can avoid needing to bump a major version a bit longer.

Its not actually a rename just changes the tsdoc. The class was always named like this. But yea if needed I can make a seperate pr for it!

packages/utils/src/builders/embeds.ts
```diff
@@ -43,6 +43,18 @@ export class EmbedsBuilder extends Array<DiscordEmbed> {
...
   }
 
+  /**
+   * Adds multiple new fields to the embed fields array
+   *
+   * @param {DiscordEmbedField[]} fields
+   * @returns {EmbedsBuilder}
+   */
+  addFields(fields: DiscordEmbedField[]): this {
+    this.#currentEmbed.fields.push(fields)
```
Fleny113, line 53:
```suggestion
    this.#currentEmbed.fields.push(...fields)
```

packages/utils/src/builders/embeds.ts
```diff
@@ -37,12 +37,28 @@ export class EmbedsBuilder extends Array<DiscordEmbed> {
...
+      ...(inline !== undefined && { inline }),
     })
 
     return this
   }
 
+  /**
+   * Adds multiple new fields to the embed fields array
+   *
+   * @param {DiscordEmbedField[]} fields
```
Fleny113, line 49:
```suggestion
   * @param {DiscordEmbedField[]} fields - The fields to add
```

Fleny113 (dismissed):

Fleny113 (dismissed):

Skillz4Killz (approved):

PR #4088, opened 2025-01-06
refactor(types): split `discord.ts` into multiple files

Fleny113:
Split the content of `discord.ts` as currently it is quite hard to find objects

> [!TIP]
> For review you are better off doing one commit at a time so you don't get massive diffs, you may still get some however they will be way more manageable both to review and load in the GitHub UI


packages/types/src/discordeno.ts
lts20050703, line 1: Do you plan to split discordeno.ts into multiple files too?

packages/gateway/src/manager.ts
lts20050703, line 1: Unrelated change but it's just formatting so it's fine

lts20050703 (approved):
LGTM!

lts20050703 (changes requested):
I meant request changes not approve, check the above comments

packages/types/src/index.ts
```diff
@@ -1,3 +1,30 @@
...
+export * from './discord/stageInstance.js'
+export * from './discord/sticker.js'
+export * from './discord/subscription.js'
+export * from './discord/teams.js'
+export * from './discord/user.js'
+export * from './discord/voice.js'
+export * from './discord/webhook.js'
+export * from './discord/webhookEvents.js'
 export * from './discordeno.js'
 export * from './shared.js'
```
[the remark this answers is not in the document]
Fleny113: no i did not?
anyway the pr is about discord.ts mainly tho some types where in shared.ts i just moved those as well so you are sure that if you have a docs page you should find the type as long as it was in /types to begin with (like TransportCompression is in /gateway and it will be moved in a separate pr

packages/types/src/discordeno.ts
Fleny113: yes but the pr is already big enough as is

packages/gateway/src/manager.ts
Fleny113: ~~stupid merge commits~~
I can remove it in case

lts20050703 (approved):
LGTM!

PR #4090, opened 2025-01-13
feat(rest)!: Parse json body when possibile in rest proxy

Fleny113:
Currently we always do `.text()` on the rest proxy response body, however when the server gives us json (or at least claims it does) we can parse it as JSON and use that in the error cause

### DIFF
--- a/packages/rest/src/manager.ts
+++ b/packages/rest/src/manager.ts
@@ -555,12 +555,12 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
         const result = await fetch(`${rest.baseUrl}/v${rest.version}${route}`, rest.createRequestBody(method, options))
 
         if (!result.ok) {
-          const errText = await result.text().catch(() => null)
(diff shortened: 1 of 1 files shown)


packages/rest/src/manager.ts
```diff
@@ -555,12 +555,12 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
         const result = await fetch(`${rest.baseUrl}/v${rest.version}${route}`, rest.createRequestBody(method, options))
 
         if (!result.ok) {
-          const errText = await result.text().catch(() => null)
+          const body = (result.headers.get('Content-Type') === 'apllication/json' ? await result.json() : await result.text()).catch(() => null)
```
itohatweb, line 558:
```suggestion
          const body = (result.headers.get('Content-Type') === 'application/json' ? await result.json() : await result.text()).catch(() => null)
```

itohatweb (changes requested):

PR #4096, opened 2025-01-19
fix(gateway): wrong calculation in calculateWorkerId()

Qreepex:
The old formula does not spread the shards correctly over the workers, you will get something like 1 shard on worker 0 and 7 shards on worker 1
The new formula does it correct :)

### DIFF
--- a/packages/gateway/src/manager.ts
+++ b/packages/gateway/src/manager.ts
@@ -271,7 +271,7 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
       )
     },
     calculateWorkerId(shardId) {
-      const workerId = Math.min(shardId % gateway.totalWorkers)
+      const workerId = shardId % gateway.totalWorkers
       gateway.logger.debug(
         `[Gateway] Calculating workerId: Shard: ${shardId} -> Worker: ${workerId} -> Per Worker: ${gateway.shardsPerWorker} -> Total: ${gateway.totalWorkers}`,
       )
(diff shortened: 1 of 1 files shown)


Fleny113 (approved):

packages/gateway/src/manager.ts
```diff
@@ -271,7 +271,7 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
       )
     },
     calculateWorkerId(shardId) {
-      const workerId = Math.min(shardId % gateway.shardsPerWorker, gateway.totalWorkers - 1)
+      const workerId = shardId % gateway.totalWorkers
```
AwesomeStickz, line 274:
```suggestion
      const workerId = Math.min(Math.floor(shardId / manager.shardsPerWorker), manager.totalWorkers - 1);
```

AwesomeStickz:
After some discussion, it's decided to support both ways with an option instead (spreading out evenly and incrementally), so for this PR, we can just fix the incorrect formula, unless you want to add the option yourself

packages/gateway/src/manager.ts
lts20050703: Is this the fixed formula? If so, what do you think if you just commit the suggestion and merge?

packages/gateway/src/manager.ts
```diff
@@ -271,7 +271,7 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
       )
     },
     calculateWorkerId(shardId) {
-      const workerId = Math.min(shardId % gateway.shardsPerWorker, gateway.totalWorkers - 1)
+      const workerId = Math.min(Math.floor(shardId / manager.shardsPerWorker), manager.totalWorkers - 1);
```
AwesomeStickz, line 274:
```suggestion
      const workerId = Math.min(Math.floor(shardId / gateway.shardsPerWorker), gateway.totalWorkers - 1);
```

Fleny113 (approved):

PR #4097, opened 2025-01-20
Fix desired components being undefined

SuperYummy:
Message.components is a component array. Accessing desired components may lead to type error

### DIFF
--- a/packages/bot/src/transformers/message.ts
+++ b/packages/bot/src/transformers/message.ts
@@ -159,7 +159,7 @@ export function transformMessage(
   if (props.attachments && payload.message.attachments?.length)
     message.attachments = payload.message.attachments.map((attachment) => bot.transformers.attachment(bot, attachment))
   if (props.channelId && payload.message.channel_id) message.channelId = bot.transformers.snowflake(payload.message.channel_id)
-  if (props.components && payload.message.components)
+  if (props.components && payload.message.components?.length)
     message.components = payload.message.components.map((comp) => bot.transformers.component(bot, comp))
   if (props.content) message.content = payload.message.content ?? EMPTY_STRING
   if (props.editedTimestamp && payload.message.edited_timestamp) message.editedTimestamp = Date.parse(payload.message.edited_timestamp)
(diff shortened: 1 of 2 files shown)


Fleny113:
> Accessing desired components may lead to type error
Not sure what you mean by this, however we check the length of arrays to avoid the empty array memory if i remember correctly

SuperYummy:
Avoiding the empty array memory makes the components undefined. What do you think of changing Message#components type from Component[] to Component[] | undefined?

lts20050703:
@Fleny113 I think I understand the problem that @SuperYummy's PR is trying to solve.

This is the same issue as last time, if `message.content` is an empty string `""`, `transformers/message.ts` will not set message.content to an empty string `""`, leaving it `undefined`, EVEN IF `transformers.desiredProperties.message.content = true`

OLD
https://github.com/discordeno/discordeno/blob/43624f47a6622ac4aa7053b286894af6eba28138/packages/bot/src/transformers/message.ts#L144

NEW, CURRENT
https://github.com/discordeno/discordeno/blob/f4f4665b895a8be202425155831c9a09b727bd7e/packages/bot/src/transformers/message.ts#L164


Last time we fixed this issue by stop checking for truthy `payload.message.content`, this time do we want

1. Keep the empty array `[]` if `desiredProperties.message.components = true`
2. Do not keep empty array, and change the typing from `components: Component[]` to `components: Component[] | undefined`

I don't think we're going with option 1 because we don't want to increase memory usage, but let me know what you think.

Fleny113:
I talked about this a bit with @AwesomeStickz and probably the best eay to go about this is adding the optional to these props

for the rest idk, I'm more of the idea we should check for undefined specifically always beside for array where we should do length === 0, and maybe do something about nulls as well
for example rn it is very annoying to deal with anything that might be null

AwesomeStickz:
I think we can just have an option that enables having empty arrays if the option is true. It should be good in theory but it could still cause you issues with TypeScript - for example if we don't make the `message.components` for example optional, it'll result in an error at runtime if you're using it without setting the option. Now if we make the type optional, you'll have to use `!.` or `?.` every time even if you use the option for having empty arrays, which will be annoying, but handling types for these props dynamically based on the option's value would make everything complex so idk what we can do about typings for these props

PR #4102, opened 2025-01-25
Update Slash Command Name regex

Fleny113:
- Upstream: https://github.com/discord/discord-api-docs/pull/7303

> [!NOTE]
> While the upstream pr uses `'` we use `ʼ` as that is the actual value that is now allowed


Jiralite:
You may want to be aware of https://github.com/discord/discord-api-docs/issues/7371.

Fleny113:
Moving the PR to draft until https://github.com/discord/discord-api-docs/issues/7371 gets addressed by discord

lts20050703:
> Moving the PR to draft until [discord/discord-api-docs#7371](https://github.com/discord/discord-api-docs/issues/7371) gets addressed by discord

@Fleny113 Have discord addressed this? I checked the PR and it's still open, why did you undraft this PR?

Fleny113:
> Have discord addressed this? I checked the PR and it's still open, why did you undraft this PR?

No, but reading the comment on the original discussion and the issue instead of a ' is another unicode char, the api and client do seem to confirm this fact so I'd say we can do the change anyway but with the correct char

lts20050703 (approved):
LGTM!

PR #4103, opened 2025-01-27
fix(shutdown): prevent resume when closed with `GatewayCloseEventCodes.NormalClosure`.

zargovv:
According to Discord developers portal documentation page, [initializing a disconnect](https://discord.com/developers/docs/events/gateway#initiating-a-disconnect) section suggests that close codes `1000` (`GatewayCloseEventCodes.NormalClosure`) and `1001` will invalidate bot and bot will appear offline. By taking a look at how shutting down process works in `discordeno`, it seemed like the library is lacking the ability to fully shutdown with the bot appearing offline.

Therefore, I experimented with closing gateway manually by providing the close codes `1000` and `1001`. The code `1001` is not allowed by `discordeno`, which is fine. However, the code `1000` triggers session resume, which I assumed a bug.

I kindly ask to review my pull request and consider the little change I've added. Appreciated in advance!

Open to discussions!

### DIFF
--- a/packages/gateway/src/Shard.ts
+++ b/packages/gateway/src/Shard.ts
@@ -407,6 +407,7 @@ export class DiscordenoShard {
       }
       // On these codes a manual start will be done.
       case GatewayCloseEventCodes.NormalClosure:
+      case GatewayCloseEventCodes.GoingAway:
       case ShardSocketCloseCodes.Shutdown:
       case ShardSocketCloseCodes.ReIdentifying:
       case ShardSocketCloseCodes.Resharded:
--- a/packages/gateway/src/Shard.ts
+++ b/packages/gateway/src/Shard.ts
@@ -406,6 +406,7 @@ export class DiscordenoShard {
         return
       }
       // On these codes a manual start will be done.
+      case GatewayCloseEventCodes.NormalClosure:
       case ShardSocketCloseCodes.Shutdown:
       case ShardSocketCloseCodes.ReIdentifying:
       case ShardSocketCloseCodes.Resharded:
--- a/packages/types/src/shared.ts
+++ b/packages/types/src/shared.ts
@@ -830,6 +830,8 @@ export type PermissionStrings = keyof typeof BitwisePermissionFlags
 export enum GatewayCloseEventCodes {
   /** A normal closure of the gateway. You may attempt to reconnect. */
   NormalClosure = 1000,
+  /** The endpoint is going away. Invalidates bot and bot will appear offline.  */
+  GoingAway = 1001,
   /** We're not sure what went wrong. Try reconnecting? */
   UnknownError = 4000,
   /** You sent an invalid [Gateway opcode](https://discord.com/developers/docs/topics/opcodes-and-status-codes#gateway-gateway-opcodes) or an invalid payload for an opcode. Don't do that! */


Fleny113:
> The code `1001` is not allowed by discordeno, which is fine. 

Can i ask what do you mean by this? Do you mean the fact that `GatewayCloseEventCodes` does not have a `1001` value?

> suggests that close codes `1000` and `1001` will invalidate bot and bot will appear offline.
> ...
> However, the code `1000` triggers session resume, which I assumed a bug.

[Not related to your pr] I wonder why we have `A normal closure of the gateway. You may attempt to reconnect.` as a comment on `NormalClosure` then, as that does sound misleading if discord invalidates the session on a `1000` close code.

It is also maybe worth checking with the others on the team if we should replace our `3000` value (and maybe others?) with 1000 as:
- `Shutdown` (3000) could be put to 1000 instead, assuming Shutdown has the intended behavior (scope) to terminate the session and not leave the resume available
- `TestingFinished` (3064) could be put to 1000 too as it is a private-ish code used for our integration tests
- `Resharded` / `ReIdentifying` based on how they behave but they are fine as-is i'd say

zargovv:
> Can i ask what do you mean by this? Do you mean the fact that GatewayCloseEventCodes does not have a 1001 value?

I am sorry for misleading you by this statement. After carefully reviewing the reasons, I figured out the real reason catching `InvalidAccessError` for close code `1001`. It was Deno's `WebSocket.close` method and has nothing to do with `discordeno`.

Example for the context:
```ts
await bot.gateway.shutdown(1001, "User requested bot stop"); // -> error: Uncaught (in promise) InvalidAccessError: The close code must be either 1000 or in the range of 3000 to 4999: received 1001
```

I also considered changing `Shutdown` code to `1000`, but didn't want to make any potentially breaking changes to codebase, as I am not too familiar with your library just yet.

Fleny113:
Ok so i did some testing, discord does indeed invalidate the session on both 1000 and 1001.

My only doubt it that i do have some logs about discordeno receiving a 1000 close code from discord, and this pr would make it so on such occasions the shard would go offline, but since i had some issues with my bot gateway due to other issues (like a token reset i got due to a bug we now fixed)

So, I'd say:
- can you please add the case for 1001 as well?
- if someone on @discordeno/team (or others that see this) can tell me if they too happen to have a 1000 or 1001 close code in their bot logs so we can determinate if move the code for 1000 from the manual start to the identify switch case

zargovv:
Sure! I updated the `GatewayCloseEventCodes` with a value for `Going Away` code and added a clause to the switch statement.

lts20050703 (approved):
LGTM! However, I checked https://discord.dev and there seems to be no explanation for what code 1000 or 1001 is?
I'm just wondering where you get the "Going Away" text explanation from.

Fleny113:
> However, I checked https://discord.dev and there seems to be no explanation for what code 1000 or 1001 is?
> I'm just wondering where you get the "Going Away" text explanation from.

RFC 6455, the WebSocket protocol RFC

itohatweb:
> LGTM! However, I checked https://discord.dev and there seems to be no explanation for what code 1000 or 1001 is? I'm just wondering where you get the "Going Away" text explanation from.

https://discord.com/developers/docs/events/gateway#initiating-a-disconnect

AwesomeStickz (approved):

PR #4111, opened 2025-01-30
docs: Use as instead of specifying the generic

Fleny113:
If we specify the generic other generics stop being inferred so if we specify the desired properties behavior to ChangeType it type errors as TS still thinks that it should be RemoveKey


lts20050703 (approved):
LGTM!

PR #4123, opened 2025-02-02
feat(bot,rest,types): Guild incidents data

Fleny113:
- upstream: https://github.com/discord/discord-api-docs/pull/7360


packages/types/src/discordeno.ts
lts20050703, line 1: Should we add the following notes?
- Both `invites_disabled_until` and `dms_disabled_until` can be enabled for a maximal timespan of 24 hours in the future.
- \* Supplying `null` disables the action.

lts20050703 (dismissed):
LGTM!

packages/types/src/discordeno.ts
Fleny113: I also added the note about the fact that the string is an ISO string to be specific

lts20050703 (approved):
LGTM!

AwesomeStickz (approved):

PR #4124, opened 2025-02-05
fix(types, gateway)!: UpdatePresence inconsistencies

### LINKED ISSUE #4104: [Bug]: Shard presence has incomplete and out of date type(s)
### Runtime

NodeJS

### Runtime Version

23.6.0

### Version

21.0.0

### Describe the bug

Currently we have 2 (if not more) interfaces to update the presence of a shard:
- `BotStatusUpdate`
- `StatusUpdate`
both in /packages/gateway/src/types.ts

However, `BotStatusUpdate` uses `BotActivity` for the activity which is missing `state` for custom presences, and `StatusUpdate` uses `DiscordActivity` (from which it omits `created_at` for some reason) which is also incorrect as discord documents that bots can only use `name`, `type`, `state` and `url`

### What should've happened?

There should just be 1 interface with the correct types

### Code to reproduce the bug

```typescript

```

Fleny113:
Remove `BotStatusUpdate`, `StatusUpdate` and `BotActivity` in favor of `DiscordUpdatePresence` for the first two and `DiscordBotActivity` for `BotActivity`, both new types are from `@discordeno/types`

It also removes usage of the now removed types from our lib code, there is a reference of `StatusUpdate` in the bigbot example

closes #4104


lts20050703 (dismissed):
LGTM!

packages/types/src/discord/gateway.ts
```diff
@@ -291,7 +291,18 @@ export interface DiscordGatewayPayload {
...
-// TODO: Add Presence Update: https://discord.com/developers/docs/events/gateway-events#update-presence-gateway-presence-update-structure
+
+/** https://discord.com/developers/docs/events/gateway-events#update-presence-gateway-presence-update-structure */
+export interface DiscordUpdatePresence {
+  /** Unix time (in milliseconds) of when the client went idle, or null if the client is not idle */
+  since: number | null
+  /** The user's activities */
+  activities: DiscordBotActivity[]
+  /** The user's new status */
+  status: keyof typeof PresenceStatus
```
Fleny113, line 302: Do we want to change `PresenceStatus` to be a string enum and use it instead of its keys?

packages/gateway/src/types.ts
```diff
@@ -175,22 +175,17 @@ export interface ShardSocketRequest {
...
-  /** The user's activities */
-  activities: BotActivity[]
-  /** The user's new status */
-  status: keyof typeof PresenceStatus
-}
+/**
+ * https://discord.com/developers/docs/topics/gateway-events#update-presence
+ * @deprecated Use {@link DiscordUpdatePresence} from `@discordeno/types` instead
+ */
+export type BotStatusUpdate = DiscordUpdatePresence
```
Fleny113, line 182: We could remove those since the next version of `discordeno` is going to be a major anyway

packages/types/src/discord/gateway.ts
Fleny113: Checking this in code it doesn't really work based on how we use this, so for now at least this is fine as is

packages/gateway/src/types.ts
AwesomeStickz: can be removed, it's not something that'll annoy people id say

lts20050703 (approved):
Looks like we stopped releasing minor version since v18, huh?

Fleny113:
> Looks like we stopped releasing minor version since v18, huh?

We have another braking in rest anyway, we change a .text() to .json() /shrug

AwesomeStickz (approved):

PR #4126, opened 2025-02-08
refactor(gateway): Improve request identify flow

Fleny113:
This is a slight refactor to the `tellWorkerToIdentify`, `identify` and `requestIdentify` methods in the gateway manager.

Currently the identify flow is as follows:
The code prepares the buckets and starts all the promises (one for bucket) where each spawn one shard at a time
However, the way it uses to spawn one shard at a time is by calling `tellWorkerToIdentify` which does nothing beside a redirect to `identify` which creates the shard, appends the promise resolver to the bucket identify requests array and call the shard identify method to create the actual connection.
Resolver to then be called by the shard `shardIsReady` method/event to let the `identify` promise resolve, to let `tellWorkerToIdentify` resolve, to let the for-of loop in `spawnShards` continue

However, this has 2 issues:
1. The identify requests in a bucket are always one due to this, which is not really an issue per-se
2. If multiple shards get an invalid session and need to re-identify they will all do that at the same time, as the current code does not really request anything, it just does it and the "request" is just a way to tell when we are done (this also applies to anything calling `.identify()` on the manager without waiting for the previous identify to complete & wait the delay)

---

The refactor makes use of `requestIdentify` which was previously not used

Now the code actually requests the identify and does not just do it, so if we call `.identify()` manually on 10 shards they will do it one at a time (assuming 1 bucket), something that before would have created issues

This also fixes the "bulk invalid session" issue as since they do actually do a request and wait for the permission, they wait to re-identify one at a time


packages/gateway/src/manager.ts
```diff
@@ -413,30 +397,35 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
...
       }
 
+      await gateway.requestIdentify(shardId)
+
+      gateway.logger.debug(`[Gateway] Shard #${shardId} successfully requested the identify, proceding with the identify.`)
+      await shard.identify()
+    },
+
+    async requestIdentify(shardId) {
+      gateway.logger.debug(`[Gateway] Shard #${shardId} requested an identify`)
```
AwesomeStickz, line 403:
```suggestion
      gateway.logger.debug(`[Gateway] Shard #${shardId} requested an identify.`)
```

packages/gateway/src/manager.ts
```diff
@@ -413,30 +397,35 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
...
-      return await new Promise((resolve) => {
-        // Mark that we are making an identify request so another is not made.
-        bucket.identifyRequests.push(resolve)
-        gateway.logger.debug(`[Gateway] Identifying Shard #${shardId}.`)
-        // This will trigger identify and when READY is received it will resolve the above request.
-        shard?.identify()
-      })
+
+      if (!bucket) {
+        throw new Error("Can't request an identify for a shard that can't be assigned to a bucket.")
```
AwesomeStickz, line 408:
```suggestion
        throw new Error("Can't request identify for a shard that is not assigned to any bucket.")
```

packages/gateway/src/manager.ts
```diff
@@ -413,30 +397,35 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
...
+      }
+
+      await bucket.leakyBucket.acquire()
     },
+
     async kill(shardId: number) {
       const shard = this.shards.get(shardId)
       if (!shard) {
-        return gateway.logger.debug(`[Gateway] A kill for Shard #${shardId} was requested, but the shard could not be found`)
+        gateway.logger.debug(`[Gateway] A kill for Shard #${shardId} was requested, but the shard could not be found`)
```
AwesomeStickz, line 419:
```suggestion
        gateway.logger.debug(`[Gateway] Shard #${shardId} was requested to be killed, but the shard could not be found.`)
```

packages/gateway/src/manager.ts
```diff
@@ -776,7 +765,13 @@ export interface GatewayManager extends Required<CreateGatewayManagerOptions> {
   /** Shutdown all shards. */
   shutdown: (code: number, reason: string, clearReshardingInterval?: boolean) => Promise<void>
   sendPayload: (shardId: number, payload: ShardSocketRequest) => Promise<void>
-  /** Allows users to hook in and change to communicate to different workers across different servers or anything they like. For example using redis pubsub to talk to other servers. */
+  /**
+   * Allows users to hook in and change to communicate to different workers across different servers or anything they like.
+   * For example using redis pubsub to talk to other servers.
+   *
+   * @remarks
+   * This should wait for the worker to have identified the shard before resolving the returned promise
```
AwesomeStickz, line 771:
```suggestion
   * This gets called for all shards at once so you may want to use `gateway.requestIdentify` before passing the message if you're overriding this function.
```

packages/gateway/src/manager.ts
```diff
@@ -738,8 +725,8 @@ export interface GatewayManager extends Required<CreateGatewayManagerOptions> {
     number,
     {
       workers: Array<{ id: number; queue: number[] }>
-      /** Requests to identify shards are made based on whether it is available to be made. */
-      identifyRequests: Array<(value: void | PromiseLike<void>) => void>
+      /** The bucket to queue the identifies */
```
AwesomeStickz, line 728:
```suggestion
      /** The bucket to queue the identifies. */
```

packages/gateway/src/manager.ts
Fleny113: This does not get called all at once if you wait for the shard to have identified, so this is not correct

packages/gateway/src/manager.ts
```diff
@@ -392,14 +378,9 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
           },
           events: options.events ?? {},
           logger: this.logger,
-          requestIdentify: async () => {
-            await gateway.identify(shardId)
-          },
+          requestIdentify: async () => await gateway.requestIdentify(shardId),
           shardIsReady: async () => {
```
lts20050703, line 382: Should we rename the method from `shardIsReady` to `shardIdentified` or `shardIsIdentified`?

lts20050703 (approved):
LGTM!

packages/gateway/src/Shard.ts
```diff
@@ -809,7 +811,7 @@ export interface ShardCreateOptions {
   logger?: Pick<typeof logger, 'debug' | 'info' | 'warn' | 'error' | 'fatal'>
   /** The handler to request a space to make an identify request. */
   requestIdentify?: () => Promise<void>
-  /** The handler to alert the gateway manager that this shard has received a READY event. */
+  /** The handler to alert the gateway manager that this shard has identified. */
   shardIsReady?: () => Promise<void>
```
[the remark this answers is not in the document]
AwesomeStickz: It's likely getting removed in another PR as it no longer serves any purpose

packages/gateway/src/Shard.ts
```diff
@@ -618,7 +618,9 @@ export class DiscordenoShard {
 
         // When resumable is false we need to re-identify
         if (!resumable) {
+          // TODO: check if we need to close the connection before waiting for the identify
```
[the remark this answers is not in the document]
Fleny113: mainly forgot to remove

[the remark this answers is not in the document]
AwesomeStickz: I think it's no longer there

lts20050703 (approved):

AwesomeStickz (approved):

Issue #4107, opened 2025-01-30
[Bug]: Gateway spawning all workers and shards immediatly


Fleny113:
### Runtime

NodeJS

### Runtime Version

23.6.0

### Version

21.0.0

### Describe the bug

The current gateway code will create ALL shards and workers immediately, especially with a lot of shards this creates a lot of issues 

### What should've happened?

The shards should be spawned right before we need to / can identify them

### Code to reproduce the bug

```typescript
import { createBot } from '@discordeno/bot'

const bot = createBot({
  token: process.env.TOKEN,
  events: {
    // Add the events, which are needed to show the bug
  },
  // Don't forget to add all necessary intents
  intents: 0,
})
bot.start()
```

Fleny113:
After checking the code, the issue is with the big bot template (where this can be reproduced)

However, there is a slight naming issue with now the name our methods:
- `tellWorkerToIdentify` should return a promise that is resolved when the shard has identified, not before
- The current code has a `requestIdentify` method that... does nothing

PR #4127, opened 2025-02-08
example(bigbot): Spawn one shard at a time

Fleny113:
Spawn only one shard a time in the bigbot example.

since the issue was about the bigbot template and not the library code itself, fixes the bug in #4107


examples/bigbot/src/gateway/gatewayManager.ts
```diff
@@ -37,6 +38,23 @@ gatewayManager.tellWorkerToIdentify = async (workerId, shardId, bucketId) => {
...
+    }
+  }
+
+  worker.on('message', waitForShardIdentified)
+
+  await promise
+
+  worker.off('message', waitForShardIdentified)
+
+  // This needs to be here AND in the generic worker message event, because this methods needs wait before it resolves
```
lts20050703, line 56:
```suggestion
  // This needs to be here AND in the generic worker message event, because this methods needs to wait before it resolves
```

lts20050703 (approved):
LGTM!

AwesomeStickz (changes requested):
This PR should be merged after #4126 depending on the changes in there

PR #4128, opened 2025-02-08
examples: Use `tsc` to build instead of `swc`

Fleny113:
There is (close to) no reason to use SWC to build the examples when TSC is fine.

Sure, typescript is slower since it is also doing the type check, however since discordeno does not have any issue building when using `tsc` and we need to install it anyway there is no reason to not use it


lts20050703:
@Fleny113 what's wrong with the current setup that uses swc? (aka if it aint broke, why fix it)

Fleny113:
> what's wrong with the current setup that uses swc? 

You don't get type checking unless you are in a editor with TSS

Also its like <a lot> less dependencies that do not really matter to discordeno

lts20050703 (approved):

PR #4138, opened 2025-02-23
fix(gateway): Keep state to avoid unwanted disconnections

Fleny113:
This keeps a state to know that we wanted to end the connection with a code of 1000 or 1001 to avoid unwanted disconnections in case discord / someone in the middle of the TCP connection (such as Cloudflare) sends a close code of 1000 or 1001.

### DIFF
--- a/packages/gateway/src/Shard.ts
+++ b/packages/gateway/src/Shard.ts
@@ -55,6 +55,16 @@ export class DiscordenoShard {
   bucket: LeakyBucket
   /** Logger for the bucket. */
   logger: Pick<typeof logger, 'debug' | 'info' | 'warn' | 'error' | 'fatal'>
+  /**
+   * Is the shard going offline?
+   *
+   * @remarks
+   * This will be true if the close method has been called with either 1000 or 1001
+   *
+   * @internal
+   * This is for internal purposes only, and subject to breaking changes.
+   */
+  goingOffline = false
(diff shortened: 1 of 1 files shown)


packages/gateway/src/Shard.ts
```diff
@@ -440,6 +450,17 @@ export class DiscordenoShard {
...
+      case GatewayCloseEventCodes.NormalClosure:
+      case GatewayCloseEventCodes.GoingAway: {
+        if (this.goingOffline) {
+          this.state = ShardState.Disconnected
+          this.events.disconnected?.(this)
+
+          return
+        }
+
+        // if goingOffline is false we can fall through to the switch and since this is here it will go in the default case
```
AwesomeStickz, line 462:
```suggestion
        // If goingOffline is false, default case should be run, so we don't use break / return here
```

packages/gateway/src/Shard.ts
```diff
@@ -440,6 +450,17 @@ export class DiscordenoShard {
         await this.identify()
         return
       }
+      case GatewayCloseEventCodes.NormalClosure:
```
AwesomeStickz, line 453:
```suggestion
      // NOTE: This case must always be right above the cases that runs with default case because of how switch works when you don't break / return, more info below.
      case GatewayCloseEventCodes.NormalClosure:
```

AwesomeStickz (approved):

PR #4155, opened 2025-03-13
fix(rest): deleteWebhookMessage() and editOriginalInteractionResponse() doesn't require auth

AwesomeStickz:
These endpoints don't require auth, source from community maintained docs:
- [deleteWebhookMessage](https://docs.discord.sex/resources/webhook#delete-webhook-message)
- [editOriginalInteractionResponse](https://docs.discord.sex/interactions/receiving-and-responding#edit-original-interaction-response)

### DIFF
--- a/packages/rest/src/manager.ts
+++ b/packages/rest/src/manager.ts
@@ -894,7 +894,7 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
     },
 
     async deleteWebhookMessage(webhookId, token, messageId, options) {
-      await rest.delete(rest.routes.webhooks.message(webhookId, token, messageId, options))
+      await rest.delete(rest.routes.webhooks.message(webhookId, token, messageId, options), { unauthorized: true })
     },
 
     async deleteWebhookWithToken(webhookId, token) {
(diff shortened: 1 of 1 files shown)


lts20050703:
How much do we trust the community maintained docs? Have you tried and check to see if it's actually unauth? Is there any reason you need it to be unauth?

AwesomeStickz:
> How much do we trust the community maintained docs?

It's maintained by multiple people so it's mostly right if not entirely. I wouldn't say we should _rely_ on it but rather use as a reference when we want to find undocumented things such as this so it's best to check it before changing something based on that docs

> Have you tried and check to see if it's actually unauth?

Yes

> Is there any reason you need it to be unauth?

It can cause multiple edge case scenario problems, like:
- https://github.com/germanoeich/nirn-proxy/issues/23
- Proxies limiting to 50 reqs/s when you send bot token, which would mean that you won't be able to handle interactions properly if you have more than 50 per sec even though Discord allows you
- An issue where Discord sends back ratelimit headers in some routes like interaction callback endpoint when you pass bot token (which they don't send back when you don't send bot token) even though Discord doesn't enforce them, causing proxies to actually consider the ratelimit headers like the above 50 reqs/s for example

Overall it's just the best idea to not send auth header to unauth endpoints to prevent any unexpected behaviour

lts20050703 (approved):

Fleny113 (approved):

PR #4189, opened 2025-05-05
fix(types): Remove comment limit on number of components

Fleny113:
The limit is still in place however it is 40 total now, so it is a bit more complicated to explain and the discord docs are better at doing so than what we can do with JSDoc

- Upstream: https://github.com/discord/discord-api-docs/pull/7534

### DIFF
--- a/packages/types/src/discord/components.ts
+++ b/packages/types/src/discord/components.ts
@@ -318,7 +318,7 @@ export enum SeparatorSpacingSize {
 export interface DiscordContainerComponent extends DiscordBaseComponent {
   type: MessageComponentTypes.Container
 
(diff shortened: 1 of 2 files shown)


AwesomeStickz:
Do we need this comment anymore since it's basically just describing the code?

Fleny113:
> Do we need this comment anymore since it's basically just describing the code?

It is JSDoc so i think it is fine if it just describes the type

AwesomeStickz:
> > Do we need this comment anymore since it's basically just describing the code?
> 
> It is JSDoc so i think it is fine if it just describes the type

fair

AwesomeStickz (approved):

PR #4203, opened 2025-05-23
ci: Use node v20 in CI

Fleny113:
Node 18 went EOL in April 31 2025.

The only reason we may want to keep using Node 18 for is for unit testing to ensure we don't break compatibility with that as we still support it. So for now unit tests still use Node 18


AwesomeStickz (dismissed):

lts20050703 (dismissed):
LGTM!

Is there any actual reason to continue using Node 18 for unit test though?

Skillz4Killz (approved):

PR #4204, opened 2025-05-24
feat(gateway): add options.spreadShardsInRoundRobin to spread shards in a round-robin manner

### LINKED ISSUE #4108: [Feat]: option on gateway for spreading shards evenly
We currently have the shards spread across workers incrementally, for example if there are 100 shards and 7 workers with `shardsPerWorker` set to 10, it'll put shard 0 - 9 in worker 0, shard 10 - 19 in worker 1 and so on, and put the remaining shards in the last worker, so worker 6 will have shard 60 - 69, and worker 7 will have shard 70 - 99.

This method may be ideal for some people but not all, some people may prefer spreading them evenly among the workers, so in this case, all workers except 0 and 1 will have 14 shards, while worker 0 and 1 will have 15 shards, this spreads the load evenly among the workers. This issue proposes an option to do this so people can use the one they like.

AwesomeStickz:
Resolves #4108


Fleny113 (approved):

lts20050703 (approved):
LGTM!

PR #4206, opened 2025-05-25
feat(gateway): Resharding with workers 

### LINKED ISSUE #4110: [Feat]: Support resharding when having multiple workers
Currently resharding only works when you don't have workers, but it should be changed to support it when using workers as well

Fleny113:
Rework how the resharding works to make it better when using outside of the in-memory gatewayManager:
1. When a call to reshard is done it re-does all the buckets, worker id calculations etc
1. Calls `gateway.resharding.tellWorkerToPrepare`, it works the same wait as `gateway.tellWorkerToIdentify` but instead it tells the worker to prepare (and identify) the shard for later. This method should still way for the shard to have identified.
1. When all the calls to `gateway.resharding.tellWorkerToPrepare` are done it calls `gateway.resharding.onReshardingSwich`[^name] which does the work that was done before in `gateway.reshard.shardIsPending` but without any of the work to ensure it gets called at the end. This work includes setting the proper message event and shutting down the old shards.

[^name]: Better name suggestions are welcome

And example of this is implemented in the bigbot example

---

closes #4110


examples/bigbot/src/gateway/gatewayManager.ts
```diff
@@ -25,8 +25,48 @@ const gatewayManager = createGatewayManager({
...
+  resharding: {
+    getSessionInfo: restManager.getGatewayBot,
+  },
 })
 
+gatewayManager.resharding.tellWorkerToPrepare = async (workerId, shardId, bucketId) => {
+  logger.info(`Tell worker to prepare, workerId: ${workerId}, shardId: ${shardId}, bucketId: ${bucketId}`)
+
+  const worker = workers.get(workerId) ?? createWorker(workerId)
+  workers.set(workerId, worker)
```
lts20050703:
```suggestion
  let worker = workers.get(workerId)
  if (!worker) {
    worker = createWorker(workerId)
    workers.set(workerId, worker)
  }
```

examples/bigbot/src/gateway/worker/worker.ts
```diff
@@ -38,6 +41,65 @@ parentPort.on('message', async (message: WorkerMessage) => {
 
     return
   }
+  if (message.type === 'PrepareShard') {
+    logger.info(`Preparing shard #${message.shardId}`)
+    totalShards = message.totalShards
+    const shard = pendingShards.get(message.shardId) ?? createShard(message.shardId)
+    pendingShards.set(message.shardId, shard)
```
lts20050703:
```suggestion
    let shard = pendingShards.get(message.shardId)
    if (!shard) {
      shard = createShard(message.shardId)
      pendingShards.set(message.shardId, shard)
    }
```

examples/bigbot/src/gateway/worker/worker.ts
```diff
@@ -38,6 +41,65 @@ parentPort.on('message', async (message: WorkerMessage) => {
...
+
+    // Ignore the events
+    // TODO: If you need 'gateway.resharding.updateGuildsShardId' it you can listen to only the ready event and use the data from that event for the function call
```
lts20050703, line 54: Does this TODO needs to be fixed or not?

examples/bigbot/src/gateway/worker/worker.ts
```diff
@@ -38,6 +41,65 @@ parentPort.on('message', async (message: WorkerMessage) => {
...
+    for (const shard of pendingShards.values()) {
+      shard.events.message = handleShardMessageEvent
+    }
+
+    // Old shards stop processing events
+    for (const shard of shards.values()) {
+      const oldHandler = shard.events.message
+
+      shard.events.message = async function (_, message) {
+        // Member checks need to continue but others can stop
```
lts20050703, line 79: Why does member checks need to be continued?

packages/gateway/src/manager.ts
```diff
@@ -62,8 +62,7 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
       enabled: options.resharding?.enabled ?? true,
       shardsFullPercentage: options.resharding?.shardsFullPercentage ?? 80,
       checkInterval: options.resharding?.checkInterval ?? 28800000, // 8 hours
-      shards: new Collection(),
-      pendingShards: new Collection(),
+      shards: new Map(),
```
lts20050703, line 66: Why are we switching from Collection to Map?

packages/gateway/src/manager.ts
```diff
@@ -720,22 +711,34 @@ export interface GatewayManager extends Required<CreateGatewayManagerOptions> {
...
+    shards: Map<number, Shard>
     /** Handler to check if resharding is necessary. */
     checkIfReshardingIsNeeded: () => Promise<{ needed: boolean; info?: Camelize<DiscordGetGatewayBot> }>
-    /** Handler to begin resharding. */
+    /**
+     * Handler to begin resharding.
+     *
+     * @remarks
+     * This function will resolve once the resharding is done.
+     * So when all the calls to {@link tellWorkerToPrepare} and {@link onReshardingSwich} are done.
```
lts20050703:
```suggestion
     * This function will resolve when the resharding is done (when all the calls to {@link tellWorkerToPrepare} and {@link onReshardingSwich} are done)
```

examples/bigbot/src/gateway/worker/worker.ts
AwesomeStickz: It's not really member "checks" but guild member chunks need to be continued because there may be pending chunks in old shards and you want to receive them before you shutdown them, but now that I think about it, we shut them down anyway so this won't do anything 🤔

Fleny113: 1) i guess i meant to write "chunks" not "checks"
2) it was already this way so i kept it this way (for stickz comment)

packages/gateway/src/manager.ts
Fleny113: We don't use anything specific from a Collection and i dont remember if there was another reason i changed it

examples/bigbot/src/gateway/worker/worker.ts
Fleny113: ... maybe

I don't know what to use the `updateGuildsShardId` for in  bigbot, so we could set it up and have an empty function that either has a todo/dd-setup (assuming i remembering how we called those) comment or has a log of the guild ids

examples/bigbot/src/gateway/worker/worker.ts
lts20050703: @Fleny113 I don't think we wait for the last member chunk (like Stickz said, we shut them down anyways so this doesn't do anything), so it doesn't really matter if we keep it like it this or make the function body empty (though I would vote for making the function empty)

Fleny113: The issue is that there are 2 issues:
1- If you have a requestMembers promise, that promise will never resolve as it will never receive the last event, and I'd say this is a bug.
2- We do not have a way to prevent this either as of now, i don't remember exactly how we handle the chunks, however i believe they are sent to the bot process to collect and resolve the promise. So, to actually wait for all chunks to go before shutting down we need code in the shards/workers to know all the request members made and when we get the last chunk of a request "mark it" as done so that we can shut down a shard only after it is done

tl;dr: keeping as is (in main, and this pr) and having it empty does create a bug

lts20050703: Hmm...
Then in that case feel free to kept the code that way then fix it later in another PR

AwesomeStickz: Would suggest keeping it the way it is and just fixing in another PR too, yes

examples/bigbot/src/gateway/worker/worker.ts
lts20050703: Ok, if we dedice to have `updateGuildsShardId` in bigbot, let's do so in another PR

PR #4237, opened 2025-07-03
feat!: Add invite flags for guest invites

Fleny113:
- upstream: https://github.com/discord/discord-api-docs/pull/6247

> [!WARNING]
> This is breaking because now a member joined_at is nullable

### DIFF
--- a/packages/bot/src/transformers/invite.ts
+++ b/packages/bot/src/transformers/invite.ts
@@ -1,5 +1,5 @@
 import type { DiscordApplication, DiscordInviteCreate, DiscordInviteMetadata } from '@discordeno/types'
-import { type InternalBot, type Invite, isInviteWithMetadata } from '../index.js'
+import { type InternalBot, type Invite, ToggleBitfield, isInviteWithMetadata } from '../index.js'
 
 export function transformInvite(
   bot: InternalBot,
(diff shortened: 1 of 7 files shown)


packages/bot/src/transformers/invite.ts
```diff
@@ -42,6 +42,7 @@ export function transformInvite(
     if (props.expiresAt && payload.invite.expires_at) {
       invite.expiresAt = Date.parse(payload.invite.expires_at)
     }
+    if (props.flags && payload.invite.flags) invite.flags = new ToggleBitfield(payload.invite.flags)
```
lts20050703, line 45:
```suggestion
    if (props.flags && payload.invite.flags) invite.flags = payload.invite.flags
```
Took me a while to realize, but I think you're confusing between Toggles and Flags
- When there are multiple boolean keys returned from Discord, Toggles are constructed from ToggleBitfield. Toggles merge multiple boolean into one bitfield.
- For Flags, the transformer keep it as-is (since it's already a bitfield). For example, https://github.com/discordeno/discordeno/blob/598b839af7b2473507d0a0833adb9aa0c2829c40/packages/bot/src/transformers/channel.ts#L107

lts20050703 (changes requested):

packages/bot/src/transformers/invite.ts
Fleny113: We do have transformer that use ToggleBitfield for flags (see the message flags and the user flags for example)

lts20050703: My bad then, I guess the channel transformer is the one that needs to be updated.

lts20050703 (approved):
LGTM!

PR #4238, opened 2025-07-03
feat: user tags

Fleny113:
- upstream: https://github.com/discord/discord-api-docs/pull/6836

### DIFF
--- a/packages/bot/src/transformers/types.ts
+++ b/packages/bot/src/transformers/types.ts
@@ -1768,6 +1768,8 @@ export interface User {
   verified: boolean
   /** data for the user's collectibles */
   collectibles?: Collectibles
+  /** The user's primary guild */
+  primaryGuild?: UserPrimaryGuild | null
 }
 
 export interface Collectibles {
@@ -1784,6 +1786,17 @@ export interface Nameplate {
   label: string
   /** background color of the nameplate, one of: `crimson`, `berry`, `sky`, `teal`, `forest`, `bubble_gum`, `violet`, `cobalt`, `clover`, `lemon`, `white` */
   palette: string
+}
+
+export interface UserPrimaryGuild {
+  /** The id of the primary guild */
+  identityGuildId?: bigint
+  /** Thether the user is displaying the primary guild's server tag. This can be `null` if the system clears the identity, e.g. because the server no longer supports tags. */
+  identityEnabled?: boolean
+  /** The text of the user's server tag. Limited to 4 characters */
+  tag?: string
+  /** The server tag badge hash */
+  badge?: bigint
 }
 
 export interface VoiceRegion {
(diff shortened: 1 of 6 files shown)


packages/bot/src/transformers/types.ts
```diff
@@ -1768,6 +1768,8 @@ export interface User {
   verified: boolean
   /** data for the user's collectibles */
   collectibles?: Collectibles
+  /** The user's primary guild */
+  primaryGuild?: UserPrimaryGuild | null
```
lts20050703:
```suggestion
  primaryGuild?: UserPrimaryGuild
```

If I recall correctly, there was a discussion on preserving null values, but for now it should be optional

packages/bot/src/transformers/types.ts
```diff
@@ -1786,6 +1788,17 @@ export interface Nameplate {
   palette: string
 }
 
+export interface UserPrimaryGuild {
+  /** The id of the primary guild */
+  identityGuildId?: bigint
+  /** Thether the user is displaying the primary guild's server tag. This can be `null` if the system clears the identity, e.g. because the server no longer supports tags. */
```
lts20050703:
```suggestion
  /** Whether the user is displaying the primary guild's server tag. This can be undefined if the system clears the identity, e.g. because the server no longer supports tags. */
```

Technically after the transformer run, this key is not set, so it's "not defined", so I'm debating this

Alternatively we could keep the wording like Discord, it's just wrong since we intentionally skip null values

packages/types/src/discord/user.ts
```diff
@@ -40,6 +40,8 @@ export interface DiscordUser {
   avatar_decoration_data?: DiscordAvatarDecorationData
   /** data for the user's collectibles */
   collectibles?: DiscordCollectibles
```
lts20050703:
```suggestion
  avatar_decoration_data?: DiscordAvatarDecorationData | null
  /** data for the user's collectibles */
  collectibles?: DiscordCollectibles | null
```

<https://discord.com/developers/docs/resources/user#user-object-user-structure>

packages/types/src/discord/user.ts
```diff
@@ -69,6 +71,18 @@ export enum PremiumTypes {
   NitroBasic,
 }
 
+/** https://discord.com/developers/docs/resources/user#user-object-user-primary-guild */
+export interface DiscordUserPrimaryGuild {
+  /** The id of the primary guild */
+  identity_guild_id: string | null
+  /** Thether the user is displaying the primary guild's server tag. This can be `null` if the system clears the identity, e.g. because the server no longer supports tags. */
```
lts20050703:
```suggestion
  /** Whether the user is displaying the primary guild's server tag. This can be `null` if the system clears the identity, e.g. because the server no longer supports tags. */
```

lts20050703 (changes requested):

lts20050703 (approved):
LGTM!

PR #4244, opened 2025-07-12
fix(rest): Use `.startsWith` on the Content-Type check

Fleny113:
The proxy may respond with a Content-Type such as `application/json; charset=utf-8` and we should still use result.json() in such scenario, for this reason we need to check if the start of the header value is `application/json`


lts20050703 (approved):
LGTM!

PR #4245, opened 2025-07-12
feat(rest): Add events

Fleny113:


packages/rest/src/manager.ts
```diff
@@ -467,15 +489,10 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
...
         if (response.status !== HttpResponseCode.TooManyRequests) {
-          const body = response.headers.get('Content-Type') === 'application/json' ? ((await response.json()) as object) : await response.text()
-
           options.reject({ ok: false, status: response.status, statusText: response.statusText, body })
           return
         }
 
-        // Consume the response body to avoid leaking memory
-        await response.arrayBuffer()
-
```
lts20050703, line 477: Did you move this to somewhere else? Can't seem to find it.

packages/rest/src/types.ts
```diff
@@ -199,6 +199,30 @@ export interface CreateRestManagerOptions {
    * @default logger // The logger exported by `@discordeno/utils`
    */
   logger?: Pick<typeof logger, 'debug' | 'info' | 'warn' | 'error' | 'fatal'>
+  /** Events for the rest manager */
+  events?: {
+    /**
+     * Emitted when a request is made to the API.
+     *
+     * @remarks
+     * The body that will be sent to the API is available in the `extra` parameter, do not consume the body in the Request object.
```
AwesomeStickz:
```suggestion
     * The body that will be sent to the API is available in the `extra` parameter. Do not consume the body in the `Request` object and use the one in the `extra` parameter instead.
```

packages/rest/src/types.ts
```diff
@@ -199,6 +199,30 @@ export interface CreateRestManagerOptions {
...
+     */
+    request?: (request: Request, extra: { body: any }) => void
+    /**
+     * Emitted when a response is received from the API.
+     *
+     * @remarks
+     * Both the request and the response are available in the `extra` parameter.
```
AwesomeStickz:
```suggestion
     * Both the request and the response body are available in the `extra` parameter. Do not consume the body in the `Request` or `Response` object and use the one in the `extra` parameter instead.
```

packages/rest/src/manager.ts
```diff
@@ -455,6 +467,13 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
       })
       rest.logger.debug(`request fetched from ${url} with status ${response.status} & ${response.statusText}`)
 
+      const body = response.headers.get('Content-Type') === 'application/json' ? ((await response.json()) as object) : await response.text()
```
AwesomeStickz:
```suggestion
        // Sometimes the Content-Type may be "application/json; charset=utf-8", for this reason, we need to check the start of the header
        const body = await (result.headers.get('Content-Type')?.startsWith('application/json') ? result.json() : result.text()).catch(() => null)
```

packages/rest/src/manager.ts
```diff
@@ -455,6 +467,14 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
...
 
+      // Sometimes the Content-Type may be "application/json; charset=utf-8", for this reason, we need to check the start of the header
+      const body = await (response.headers.get('Content-Type')?.startsWith('application/json') ? response.json() : response.text()).catch(() => null)
+
+      rest.events.response(request, response, {
+        requestBody: options.requestBodyOptions?.body,
+        responseBody: body,
+      })
+
       // Mark request and completed
```
lts20050703:
```suggestion
      // Mark request as completed
```

packages/rest/src/manager.ts
```diff
@@ -441,9 +447,15 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
...
+      rest.events.request(request, {
+        body: options.requestBodyOptions?.body,
+      })
+
       rest.logger.debug(`sending request to ${url}`, 'with payload:', { ...payload, headers: loggingHeaders })
-      const response = await fetch(url, payload).catch(async (error) => {
+      const response = await fetch(request).catch(async (error) => {
         rest.logger.error(error)
+        rest.events.requestError(request, error, { body: options.requestBodyOptions?.body })
         // Mark request and completed
```
lts20050703:
```suggestion
        // Mark request as completed
```

packages/rest/src/manager.ts
```diff
@@ -500,7 +515,7 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
       }
 
       // Discord sometimes sends no response with no content.
```
lts20050703:
```suggestion
      // Discord sometimes sends a response with no content
```

packages/rest/src/manager.ts
```diff
@@ -585,7 +610,7 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
           throw error
         }
 
-        return result.status !== 204 ? await result.json() : undefined
+        return result.status !== 204 ? (typeof body === 'string' ? JSON.parse(body) : body) : undefined
```
lts20050703, line 612: Are we returning the correct thing?

Fleny113: so, before it was always an object bc of result.json, now i have to get the body before this point for the event so we get either a string or an object, so if we JSON.parse the string we get an object. So this function will get an object in the end, like it was before

PR #4246, opened 2025-07-12
build(deps): Update biome to v2

Fleny113:
- **update biome to v2**
- **Run biome check --write**


biome.jsonc
```diff
@@ -44,22 +48,7 @@
...
-      // Remove the yarn script as it is minified
-      ".yarn",
-      // Ignore build outputs
-      "bunTestsDist",
-      "build",
-      "coverage",
-      "denoTestsDist",
-      "dist"
-    ],
+    "includes": ["**", "!**/package.json", "!**/node_modules", "!**/.docusaurus", "!**/.turbo", "!**/.yarn", "!**/build", "!**/coverage", "!**/dist"],
```
lts20050703:
```suggestion
    "includes": [
      "**",
      // Ignore the package.json and leave yarn to format it
      "!**/package.json",
      "!**/node_modules",
      // Ignore turbo and docusaurus cache/stuff
      "!**/.docusaurus",
      "!**/.turbo"
      // Remove the yarn script as it is minified,
      "!**/.yarn",
      // Ignore build outputs
      "!**/build",
      "!**/coverage",
      "!**/dist"
    ],
```

lts20050703 (dismissed):

examples/reaction-roles/src/bot.ts
```diff
@@ -1,7 +1,7 @@
 import { createBot } from '@discordeno/bot'
+import { config } from 'dotenv'
 import events from './events/index.js'
 
-import { config } from 'dotenv'
 config()
```
SerenModz21: Suggestion:
Instead of importing the config function and executing it, why not import dotenv/config instead?
```ts
import 'dotenv/config'
```
It automatically calls the config function with its default options, saving you 1 line.

Fleny113: This is the way we do it because it's what we also have on https://discordeno.js.org/docs/beginner/env and https://discordeno.js.org/docs/examples/reactionroles, and we use the function as it does look similar to the Deno code.

Also this would not be the appropriate pr to change this, however thanks anyway for the suggestion.

lts20050703: I think we should do this for the ESM only PR (https://github.com/discordeno/discordeno/pull/4227), since that's what the dotenv README suggests for ESM https://www.npmjs.com/package/dotenv#how-do-i-use-dotenv-with-import

Fleny113: The ESM only is for publishing ESM only, so stripping the SWC config for cjs and editing the package.json(s) to allow for `require(esm)` if you use commonjs. (and about that pr: if someone uses commonjs and could test that the `require(esm)` doesn't break anything that would be appreciated)

My guess is that the authors of dotenv use `dotenv/config` this way if your other imports rely on process.env.XYZ they will be populated unlike if you call the function due to the hoisting of imports

lts20050703: I think for Node usage, we should switch to `dotenv/config`, to avoid users/devs confusion when `process.env` is empty in other files

lts20050703 (dismissed):

examples/reaction-roles/src/bot.ts
SerenModz21: I have now created a separate PR for such: https://github.com/discordeno/discordeno/pull/4258

My apologies for incorrectly mentioning it in this PR!

lts20050703: No worries, it's OK to mention it in this PR so that later we can decide if we want to change it in this PR or in another PR.

lts20050703 (approved):

Skillz4Killz (approved):
LGTM

PR #4255, opened 2025-07-19
feat(types): Add new rich presence url & status display type fields

Fleny113:
-upstream: https://github.com/discord/discord-api-docs/pull/7674

### DIFF
--- a/packages/types/src/discord/gateway.ts
+++ b/packages/types/src/discord/gateway.ts
@@ -777,7 +777,7 @@ export enum ActivityTypes {
 export enum DiscordStatusDisplayType {
   /** Shows the name of the activity */
   Name = 0,
-  /** Shows the name of the state of the activity */
+  /** Shows the state of the activity */
   State = 1,
   /** Shows the details of the activity */
   Details = 2,
--- a/packages/types/src/discord/gateway.ts
+++ b/packages/types/src/discord/gateway.ts
@@ -775,11 +775,11 @@ export enum ActivityTypes {
 
 /** https://discord.com/developers/docs/events/gateway-events#activity-object-status-display-types */
 export enum DiscordStatusDisplayType {
-  /** Shows the name of the activity */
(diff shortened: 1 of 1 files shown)


packages/types/src/discord/gateway.ts
```diff
@@ -767,6 +773,16 @@ export enum ActivityTypes {
   Competing = 5,
 }
 
+/** https://discord.com/developers/docs/events/gateway-events#activity-object-status-display-types */
+export enum DiscordStatusDisplayType {
+  /** Shows the name of the activity */
+  Name = 0,
+  /** Shows the name of the state of the activity */
```
lts20050703:
```suggestion
  /** Shows the state of the activity */
```

packages/types/src/discord/gateway.ts
```diff
@@ -767,6 +773,16 @@ export enum ActivityTypes {
...
+  State = 1,
+  /** Shows the details of the activity */
+  Details = 2,
```
lts20050703: Why not use Discord's examples?

```suggestion
  /** Example: "Listening to Spotify" */
  Name = 0,
  /** Example: "Listening to Rick Astley" */
  State = 1,
  /** Example: "Listening to Never Gonna Give You Up" */
  Details = 2,
```

lts20050703 (dismissed):

packages/types/src/discord/gateway.ts
Fleny113: I find this example without any details kinda pointless but if you think it is better to stick to whatever discord has in their docs then sure

lts20050703: @Fleny113 I would argue that the examples immediately give you an idea of what it looks like, while the details are somewhat nothing burger (like it's just saying the same thing as the enum keys but unnecessarily longer) and doesn't give you an idea of what it looks like

lts20050703 (approved):

PR #4256, opened 2025-07-19
Split extra fields into their own types

Fleny113:
For now they are added back to the types they belong to to avoid
breaking in this pr


packages/types/src/discord/gateway.ts
```diff
@@ -411,21 +422,75 @@ export interface DiscordChannelPinsUpdate {
...
+  unavailable?: boolean
+  /** Total number of member in this guild */
+  member_count: number
+  /**
+   * States of members currently in voice channels
+   *
+   * @remarks
+   * Lacks the `guild_id` key
+   */
+  voice_states: Partial<DiscordVoiceState>[]
```
lts20050703: Why not `Omit` ?
```suggestion
  /**
   * States of members currently in voice channels
   *
   * @remarks
   * Lacks the `guild_id` key
   */
  voice_states: Omit<DiscordVoiceState, "guild_id">[]
```

packages/types/src/discord/channel.ts
```diff
@@ -1,12 +1,13 @@
...
 
 import type { DiscordMessageComponents } from './components.js'
+import type { DiscordThreadCreateExtra } from './gateway.js'
 import type { DiscordMember } from './guild.js'
 import type { DiscordAllowedMentions, DiscordAttachment, DiscordEmbed, MessageFlags } from './message.js'
 import type { DiscordUser } from './user.js'
 
 /** https://discord.com/developers/docs/resources/channel#channel-object-channel-structure */
-export interface DiscordChannel {
+export interface DiscordChannel extends Partial<DiscordThreadCreateExtra> {
```
lts20050703, line 10: Does this needs to be Partial when `DiscordThreadCreateExtra`'s `newly_created?: boolean` is already key optional?

```suggestion
export interface DiscordChannel extends DiscordThreadCreateExtra {
```

packages/types/src/discord/gateway.ts
```diff
@@ -411,21 +422,75 @@ export interface DiscordChannelPinsUpdate {
...
+/** https://discord.com/developers/docs/events/gateway-events#guild-ban-add-guild-ban-add-event-fields */
+export interface DiscordGuildBanAdd {
+  /** id of the guild */
+  guild_id: string
+  /** The banned user */
+  user: DiscordUser
+}
+
+/** https://discord.com/developers/docs/events/gateway-events#guild-ban-remove-guild-ban-remove-event-fields */
+export interface DsicordGuildBanRemove {
```
lts20050703:
```suggestion
export interface DiscordGuildBanRemove {
```

packages/types/src/discord/gateway.ts
```diff
@@ -411,21 +422,75 @@ export interface DiscordChannelPinsUpdate {
...
   /** The banned user */
   user: DiscordUser
 }
 
+/**
+ * https://discord.com/developers/docs/events/gateway-events#guild-ban-add-guild-ban-add-event-fields
+ * https://discord.com/developers/docs/events/gateway-events#guild-ban-remove-guild-ban-remove-event-fields
+ *
+ * @deprecated
+ * Use {@link DiscordGuildBanAdd} and {@link DsicordGuildBanRemove} instead.
```
lts20050703:
```suggestion
 * Use {@link DiscordGuildBanAdd} and {@link DiscordGuildBanRemove} instead.
```

lts20050703 (changes requested):

packages/types/src/discord/gateway.ts
Fleny113: discord in the docs still says Partial voice state, I wouldn't bet on the fact that guild_id is the only key missing, we can try and change it tho

packages/types/src/discord/channel.ts
Fleny113: I'd rather keep using Partial even if it isn't doing anything,  if discord were to add something that always exists to this Extra we would need to add the partial back
(and anyway these extends are probably going away in another pr, we (I and stickz seemed to agree) want to split the gateway event type from the type that may come from the rest api)

packages/types/src/discord/gateway.ts
Fleny113: checking the type from before it did omit, so ig it is fine to use omit

lts20050703: I think the docs doesn't have the "Omit" type, so that's why in the description they says it lacks the "guild_id" fields

lts20050703 (dismissed):
LGTM!

PR #4257, opened 2025-07-19
fix(e2e): Fix broken e2e (by removing tests)

Fleny113:
Discord has deprecated and now disabled bot from using the createGuild endpoint, we used it heavly for our tests, for now we disable these test (38 tests) however we should find another way to do these tests

### DIFF
--- a/packages/rest/tests/e2e/automod.spec.ts
+++ b/packages/rest/tests/e2e/automod.spec.ts
@@ -1,240 +1,216 @@
-import type { Camelize, DiscordChannel } from '@discordeno/types'
-import { AutoModerationActionType, AutoModerationEventTypes, AutoModerationTriggerTypes } from '@discordeno/types'
-import { expect } from 'chai'
-import { e2eCache, rest } from './utils.js'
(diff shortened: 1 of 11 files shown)


lts20050703 (commented):
Hmm yes we fixed broken tests by removing them :kekw:

Can we re-write the test to allow locally test runner providing a guild ID that they invited the bot into?

As for the automated test, how about someone creating a new Discord account (or re-use their own account), create a server, then provide the guild ID into the automated test?

Fleny113:
Ok jokes aside on removing tests when they break, i decided to remove them for now because
- github will send us a notification every single time e2e fail in main on every single commit
- we need to ask H0/ITOH/skillz what is the permissions the bot has in the guild id that is given to the CI in the repository settings, and if it isn't an Administrator permission or something do something about it

so while we fix that we disable them so that we don't get spammed by gh every single pr we merge (i do still need to ping someone, i want to do it today tho, yesterday i was tired)

lts20050703 (approved):
LGTM :kekw:

PR #4258, opened 2025-07-19
refactor: use dotenv/config instead of calling the function manually

SerenModz21:
Originally discussed in https://github.com/discordeno/discordeno/pull/4246, I have updated dotenv imports to use `dotenv/config` instead of calling `config()` as per the recommendation from the dotenv creators on their GitHub and npm readme.

I have tried to be consistent with other usages by having it at the top of the files with a new line below. However, I'm not sure if the same applies to markdown examples.


lts20050703 (approved):
LGTM!

PR #4262, opened 2025-07-24
api-docs: Add max of 512 chars on button links

Fleny113:


packages/types/src/discordeno.ts
```diff
@@ -178,7 +178,7 @@ export interface ButtonComponent extends BaseComponent {
   }
   /** Identifier for a purchasable SKU, only available when using premium-style buttons */
   skuId?: BigString
-  /** optional url for link-style buttons that can navigate a user to the web. Only type 5 Link buttons can have a url */
+  /** optional url for link-style buttons that can navigate a user to the web. Only type 5 Link buttons can have a url; Maximum 512 characters. */
```
lts20050703:
```suggestion
  /**
   * optional url for link-style buttons that can navigate a user to the web.
   * 
   * @remarks
   * Only {@link ButtonStyles.Link | Link} buttons can have a url.
   *
   * Maximum 512 characters.
   */
```

PR #4266, opened 2025-07-24
api-docs: Update the description for identity_enabled to be clearer

Fleny113:
- upstream: https://github.com/discord/discord-api-docs/pull/7701

### DIFF
--- a/packages/bot/src/transformers/types.ts
+++ b/packages/bot/src/transformers/types.ts
@@ -1814,7 +1814,13 @@ export interface Nameplate {
 export interface UserPrimaryGuild {
   /** The id of the primary guild */
   identityGuildId?: bigint
-  /** Whether the user is displaying the primary guild's server tag. This can be undefined if the system clears the identity, e.g. because the server no longer supports tags. */
+  /**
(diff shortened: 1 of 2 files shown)


packages/bot/src/transformers/types.ts
```diff
@@ -1814,7 +1814,13 @@ export interface Nameplate {
 export interface UserPrimaryGuild {
   /** The id of the primary guild */
   identityGuildId?: bigint
-  /** Whether the user is displaying the primary guild's server tag. This can be undefined if the system clears the identity, e.g. because the server no longer supports tags. */
+  /**
+   * Whether the user is displaying the primary guild's server tag.
+   *
+   * @remarks
+   * This can be `null` if the system clears the identity, e.g. because the server no longer supports tags.
```
lts20050703:
```suggestion
   * This can be `undefined` if the system clears the identity, e.g. because the server no longer supports tags.
```
Technically the property is not set, but I think optional is close enough?

lts20050703 (approved):
LGTM!

PR #4280, opened 2025-08-03
fix(bot): ACK the interaction only after a successful response

Fleny113:
If we ACK the interaction and the request is not successful and then the user tries to use the helpers again they will almost always get an error about unknown webhooks or unknown interactions

test code:
```ts
bot.events.interactionCreate = async (interaction) => {
  try {
    await interaction.respond({ components: [{ type: MessageComponentTypes.TextDisplay, content: 'Hello World!' }] })
  } catch (error) {
    await interaction.respond({ content: 'An error occurred while processing your request.\nDetails: ```ts\n' + inspect(error) + '```' })
  }
}
```


lts20050703 (approved):
LGTM!

PR #4281, opened 2025-08-03
changes and add stuffs

Ratnia:
Reason for pull request:
adding installation just for make it east way to add it the package more quickly.
Changes stuffs?: 
README.md add only the installation
and some files just only make sure no any errors.
any message: 
Is my first commit here, also Pull request, thank you!


packages/gateway/src/manager.ts
```diff
@@ -158,7 +158,7 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
             async message(_shard, payload) {
               if (payload.t === 'READY') {
                 await gateway.resharding.updateGuildsShardId?.(
-                  (payload.d as DiscordReady).guilds.map((g) => g.id),
+                  (payload.d as DiscordReady).guilds.map((g: any) => g.id),
```
Fleny113: Why the `any` type annotation?

packages/gateway/src/manager.ts
```diff
@@ -517,7 +517,7 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
           query: options?.query ?? (options?.limit ? undefined : ''),
           limit: options?.limit ?? 0,
           presences: options?.presences ?? false,
-          user_ids: options?.userIds?.map((id) => id.toString()),
+          user_ids: options?.userIds?.map((id: any) => id.toString()),
```
Fleny113: Why the `any` type annotation?

README.md
```diff
@@ -8,6 +8,13 @@ Discord API library for [Node.JS](https://nodejs.org), [Deno](https://deno.land)
 [![codecov](https://codecov.io/gh/discordeno/discordeno/branch/main/graph/badge.svg?token=SQI9OYJ7AK)](https://codecov.io/gh/discordeno/discordeno)
 ![action status](https://github.com/discordeno/discordeno/actions/workflows/lib-check.yml/badge.svg?event=push)
 
+# installation
```
Fleny113: You shouldn't have more than 1 main header (H1 in html)

```suggestion
## Installation
```

README.md
```diff
@@ -8,6 +8,13 @@ Discord API library for [Node.JS](https://nodejs.org), [Deno](https://deno.land)
...
+```
+- npm install @discordeno/bot
+- yarn add @discordeno/bot
+- pnpm install @discordeno/bot
+- bun install @discordeno/bot
```
Fleny113, line 16: I feel like this should be in the "Getting Started" we have below as an addition to the docs link we already have.

But it is a bit "hidden" down there, @discordeno/team any opinion on this?

Fleny113 (changes requested):

README.md
ratniaprivte: Thanks for info me this

packages/gateway/src/manager.ts
ratniaprivte: I actually was so confused by error as i did the contributions correctly but got confused so I added any for make annoying errors stops shows up

Fleny113: Can you share the error you were getting? If i had to guess it was related to the fact you never ran `yarn build:type` (or `yarn release-build`) so Typescript was confused on the type of something, but I might be wrong

Fleny113:
> Now better?

not really... you deleted the entire gateway manager, you just needed to revert the 2 changes you did in there

Ratnia:
Close it, sorry

PR #4282, opened 2025-08-03
Adding the new Stuffs on README.md

Ratnia:
i'm sorry for what i did on my old pull request so i did fixing stufs and is now back normally so i did this pull request this time could be better changes, kinda and not harmless

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Discord API library for [Node.JS](https://nodejs.org), [Deno](https://deno.land)
 [![codecov](https://codecov.io/gh/discordeno/discordeno/branch/main/graph/badge.svg?token=SQI9OYJ7AK)](https://codecov.io/gh/discordeno/discordeno)
 ![action status](https://github.com/discordeno/discordeno/actions/workflows/lib-check.yml/badge.svg?event=push)
 
-# installation
+# Installation
 ```
 - npm install @discordeno/bot
 - yarn add @discordeno/bot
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -8,6 +8,13 @@ Discord API library for [Node.JS](https://nodejs.org), [Deno](https://deno.land)
 [![codecov](https://codecov.io/gh/discordeno/discordeno/branch/main/graph/badge.svg?token=SQI9OYJ7AK)](https://codecov.io/gh/discordeno/discordeno)
 ![action status](https://github.com/discordeno/discordeno/actions/workflows/lib-check.yml/badge.svg?event=push)
 
+# Installation
```
Fleny113: As i said in the previous pr, you should only have 1 main header

```suggestion
## Installation
```

if it helps, if know html you may notice that `#` is an h1, `##` is an h2 and so on until h6 which is `######`

Fleny113: I feel like this should be in the "Getting Started" we have below as an addition to the docs link we already have.

But it is a bit "hidden" down there, @discordeno/team any opinion on this?

###### Copy pasted from the previous pr, i think it is fine having it here or around here in general, but i want to know the opinions of the others on the team

Ratnia (commented):
Alright then, that’s enough i guess!

README.md
lts20050703: @Fleny113 Maybe move the getting started to the top?

Fleny113:
I have changed "Installation" -> "Getting Started" & used [!TIP] for the tips myself.

I'd say se can merge this

Fleny113 (approved):

lts20050703 (approved):
LGTM!

PR #4283, opened 2025-08-03
test(rest): Don't create as many channel, ensure we delete the resources we create

Fleny113:
Currently we create a lot of channels despite having one "global" channel per E2E execution

Also we created emojis but never deleted them, so now tests are failing because we ran out of emoji capacity


Fleny113:
This fails as it needs a fix from #4284

PR #4285, opened 2025-08-04
fix(types)!: Fix nullability for DiscordEmoji

Fleny113:


packages/types/src/discord/emoji.ts
```diff
@@ -4,10 +4,10 @@ import type { DiscordUser } from './user.js'
 
 /** https://discord.com/developers/docs/resources/emoji#emoji-object-emoji-structure */
 export interface DiscordEmoji {
-  /** Emoji name (can only be null in reaction emoji objects) */
-  name?: string
   /** Emoji id */
-  id?: string
+  id: string | null
+  /** Emoji name (can be null only in reaction emoji objects) */
```
lts20050703:
```suggestion
  /**
   * Emoji name
   * @remarks Can be null only in reaction emoji objects
   */
```

PR #4288, opened 2025-08-04
fix(types)!: Sort & fix guild.ts properties

Fleny113:


TriForMine (dismissed):
LGTM!

packages/types/src/discord/guild.ts
```diff
@@ -3,33 +3,67 @@
...
+  discovery_splash: string | null
   /** True if the user is the owner of the guild */
   owner?: boolean
+  /** Id of the owner */
+  owner_id: string
+  /** Total permissions for the user in the guild (excludes overwrites and implicit permissions) */
+  permissions?: string
+  /**
+   * Voice region id for the guild
+   * @deprecated
```
lts20050703, line 45:
```suggestion
   * @deprecated This field is deprecated and is replaced by {@link DiscordChannel.rtc_region | Channel.rtc_region}
```

packages/types/src/discord/guild.ts
```diff
@@ -3,33 +3,67 @@
...
+  /** Splash hash */
+  splash: string | null
+  /** Discovery splash hash; only present for guilds with the "DISCOVERABLE" feature */
+  discovery_splash: string | null
   /** True if the user is the owner of the guild */
   owner?: boolean
+  /** Id of the owner */
+  owner_id: string
+  /** Total permissions for the user in the guild (excludes overwrites and implicit permissions) */
+  permissions?: string
```
lts20050703: https://discord.com/developers/docs/resources/guild#guild-object-guild-structure

> These fields are only sent when using the [GET Current User Guilds](https://discord.com/developers/docs/resources/user#get-current-user-guilds) endpoint and are relative to the requested user

Should we add this note?


```suggestion
  /** 
   * Total permissions for the user in the guild (excludes overwrites and implicit permissions)
   * @remarks This field is only sent when using the {@link getGuilds | getGuilds} helper
   */
  permissions?: string
```

packages/types/src/discord/guild.ts
```diff
@@ -3,33 +3,67 @@
...
+  /** Icon hash */
+  icon: string | null
+  /** Icon hash, returned when in the template object */
+  icon_hash?: string | null
+  /** Splash hash */
+  splash: string | null
+  /** Discovery splash hash; only present for guilds with the "DISCOVERABLE" feature */
+  discovery_splash: string | null
   /** True if the user is the owner of the guild */
   owner?: boolean
```
lts20050703: https://discord.com/developers/docs/change-log#whats-changing

> Existing Guilds owned by bots will have their ownership transferred to a real user

The docs hasn't updated yet. Should we keep it as-is or add a `@deprecated` note?

Also: 

https://discord.com/developers/docs/resources/guild#guild-object-guild-structure

> These fields are only sent when using the [GET Current User Guilds](https://discord.com/developers/docs/resources/user#get-current-user-guilds) endpoint and are relative to the requested user

Should we add this note?

```suggestion
/**
 * True if the user is the owner of the guild
 * @remarks This field is only sent when using the {@link getGuilds | getGuilds} helper
 */
owner?: boolean
```

packages/types/src/discord/guild.ts
```diff
@@ -598,6 +618,7 @@ export interface DiscordIncidentsData {
   raid_detected_at?: string | null
 }
 
+// TODO: We usually don't have types for the JSON request params, we should probably remove this
```
lts20050703, line 633: We already have (camelized) types for these, I think we should remove these since they're not used anywhere

packages/types/src/discord/guild.ts
```diff
@@ -646,6 +667,7 @@ export interface DiscordCreateGuildChannel {
   default_sort_order?: SortOrderTypes | null
 }
 
+// TODO: We usually don't have types for the JSON request params, we should probably remove this
```
lts20050703, line 682: We already have (camelized) types for these, I think we should remove these since they're not used anywhere

packages/types/src/discord/guild.ts
```diff
@@ -681,11 +703,12 @@ export interface DiscordPrunedCount {
 
 /** https://discord.com/developers/docs/resources/guild#get-guild-vanity-url */
 // TODO: This should provably have another name for clarity to what it really rappresent
-export interface DiscordVanityUrl {
+export interface DiscordVanityUrl extends Partial<Omit<DiscordInvite, 'code'>> {
   code: string | null
   uses: number
 }
 
+// TODO: We usually don't have types for the JSON request params, we should probably remove this
```
lts20050703, line 723: We already have (camelized) types for these, I think we should remove these since they're not used anywhere

packages/types/src/discord/guild.ts
```diff
@@ -681,11 +703,12 @@ export interface DiscordPrunedCount {
 
 /** https://discord.com/developers/docs/resources/guild#get-guild-vanity-url */
 // TODO: This should provably have another name for clarity to what it really rappresent
-export interface DiscordVanityUrl {
+export interface DiscordVanityUrl extends Partial<Omit<DiscordInvite, 'code'>> {
   code: string | null
```
lts20050703, line 719:
```suggestion
export interface DiscordVanityUrl extends Partial<DiscordInvite> {
  code: string | null
```
Technically with interface, you don't need to Omit 'code', since it will be overriden by the `code: string | null` anyways, but both way is fine.

packages/bot/src/transformers/reverse/member.ts
```diff
@@ -31,8 +31,8 @@ export function transformMemberToDiscordMember(bot: InternalBot, payload: Member
     avatar: payload.avatar ? iconBigintToHash(payload.avatar) : undefined,
     permissions: payload.permissions?.toString(),
     communication_disabled_until: payload.communicationDisabledUntil ? new Date(payload.communicationDisabledUntil).toISOString() : undefined,
-    deaf: payload.toggles?.deaf,
-    mute: payload.toggles?.mute,
+    deaf: payload.toggles?.deaf ?? false,
+    mute: payload.toggles?.mute ?? false,
```
lts20050703, line 35: Why this change? `DiscordMember.deaf` and `DiscordMember.mute` is typed optional, so I don't think this change is necessary?

```suggestion
    deaf: payload.toggles?.deaf,
    mute: payload.toggles?.mute,
```

packages/types/src/discord/guild.ts
Fleny113: I think i will remove the Discord prefixed json params in another pr, so for now the TODO stays

packages/types/src/discord/guild.ts
Fleny113: I think i will remove the Discord prefixed json params in another pr, so for now the TODO stays

packages/types/src/discord/guild.ts
Fleny113: You actually get an error for this:

```
Interface 'DiscordVanityUrl' incorrectly extends interface 'Partial<DiscordInvite>'.
  Types of property 'code' are incompatible.
    Type 'string | null' is not assignable to type 'string | undefined'.
      Type 'null' is not assignable to type 'string | undefined'.ts(2430)
```

packages/bot/src/transformers/reverse/member.ts
Fleny113: https://discord.com/developers/docs/resources/guild#guild-member-object-guild-member-structure

deaf and mute are not optional, this is one of the (breaking) changes in the pr, and since we have an undefined from the `?.` on toggles we need to default the values

packages/types/src/discord/guild.ts
Fleny113: I think i will remove the Discord prefixed json params in another pr, so for now the TODO stays

packages/types/src/discord/guild.ts
Fleny113: > The docs hasn't updated yet. Should we keep it as-is or add a `@deprecated` note?

`owner` is stil used for OAuth2 so we need to keep it, we can add the note however

packages/types/src/discord/guild.ts
lts20050703: Opps my bad

PR #4296, opened 2025-08-04
fix(types)!: Sort & fix sticker.ts types

Fleny113:


TriForMine (dismissed):
LGTM!

packages/types/src/discord/sticker.ts
```diff
@@ -11,7 +11,7 @@ export interface DiscordSticker {
   /** Name of the sticker */
   name: string
   /** Description of the sticker */
-  description: string
+  description: string | null
   /** a unicode emoji representing the sticker's expression */
   tags: string
```
lts20050703: https://discord.com/developers/docs/resources/sticker#sticker-resource

```suggestion
  /**
   * Autocomplete/suggestion tags for the sticker
   * @remarks
   * Max 200 characters
   * A comma separated list of keywords is the format used in this field by standard stickers, but this is just a convention. Incidentally the client will always use a name generated from an emoji as the value of this field when creating or modifying a guild sticker.
   */
  tags: string
```

PR #4301, opened 2025-08-05
fix(types)!: Split and fix discordeno.ts

Fleny113:


packages/rest/src/types.ts
```diff
@@ -1806,6 +1806,7 @@ export interface RestManager {
    * @see {@link https://discord.com/developers/docs/interactions/application-commands#get-global-application-command}
    */
   getGlobalApplicationCommand: (commandId: BigString) => Promise<Camelize<DiscordApplicationCommand>>
+  // TODO: Add with_localizations query param
```
lts20050703, line 1809: Is this TODO needs to be fix in this PR, or will it be fixed in another PR?

packages/rest/src/types.ts
```diff
@@ -1847,6 +1848,7 @@ export interface RestManager {
    * @see {@link https://discord.com/developers/docs/interactions/application-commands#get-guild-application-command}
    */
   getGuildApplicationCommand: (commandId: BigString, guildId: BigString) => Promise<Camelize<DiscordApplicationCommand>>
+  // TODO: Add the with_localizations query param
```
lts20050703, line 1851: Is this TODO needs to be fix in this PR, or will it be fixed in another PR?

packages/types/src/discordeno/application.ts
```diff
@@ -0,0 +1,55 @@
...
+
+import type {
+  ApplicationFlags,
+  DiscordApplicationEventWebhookStatus,
+  DiscordApplicationIntegrationType,
+  DiscordInstallParams,
+} from '../discord/application.js'
+import type { DiscordWebhookEventType } from '../discord/webhookEvents.js'
+
+// TODO: Do we want to move the "All parameters to this endpoint are optional" of this to the rest manager itself?
```
lts20050703, line 11: What do you mean by this? Like making all params required (like the docs table), then in rest manager change to `Partial<EditApplication>`?

Personally I would prefer to keep them all optional, to be consistent with other types.

packages/types/src/discordeno/application.ts
```diff
@@ -0,0 +1,55 @@
...
+  roleConnectionsVerificationUrl?: string
+  /** Settings for the app's default in-app authorization link, if enabled */
+  installParams?: DiscordInstallParams
+  /** Default scopes and permissions for each supported installation context. */
+  integrationTypesConfig?: DiscordApplicationIntegrationType
+  /**
+   * App's public flags
+   *
+   * @remarks
+   * Only limited intent flags (`GATEWAY_PRESENCE_LIMITED`, `GATEWAY_GUILD_MEMBERS_LIMITED`, and `GATEWAY_MESSAGE_CONTENT_LIMITED`) can be updated via the API.
```
lts20050703, line 28:
```suggestion
   * Only limited intent flags ({@link ApplicationFlags.GatewayPresenceLimited | GatewayPresenceLimited}, {@link ApplicationFlags.GatewayGuildMembersLimited | GatewayGuildMembersLimited}, and {@link ApplicationFlags.GatewayMessageContentLimited | GatewayMessageContentLimited}) can be updated via the API.
```

packages/types/src/discordeno/reference.ts
lts20050703, line 1: Can't seem to find FileContent in discord API docs, is this a Discordeno thing?

packages/types/src/discordeno/user.ts
```diff
@@ -0,0 +1,23 @@
...
+  withCounts?: boolean
+}
+
+/** https://discord.com/developers/docs/resources/user#create-group-dm-json-params */
+export interface GetGroupDmOptions {
+  /** Access tokens of users that have granted your app the `gdm.join` scope */
+  accessTokens: string[]
+  /** A mapping of user ids to their respective nicknames */
+  nicks?: Record<string, string>
+}
```
lts20050703, line 23:
```suggestion
/** https://discord.com/developers/docs/resources/user#create-group-dm-json-params */
export interface CreateGroupDmOptions {
  /** Access tokens of users that have granted your app the `gdm.join` scope */
  accessTokens: string[]
  /** A mapping of user ids to their respective nicknames */
  nicks: Record<string, string>
}
```

packages/types/src/discordeno/voice.ts
```diff
@@ -0,0 +1,23 @@
+/** Types for: https://discord.com/developers/docs/resources/voice */
+
+import type { BigString } from '../shared.js'
+
+/** https://discord.com/developers/docs/resources/voice#modify-current-user-voice-state-json-params */
+export interface EditOwnVoiceState {
+  /** The id of the channel the user is currently in */
+  channelId: BigString
```
lts20050703, line 8:
```suggestion
  channelId?: BigString
```

lts20050703 (commented):
Part 1.
Please don't blindly commit all of these suggestions, I might be wrong or some changes are unnecessary.

packages/rest/src/types.ts
Fleny113: another pr, this pr is already big as is there is no need to add even more stuff involving rest/manager.ts, rest/types.ts, rest/routes/types.ts and rest/routes.ts (these 4 files or less based on what is actually needed)

packages/rest/src/types.ts
Fleny113: See other conv https://github.com/discordeno/discordeno/pull/4301#discussion_r2272518553

packages/types/src/discordeno/application.ts
Fleny113: Yes, the reason i was considering it was because it is tecnically a "separate" part of the docs for the endpoint the note and the type itself

but it is fine keeping the optional in the type as well (will remove the todo(s) about this in this pr)

packages/types/src/discordeno/application.ts
Fleny113: As a note (this is valid for all of these kind of changes) i need to check if typedoc can actually resolve these links, because we currently also have a bunch of unresolved links in the rest manager as well and i can't figure out how to make `@import` work (skill issue provably, ngl)

packages/types/src/discordeno/entitlement.ts
```diff
@@ -0,0 +1,42 @@
...
+  guildId?: BigString
+  /** Whether or not ended entitlements should be omitted. Defaults to false, ended entitlements are included by default. */
+  excludeEnded?: boolean
+  /** Whether or not deleted entitlements should be omitted. Defaults to true, deleted entitlements are not included by default. */
+  excludeDeleted?: boolean
+}
+
+// TODO: This should provably get renamed to CreateTestEntitlement
+/** https://discord.com/developers/docs/monetization/entitlements#create-test-entitlement-json-params */
+export interface CreateEntitlement {
```
Fleny113, line 27: Provably it is better for another pr just because this one is big as is, and a rename like this would get lost really easly

packages/types/src/discordeno/reference.ts
Fleny113: Yes, if you want a point of reference for FileContent: https://discord.com/developers/docs/reference#uploading-files second paragraph

packages/types/src/discordeno/application.ts
lts20050703: @Fleny113 Why do we need `@import` if the comment is referencing a already imported type needed to annotate fields?

For examples, this comment links to `ApplicationFlags`, which is already imported to be used in `flags: ApplicationFlags`

If for some reason we're linking to something that's not already being imported, then just add an `@import` at the top like this?

```ts
/** @import { SomeType } from "some-module" */
```

(Example taken from https://typedoc.org/documents/Tags._import.html)

packages/rest/src/types.ts
lts20050703: Opened an issue #4340

packages/rest/src/types.ts
lts20050703: Opened an issue #4340

packages/types/src/discordeno/entitlement.ts
lts20050703: Created an issue #4341

lts20050703 (commented):
Part 2.
Part 3 soon:tm:

packages/types/src/discordeno/application.ts
Fleny113: I was talking in general + i didn't check the code yet when I wrote that

I wish it was that easy, i tried a couple of times to do that with the rest manager but it never worked

packages/types/src/discordeno/application.ts
```diff
@@ -0,0 +1,56 @@
...
+/** https://discord.com/developers/docs/resources/application#edit-current-application-json-params */
+export interface EditApplication {
+  /** Default custom authorization URL for the app, if enabled */
+  customInstallUrl?: string
+  /** Description of the app */
+  description?: string
+  /** Role connection verification URL for the app */
+  roleConnectionsVerificationUrl?: string
+  /** Settings for the app's default in-app authorization link, if enabled */
+  installParams?: DiscordInstallParams
```
lts20050703, line 21:
```suggestion
  installParams?: ApplicationInstallParams
```

packages/types/src/discordeno/application.ts
```diff
@@ -0,0 +1,56 @@
...
+  /** Default scopes and permissions for each supported installation context. */
+  integrationTypesConfig?: DiscordApplicationIntegrationType
```
lts20050703, line 23:
```suggestion
  /** Default scopes and permissions for each supported installation context. */
  integrationTypesConfig?: Partial<Record<DiscordApplicationIntegrationType, ApplicationIntegrationTypeConfiguration>>
```

packages/types/src/discordeno/user.ts
```diff
@@ -0,0 +1,23 @@
...
+  /** Get guilds after this guild ID */
+  after?: BigString
+  /** Maximum number of entries (between 1-200) to return, defaults to 200 */
+  limit?: number
+  /** Include approximate member and presence counts in response, defaults to false */
+  withCounts?: boolean
+}
+
+/** https://discord.com/developers/docs/resources/user#create-group-dm-json-params */
+export interface GetGroupDmOptions {
```
lts20050703, line 18:
```suggestion
/** https://discord.com/developers/docs/resources/user#create-group-dm-json-params */
export interface CreateGroupDmOptions {
```

lts20050703 (dismissed):
I think this is getting out of the scope of this PR. Feel free to
- Revert all commenting changes made prior (to keep the commit history tree looks clean)
- Ignore the following commenting changes

I will make another PR to address #4333

PR #4302, opened 2025-08-05
fix(bot)!: Remove InternalBot, add messagePin transformer

Fleny113:
- InternalBot has created more issues than it solved, so found a way to make transformers "happy" with the plain Bot type we now don't need it anymore 
- The messagePin transformer was declared but never put in the bot.transformers and TS did not error beause it was suppressed by the `as` we had in `createTransformers`
- Now interaction.data.resolved uses `InteractionResolvedDataChannel` and `InteractionResolvedDataMember` to properly support desired properties


TriForMine (dismissed):
LGTM!

PR #4305, opened 2025-08-08
fix(rest): Rewrite rest.simplifyUrl

Fleny113:
A way to break it currently is to spam a few reactions and it will cause 429

### DIFF
--- a/packages/rest/src/manager.ts
+++ b/packages/rest/src/manager.ts
@@ -518,30 +518,70 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
     },
 
     simplifyUrl(url, method) {
-      const parts = url.split('/')
-      const secondLastPart = parts[parts.length - 2]
+      const routeInformationKey: string[] = [method]
 
-      if (secondLastPart === 'channels' || secondLastPart === 'guilds') {
-        return url
+      const queryParamIndex = url.indexOf('?')
+      const route = queryParamIndex !== -1 ? url.slice(0, queryParamIndex) : url
+
+      // Since the urls start with / the first part will always be empty
+      const splittedRoute = route.split('/')
+
+      // 1) Strip the minor params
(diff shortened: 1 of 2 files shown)


packages/rest/src/manager.ts
```diff
@@ -518,30 +518,70 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
...
+          const isNumber = Number.isFinite(parseInt(part, 10))
+
+          if (!isNumber) {
+            // Reactions emoji need to be stripped as it is a minor parameter
+            if (index >= 1 && array[index - 1] === 'reactions') return 'x'
+            // Check if we are on a webhook token, as it is a major parameter
+            if (index >= 2 && array[index - 2] === 'webhooks') return part
+
+            // If it is part of the route keep it
+            return part
```
lts20050703:
```suggestion
            // If we are on a webhook or if it is part of the route, keep it as it is a major parameter
            return part
```

packages/rest/src/manager.ts
```diff
@@ -518,30 +518,70 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
...
+
+      // 2) Account for exceptions
+      //    - https://github.com/discord/discord-api-docs/issues/1092
+      //    - https://github.com/discord/discord-api-docs/issues/1295
+
+      // The 2 exceptions are for message delete, so we need to check if we are in that route
+      if (method === 'DELETE' && splittedRoute.length === 5 && splittedRoute[1] === 'channels' && strippedRoute.endsWith('/messages/x')) {
+        const messageId = BigInt(splittedRoute[4])
+
+        // TODO: We should move the utility from @discordeno/bot to @discordeno/utils and use that here
```
lts20050703, line 563: Created #4344

lts20050703 (changes requested):

lts20050703 (approved):
LGTM!

PR #4319, opened 2025-08-12
ci: Use biome ci in lib-check format

Fleny113:
This will allow biome to actually show if there are any format/lint errors as right know unless you go search in the logs they won't get reported

### DIFF
--- a/.github/workflows/lib-check.yml
+++ b/.github/workflows/lib-check.yml
@@ -59,20 +59,13 @@ jobs:
           node-version: 20
           cache: yarn
       - run: yarn install --immutable
-      - name: Turbo Cache
-        id: turbo-cache
-        uses: actions/cache@v4
-        with:
-          path: .turbo
-          key: ${{ runner.os }}-turbo-lint-${{ github.sha }}
       - name: Build type cache
-        if: steps.turbo-cache.outputs.cache-hit != 'true'
         uses: actions/cache@v4
         with:
           path: .turbo
           key: ${{ runner.os }}-turbo-build:type-${{ github.sha }}
       - name: Check Formatting
-        run: yarn lint --cache-dir=".turbo"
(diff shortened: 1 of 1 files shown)


.github/workflows/lib-check.yml
```diff
@@ -59,20 +59,13 @@ jobs:
...
-          key: ${{ runner.os }}-turbo-lint-${{ github.sha }}
       - name: Build type cache
-        if: steps.turbo-cache.outputs.cache-hit != 'true'
         uses: actions/cache@v4
         with:
           path: .turbo
           key: ${{ runner.os }}-turbo-build:type-${{ github.sha }}
       - name: Check Formatting
-        run: yarn lint --cache-dir=".turbo"
+        run: yarn biome ci --reporter=github
```
lts20050703: Why did you remove the turbo cache?

```suggestion
      - name: Turbo Cache
        id: turbo-cache
        uses: actions/cache@v4
        with:
          path: .turbo
          key: ${{ runner.os }}-turbo-lint-${{ github.sha }}
      - name: Build type cache
        if: steps.turbo-cache.outputs.cache-hit != 'true'
        uses: actions/cache@v4
        with:
          path: .turbo
          key: ${{ runner.os }}-turbo-build:type-${{ github.sha }}
      - name: Check Formatting
        run: yarn biome ci --reporter=github --cache-dir=".turbo"
```

Fleny113: Because there is no point in restoring the lint cache if we are not running the lint turbo task (we running biome directly without turbo being involved)

The reason the build:type cache is restored is because biome will/can use that for some type-aware lint rules (idk we have them enabled however)
(also i just know realize that i did not added the call for `yarn build:type`)

lts20050703 (dismissed):
LGTM!

lts20050703 (approved):

PR #4320, opened 2025-08-12
chore: use `biome check` instead of separate lint and format commands

Fleny113:
Biome supports running both formatting and linting in a single command,
so we can simplify the scripts, replacing separate fmt and lint scripts
with a single check script.

This will also start executing the code assists (import sorting) as they
only run when `biome check` is executed.


.github/workflows/lib-check.yml
```diff
@@ -72,7 +72,7 @@ jobs:
           path: .turbo
           key: ${{ runner.os }}-turbo-build:type-${{ github.sha }}
       - name: Check Formatting
-        run: yarn lint --cache-dir=".turbo"
+        run: yarn check:all
```
lts20050703, line 75: Don't we still need to tell yarn the cache dir?

```suggestion
        run: yarn check:all --cache-dir=".turbo"
```

README.md
```diff
@@ -132,14 +132,11 @@ yarn test:unit
 Other useful information are available on the [website under the contributing documentation](https://discordeno.js.org/docs/contributing)
 
 Other useful scripts
-(if you run in the package's directory, you need build dist before for test and types before for lint/fmt. Running in root directory should automatically do it for you)
+(if you run in the package's directory, you need build dist before for test and types before for check. Running in root directory should automatically do it for you)
```
lts20050703:
```suggestion
(if you run in a package's directory, you need to build dist before test and build types before check. Running in root directory should automatically do it for you)
```

website/docs/contributing.md
```diff
@@ -125,15 +125,15 @@ This will invoke the Typescript compiler (tsc) to compile the type definitions (
...
+This will do all the checking, linting and formatting to make sure your code follows the linting and formatting rules in place.
 
-This will run every time you make a commit. If you disable this you should still run it manually before creating the pr
+Every time you make a commit a check will also run. If you disable this you should still run it manually before creating the pr
 
-#### `yarn lint`
+#### `yarn check:all`
 
-This will do all the checking and linting to make sure your code is in a useable state. There is also one other key thing to note about this. When you change something in the `@discordeno/types` package, you can run `yarn lint` to rebuild all the types.
+This is a script that only exists in the root of the repository, it will check all the files in repo including examples, website, etc
```
lts20050703:
```suggestion
This is a script that only exists in the root of the repository. It will check all the files in repo including examples, website, etc
```

.github/workflows/lib-check.yml
Fleny113: check:all is not a turbo task

lts20050703: Oh my bad, got it.

PR #4322, opened 2025-08-12
chore: Remove `bench/` folder

Fleny113:
This is not the actual benchmark we have (and we don't even run the one we do have)
instead this is a Deno version of the benchmark and i can't find anything
that uses this benchmark

This is also somewhat annoying to keep as something the Typescript Server
may decide to typecheck those files and it will complain no stop because
it can't find what `Deno` is and doesn't understand importing urls


lts20050703 (approved):
LGTM!

PR #4324, opened 2025-08-12
build(deps): Update fastify in docker-apps/rest-passthrough

Fleny113:


Fleny113:
> What happened to the dependabot PRs?

we set them to once a month and since there is the limit of 5 prs at a time it can't keep up

we might want to increase the frequency again

lts20050703:
@Fleny113 But the limit isn't reached? Right now there's only 1 open PR? https://github.com/discordeno/discordeno/pull/4279

Fleny113:
> @Fleny113 But the limit isn't reached? Right now there's only 1 open PR? #4279

check the pr at the time of 01/08 there were 5 on / and a few on /docker-apps/rest-passthrough

lts20050703:
Oh it's once a month, I missed that part, opps
I think we should increase the frequency.

Fleny113:
https://github.com/discordeno/discordeno/blob/main/.github/dependabot.yml#L12

PR #4325, opened 2025-08-12
build(deps-dev): Update typescript & @types/node

Fleny113:


packages/utils/tests/base64.spec.ts
```diff
@@ -9,14 +9,14 @@ describe('base64.ts', () => {
...
+      expect(encode(Uint8Array.from([77, 97, 110, 32, 208, 129, 240, 164, 173]).buffer)).to.be.equal('TWFuINCB8KSt')
+      expect(encode(Uint8Array.from([77, 97, 110, 32, 208, 129, 240, 164, 173, 162, 63]).buffer)).to.be.equal('TWFuINCB8KStoj8=')
     })
     it('can encode Buffer to base64', () => {
-      expect(encode(Buffer.from([77, 97, 110, 32, 208, 129, 240, 164, 173, 162]))).to.be.equal('TWFuINCB8KStog==')
-      expect(encode(Buffer.from([77, 97, 110, 32, 208, 129, 240, 164, 173]))).to.be.equal('TWFuINCB8KSt')
-      expect(encode(Buffer.from([77, 97, 110, 32, 208, 129, 240, 164, 173, 162, 63]))).to.be.equal('TWFuINCB8KStoj8=')
+      expect(encode(Uint8Array.from(Buffer.from([77, 97, 110, 32, 208, 129, 240, 164, 173, 162])).buffer)).to.be.equal('TWFuINCB8KStog==')
+      expect(encode(Uint8Array.from(Buffer.from([77, 97, 110, 32, 208, 129, 240, 164, 173])).buffer)).to.be.equal('TWFuINCB8KSt')
+      expect(encode(Uint8Array.from(Buffer.from([77, 97, 110, 32, 208, 129, 240, 164, 173, 162, 63])).buffer)).to.be.equal('TWFuINCB8KStoj8=')
```
lts20050703: Why did you change `Buffer.from(array)` to `Uint8Array.from(Buffer.from(array))`? The description of the unit test is "can encode Buffer to base64", so I don't know why you change the code to "can encode Uint8Array to base64", which is already covered by the previous unit test

```suggestion
      expect(encode(Buffer.from([77, 97, 110, 32, 208, 129, 240, 164, 173, 162]))).to.be.equal('TWFuINCB8KStog==')
      expect(encode(Buffer.from([77, 97, 110, 32, 208, 129, 240, 164, 173]))).to.be.equal('TWFuINCB8KSt')
      expect(encode(Buffer.from([77, 97, 110, 32, 208, 129, 240, 164, 173, 162, 63]))).to.be.equal('TWFuINCB8KStoj8=')
```

packages/utils/tests/base64.spec.ts
```diff
@@ -9,14 +9,14 @@ describe('base64.ts', () => {
       expect(encode('Man Ё𤭢')).to.be.equal('TWFuINCB8KStog==')
     })
     it('can encode Uint8Array to base64', () => {
-      expect(encode(new Uint8Array([77, 97, 110, 32, 208, 129, 240, 164, 173, 162]))).to.be.equal('TWFuINCB8KStog==')
-      expect(encode(new Uint8Array([77, 97, 110, 32, 208, 129, 240, 164, 173]))).to.be.equal('TWFuINCB8KSt')
-      expect(encode(new Uint8Array([77, 97, 110, 32, 208, 129, 240, 164, 173, 162, 63]))).to.be.equal('TWFuINCB8KStoj8=')
+      expect(encode(Uint8Array.from([77, 97, 110, 32, 208, 129, 240, 164, 173, 162]).buffer)).to.be.equal('TWFuINCB8KStog==')
+      expect(encode(Uint8Array.from([77, 97, 110, 32, 208, 129, 240, 164, 173]).buffer)).to.be.equal('TWFuINCB8KSt')
+      expect(encode(Uint8Array.from([77, 97, 110, 32, 208, 129, 240, 164, 173, 162, 63]).buffer)).to.be.equal('TWFuINCB8KStoj8=')
```
lts20050703: Why this change? What's the difference between `new Uint8Array(array)` vs `Uint8Array.from(array)`?

packages/utils/tests/base64.spec.ts
Fleny113: Typescript 5.9 changes some types for Buffer, now they don't extend Uint8Array anymore, so encode requiring `ArrayBuffer` creates a type error if you pass a Buffer to it

So, to keep the test we need to convert from a Buffer from Uint8Array (we could remove the test ig, it is the same as the Uint8Array one)

https://devblogs.microsoft.com/typescript/announcing-typescript-5-9/#notable-behavioral-changes

packages/utils/tests/base64.spec.ts
Fleny113: i don't remember exactly, `new Uint8Array(iterable)` and `TypedArray.from(iterable)` (`from` is inherited from TypedArray) seem to be basically the same. I can revert the change ig
(the .buffer is required due to type changes in TS 5.9 and/or @types/node)

packages/utils/tests/base64.spec.ts
lts20050703: Yeah I think we should remove that test since it's the same as the Uint8Array one (or alternatively, make `encode` support Buffer in addition to ArrayBuffer)

PR #4330, opened 2025-08-12
build(deps): Migrate to React Flow v12 and update website dependencies

Fleny113:
The bump also includes `reactflow` -> `@xyflow/react` (https://reactflow.dev/learn/troubleshooting/migrate-to-v12)

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -17,31 +17,31 @@
     "format": "biome format --write"
   },
   "dependencies": {
-    "@docusaurus/core": "^3.8.0",
-    "@docusaurus/faster": "^3.8.0",
-    "@docusaurus/plugin-client-redirects": "^3.8.0",
-    "@docusaurus/plugin-content-docs": "^3.8.0",
-    "@docusaurus/preset-classic": "^3.8.0",
-    "@docusaurus/theme-common": "^3.8.0",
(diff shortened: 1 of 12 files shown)


Fleny113:
> This PR should be more like "Migrate to React Flow v12 and update website dependencies"

makes sense, i did not change the title bc i had to update to react flow v12 bc of type errors with react flow v11 (i did not know there was an update in the first place in the beginning)

lts20050703 (commented):
1. Why was some `JSX.Element` return type annotations being removed? Just change them to `React.JSX.Element?`
2. Why was the logic for setting the width being removed?

Fleny113:
> 1. Why was some `JSX.Element` return type annotations being removed

Typescript knows best™️

Jokes aside, it was either removing it or using `React.JSX.Element` or else you get type errors. I think letting Typescript infer the type it needs is better

> 2. Why was the logic for setting the width being removed?

It was overflowing into the table of content on the side, ik a bit out of scope for the pr (the color change was needed as that was broken due to the update, the overflow exists in discordeno.js.org too)

Letting the browser & xyflow (reactflow) deal with resizes themself seem like a better idea to me

lts20050703 (approved):
LGTM!

PR #4334, opened 2025-08-13
feat(gateway)!: Rename shard.events.identified to shard.events.ready

Fleny113:


lts20050703:
Oh my bad. Why move the code to the ws.onmessage?

packages/gateway/src/types.ts
```diff
@@ -145,7 +145,7 @@ export interface ShardEvents {
   /** The shard has started to identify itself to Discord. */
   identifying?: (shard: Shard) => unknown
   /** The shard has successfully been identified itself with Discord. */
```
lts20050703, line 147: Actually should we update this comment or keep it detailed

Fleny113: i mean, it does make sense, the alternative is something along the lines of "The shard has received a READY event."

PR #4335, opened 2025-08-13
feat(bot): Add guildId to ThreadMember

Fleny113:
This uses an additional `extra` parameter in the transformer, this is inconsistent with other transformers, as some use a custom object as the payload and in there we have the discord payload + the additional data we may need, like guildId in channel to make an example, while others add additional parameters like the member transformer that has 2 additional parameters for the guild id and the user id.

We should unity the approch we have to pass additional parameters to transformers and customizers, so for this PR an `extra` parameter is used and this edit will be added to the other transformers in another PR after https://github.com/discordeno/discordeno/pull/4302 is merged as that pr helps with allowing proper typechecking of the transformers and customizers, along side not requiring the InternalBot "hack" to make everything work.

### DIFF
--- a/packages/bot/src/helpers.ts
+++ b/packages/bot/src/helpers.ts
@@ -138,6 +138,7 @@ import type {
   Subscription,
   Template,
   ThreadMember,
+  ThreadMemberTransformerExtra,
   User,
   VoiceState,
   Webhook,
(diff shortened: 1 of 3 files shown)


lts20050703 (changes requested):
https://discord.com/developers/docs/resources/channel#thread-member-object-thread-member-structure

Why do we need to add `guildId` to `ThreadMember`?

Fleny113:
> Why do we need to add `guildId` to `ThreadMember`?

1- We have guildId on member as well, it makes sense to associate a thread member (and a member) to the guild
2- ask @AwesomeStickz idr exactly but there were a few reasons

lts20050703 (dismissed):
LGTM!

lts20050703 (approved):
LGTM! I like the extra parameter instead of hacking the options :)

PR #4363, opened 2025-08-15
fix(types): Fix discordeno/reference.ts

Fleny113:


packages/types/src/discordeno/reference.ts
```diff
@@ -1,5 +1,15 @@
 /** Types for: https://discord.com/developers/docs/reference */
 
+/**
+ * This is not a mapping of any discord documented type, however the is some docs that refer to an object like this:
```
lts20050703:
```suggestion
 * This is not a mapping of any discord documented type, however there is some docs that refer to an object like this:
```

packages/types/src/discordeno/reference.ts
```diff
@@ -1,5 +1,15 @@
...
+ * https://discord.com/developers/docs/reference#uploading-files, the second paragraph
+ *
+ * > All files[n] parameters must include a valid `Content-Disposition` subpart header with a filename and unique name parameter.
+ *
+ * The blob maps to the FormData data sent and the name to the FormData file name.
```
lts20050703:
```suggestion
 * The blob maps to the FormData data sent and the name maps to the FormData file name.
```

lts20050703 (changes requested):

lts20050703 (approved):
LGTM!

PR #4365, opened 2025-08-15
fix(types)!: Fix discordeno/user.ts

Fleny113:


packages/bot/src/helpers.ts
```diff
@@ -976,7 +976,7 @@ export type BotHelpers<TProps extends TransformersDesiredProperties, TBehavior e
   getChannels: (guildId: BigString) => Promise<SetupDesiredProps<Channel, TProps, TBehavior>[]>
   getChannelWebhooks: (channelId: BigString) => Promise<SetupDesiredProps<Webhook, TProps, TBehavior>[]>
   getDmChannel: (userId: BigString) => Promise<SetupDesiredProps<Channel, TProps, TBehavior>>
-  getGroupDmChannel: (options: GetGroupDmOptions) => Promise<SetupDesiredProps<Channel, TProps, TBehavior>>
+  getGroupDmChannel: (options: CreateGroupDmOptions) => Promise<SetupDesiredProps<Channel, TProps, TBehavior>>
```
lts20050703, line 979: Should we rename `getDmChannel` and `getGroupDmChannel` to `createDm` and `createGroupDm` to make it consistent with the docs? It's kinda jarring that the function is called `get`, but the options interface (and discord docs) are `create`

Fleny113: we could and imo we should[^1], but since there are a lot of these (for example sendMessage should be [createMessage](https://discord.com/developers/docs/resources/message#create-message)) we might want to do a separate pr

[^1]: By should i especially mean that, beside `@discordeno/bot` that does differ a bit from the Discord API, we usually match the api docs names, so for consistency we should use what discord documents as the name

lts20050703: Opened #4383

lts20050703 (approved):
LGTM!

PR #4366, opened 2025-08-15
fix(types): Fix discordeno/voice.ts

Fleny113:


packages/types/src/discordeno/voice.ts
```diff
@@ -18,6 +18,8 @@ export interface EditUserVoiceState {
   channelId?: BigString
   /** Toggles the user's suppress state */
   suppress?: boolean
+
+  // This is part of the route but it is included in here
   /** The user id to target */
   userId: BigString
```
lts20050703, line 24: Can we move this to be part of the helper param? Since that would be more consistent with other helpers

lts20050703 (changes requested):

packages/types/src/discordeno/voice.ts
Fleny113: to be fair i want to do this to all the helpers / rest endpoints, as a route is not an option, however i wasn't sure if doing it in these pr(s) or doing a separate pr to move the route params where necessary

lts20050703: @Fleny113 I think there are two options, either rename this PR or close and create a new one that move the route from options to helpers params.

Fleny113: Then we close this one and i'll to another for the route things

PR #4373, opened 2025-08-16
feat(gateway): Use `node:zlib` zstd when available

Fleny113:
Currently we rely on `fzstd` for the zstd-stream support, node also added support for this in v22.15 and v23.8

https://nodejs.org/docs/latest/api/zlib.html#class-zlibzstddecompress

The docs do mark this as "Experimental" however fzstd is also at it's 0.1 version, and by my testing node:zlib zstd works fine

### DIFF
--- a/packages/gateway/src/Shard.ts
+++ b/packages/gateway/src/Shard.ts
@@ -1,9 +1,9 @@
 import { Buffer } from 'node:buffer'
-import { createInflate, type Inflate, inflateSync, constants as zlibConstants } from 'node:zlib'
+import zlib from 'node:zlib'
 import type { DiscordGatewayPayload, DiscordHello, DiscordReady, DiscordUpdatePresence } from '@discordeno/types'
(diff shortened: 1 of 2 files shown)


packages/gateway/src/Shard.ts
```diff
@@ -211,19 +211,51 @@ export class DiscordenoShard {
...
-            const parsedData = JSON.parse(decodedData)
-            this.decompressionPromisesQueue.shift()?.(parsedData)
+              return
+            }
+
+            this.inflateBuffer = data
+          })
+        } else {
+          this.logger.debug('[Shard] Attempting to use fzstd zstd decompression.')
+
```
lts20050703: I don't think this is necessary, like in mind there should only be three path: log using node:zlib, log using fzstd, or log disabling compression

```suggestion
```

lts20050703 (approved):
LGTM!

lts20050703 (approved):
LGTM!

PR #4374, opened 2025-08-16
fix(bot): Remove handlers payload checks

### LINKED ISSUE #4371: channelDelete event is not called if guild_id does not exist
Is there a reason we check if data.d.guild_id exists before calling the channelDelete event?

https://github.com/discordeno/discordeno/blob/main/packages/bot/src/handlers/channels/CHANNEL_DELETE.ts#L8

This seems to originate from a 4 yrs ago commit: https://github.com/discordeno/discordeno/commit/2427a5d2c88cea5f69ff2aac33e0c6c685500f2d

Fleny113:
To remove the check from VOICE_STATE_UPDATE we need to wait for #4372 to get merged

fixes #4371


lts20050703 (approved):
LGTM!

PR #4375, opened 2025-08-16
chore(rest): Remove todo about GET bodies

Fleny113:
We do not need to do anythng with GET bodies, fetch() will even throw if a body is passed to a GET


lts20050703 (approved):
LGTM!

PR #4381, opened 2025-08-17
feat(bot)!: Make desiredProperties required

Fleny113:
Allowing desired properties to be omitted makes it harder to get started with discordeno as you won't know you need them unless you read the docs


lts20050703 (approved):
LGTM!

PR #4402, opened 2025-08-26
refactor: Use TransformerFunction to define transformers

Fleny113:
This allows us to edit the transformers in bulk without many issues


lts20050703 (approved):
LGTM!

PR #4436, opened 2025-09-05
feat(transformers)!: Partial transformers

### LINKED ISSUE #3613: [Feat]: Use partials in transformers
### Tell us what problem the feature would solve

The transformers that have desired proprieties mostly can deal with `Partial<T>` variants of the payload. Since we use `Partial<T>` in `@discordeno/types` in some place where discord explicitly writes in the documentation that the object will be partial those objects still need to be transformed but TS will give an error in those situations because a `Partial<T>` type is not compatible with `T` itself.

### Describe the solution you'd like

For transformers that do support it at least use the `Partial<T>` version of the payload to avoid some typescript errors on this

### Describe alternatives you've considered

In the code we currently have some `// @ts-expect-error` where we use the transformers, however it MAY be easier to change the type of the transformers

### Additional Info

_No response_

Fleny113:
Add support for partial payloads in transformers.

> [!IMPORTANT]
> This builds on top of #4435 as it allows to simplify a bunch the code for this
>
> This should not be merged until #4435 is merged

- fixes: #3613

Explainer of the PR (i feel like this is needed):
(they are in order of how `git format-patch` is sorting the files for the most part)

- Remove all `@ts-expect-error` related to partial transformers
- Add `partial: true` extra params to transformer calls that use Partial objects
- `inviteCreate` event now no longer uses the `Invite` type
  - Reason for this change: it made it __really__ hard to deal with both objects in the invite transformer
- Other events now propertly have Partial objects when they are needed
- `handlers/invites/INVITE_DELETE.ts` no longer has comments when it creates the object, they don't have a purpose
- `transformers.ts`:
  - We no longer declare transformer **functions** manually, this caused issue with inferring the customizers, they are now inferred from the transformer informations, see below
  - We instead declare transformer **informations** manually, this allows us to not relying on typescript maybe inferring stuff correctly from the transformer functions
  - All transformer informations need to declare if they support partials (third generic param)
  - Add `TransformerInformation` type helper
  - Rewrite `TransformerFunction`, `TransformerFunctions`, `TransformerCustomizerFunction`, `TransformerCustomizers` type helpers and add comments to these describing what they are
  - add `callCustomizer`, this function is used as a wrapper for bot.transformers.customizers.XYZ, the reason for this is because it would be quite a lot of code to add in each transformer to make Typescript recognise that we aren't doing anything wrong with the function call, so it is wrapped for partial transformers
- Transformers no longer have an explicit return type: the return type is wrong as it doesn't account for neither partials nor desired props, so it is somewhat better to simply remove it
- The component transformer now fails if it doesn't receive a component type, the reason for this is because the type should always exist or else we don't know what component are we trying to transform
- The guild transformer now uses the discord payload to get the id for collections
- The invite transformer no longer handles `DiscordInviteCreate` (see `inviteCreate` event above for details)
- The toggle classes now use Partial<...> for the object part of the union they accept
- `Invite` transformed object now correctly has partials, `targetType` is now of type `TargetTypes`, `uses` comment was fixed
- `isInviteWithMetadata` type guard was removed, it is no longer needed


TriForMine (dismissed):

packages/bot/src/transformers/role.ts
```diff
@@ -46,20 +47,19 @@ export const baseRole: Role = {
...
+export function transformRole(bot: Bot, payload: Partial<DiscordRole>, extra?: { guildId?: BigString; partial?: boolean }) {
   const role: SetupDesiredProps<Role, TransformersDesiredProperties, DesiredPropertiesBehavior> = Object.create(baseRole);
   const props = bot.transformers.desiredProperties.role;
   if (props.id && payload.id) role.id = bot.transformers.snowflake(payload.id);
   // Role name can be an empty string
   if (props.name && payload.name !== undefined) role.name = payload.name;
-  if (props.position) role.position = payload.position;
+  if (props.position && payload.position !== undefined) role.position = payload.position;
   if (props.guildId && extra?.guildId) role.guildId = bot.transformers.snowflake(extra?.guildId);
   if (props.color && payload.color !== undefined) role.color = payload.color;
```
AwesomeStickz, line 58:
```suggestion
  if (props.color && payload.color !== undefined) role.color = payload.color;
  if (props.colors && payload.colors) role.colors = bot.transformers.roleColors(bot, payload.colors);
```

packages/bot/src/transformers/types.ts
```diff
@@ -1207,14 +1208,14 @@ export interface Invite {
   /** The maximum number of times the invite can be used */
   maxUses: number;
   /** The type of target for this voice channel invite */
-  targetType: number;
+  targetType: TargetTypes;
```
AwesomeStickz: Should this be optional? https://docs.discord.com/developers/events/gateway-events#invite-create

Same for `targetUser` below. If yes, can you update everywhere

packages/bot/src/events.ts
```diff
@@ -72,7 +72,22 @@ export type EventHandlers<TProps extends TransformersDesiredProperties, TBehavio
...
+    guildId?: bigint;
+    inviter?: SetupDesiredProps<User, TProps, TBehavior>;
+    maxAge: number;
+    maxUses: number;
+    targetType: TargetTypes;
+    targetUser?: SetupDesiredProps<User, TProps, TBehavior>;
+    targetApplication?: Partial<SetupDesiredProps<Application, TProps, TBehavior>>;
+    temporary: boolean;
+    uses: number;
+    expiresAt: number;
```
AwesomeStickz:
```suggestion
    expiresAt: number | null;
```

I'm not too sure on this one

packages/bot/src/handlers/invites/INVITE_CREATE.ts
```diff
@@ -6,6 +6,22 @@ export async function handleInviteCreate(bot: Bot, data: DiscordGatewayPayload,
...
+    maxAge: payload.max_age,
+    maxUses: payload.max_uses,
+    targetType: payload.target_type,
+    targetUser: payload.target_user ? bot.transformers.user(bot, payload.target_user) : undefined,
+    targetApplication: payload.target_application
+      ? bot.transformers.application(bot, payload.target_application, { shardId, partial: true })
+      : undefined,
+    temporary: payload.temporary,
+    uses: payload.uses,
+    expiresAt: Date.parse(payload.expires_at),
```
AwesomeStickz:
```suggestion
    expiresAt: payload.expires_at ? Date.parse(payload.expires_at) : undefined,
```

I'm not sure if it should be null or undefined but change accordingly

packages/bot/src/transformers/component.ts
```diff
@@ -311,21 +336,26 @@ function transformMentionableSelectMenuComponent(
     if (props.values && payload.values) select.values = payload.values;
     if (props.resolved && payload.resolved) select.resolved = bot.transformers.interactionDataResolved(bot, payload.resolved);
   } else {
-    if (props.placeholder && payload.placeholder) select.placeholder = payload.placeholder;
-    if (props.minValues && payload.min_values) select.minValues = payload.min_values;
-    if (props.maxValues && payload.max_values) select.maxValues = payload.max_values;
-    if (props.defaultValues && payload.default_values)
-      select.defaultValues = payload.default_values.map((defaultValue) => ({
+    const _payload = payload as Partial<DiscordUserSelectComponent>;
```
AwesomeStickz:
```suggestion
    const _payload = payload as Partial<DiscordMentionableSelectComponent | DiscordMentionableSelectInteractionResponseFromModal>;
```

I don't know if it should include the `| DiscordMentionableSelectInteractionResponseFromModal` considering that you didn't do in other places but change here / other places accordingly

packages/bot/src/transformers/component.ts
```diff
@@ -281,23 +304,25 @@ function transformRoleSelectMenuComponent(bot: Bot, payload: DiscordRoleSelectCo
...
```
AwesomeStickz:
```suggestion
    const _payload = payload as Partial<DiscordRoleSelectComponent | DiscordRoleSelectInteractionResponseFromModal>;
```

packages/bot/src/transformers/component.ts
AwesomeStickz:
```suggestion
  if (props.emoji && payload.emoji) button.emoji = bot.transformers.emoji(bot, payload.emoji, { partial: true });
```

packages/bot/src/transformers/poll.ts
```diff
@@ -24,15 +25,19 @@ export function transformPoll(bot: Bot, payload: DiscordPoll): Poll {
...
+  });
 }
 
-export function transformPollMedia(bot: Bot, payload: DiscordPollMedia): PollMedia {
+export function transformPollMedia(bot: Bot, payload: Partial<DiscordPollMedia>, extra?: { partial?: boolean }) {
   const props = bot.transformers.desiredProperties.pollMedia;
   const pollMedia = {} as SetupDesiredProps<PollMedia, TransformersDesiredProperties, DesiredPropertiesBehavior>;
 
   if (props.text && payload.text) pollMedia.text = payload.text;
   if (props.emoji && payload.emoji) pollMedia.emoji = bot.transformers.emoji(bot, payload.emoji as DiscordEmoji);
```
AwesomeStickz:
```suggestion
  if (props.emoji && payload.emoji) pollMedia.emoji = bot.transformers.emoji(bot, payload.emoji, { partial: true });
```

packages/bot/src/transformers/invite.ts
```diff
@@ -1,46 +1,42 @@
...
+  if (props.maxAge && payload.max_age !== undefined) invite.maxAge = payload.max_age;
+  if (props.maxUses && payload.max_uses !== undefined) invite.maxUses = payload.max_uses;
   if (props.targetType && payload.target_type) invite.targetType = payload.target_type;
   if (props.targetUser && payload.target_user) invite.targetUser = bot.transformers.user(bot, payload.target_user);
   if (props.targetApplication && payload.target_application)
-    // @ts-expect-error TODO: Partials
-    invite.targetApplication = bot.transformers.application(bot, payload.target_application, { shardId: extra?.shardId });
+    invite.targetApplication = bot.transformers.application(bot, payload.target_application, { shardId: extra?.shardId, partial: true });
   if (props.temporary && payload.temporary) invite.temporary = payload.temporary;
   if (props.uses && payload.uses) invite.uses = payload.uses;
```
AwesomeStickz:
```suggestion
  if (props.uses && payload.uses !== undefined) invite.uses = payload.uses;
```

packages/bot/src/transformers/component.ts
```diff
@@ -39,12 +39,22 @@ import {
...
+export function transformComponent(
+  bot: Bot,
+  payload: Partial<DiscordMessageComponent | DiscordMessageComponentFromModalInteractionResponse>,
+  extra?: { partial?: boolean },
+) {
   let component: SetupDesiredProps<Component, TransformersDesiredProperties, DesiredPropertiesBehavior>;
 
+  // I don't know what we could do with a component without a type, so we just throw an error
+  if (!payload.type) {
+    throw new Error(`[Component transformer] Received component payload without a type.`);
```
AwesomeStickz:
```suggestion
    throw new Error(`[Discordeno] Received a component payload without a type in the component transformer.`);
```

packages/bot/src/transformers/component.ts
```diff
@@ -39,12 +39,22 @@ import {
...
 
-export function transformComponent(bot: Bot, payload: DiscordMessageComponent | DiscordMessageComponentFromModalInteractionResponse): Component {
+export function transformComponent(
+  bot: Bot,
+  payload: Partial<DiscordMessageComponent | DiscordMessageComponentFromModalInteractionResponse>,
+  extra?: { partial?: boolean },
+) {
   let component: SetupDesiredProps<Component, TransformersDesiredProperties, DesiredPropertiesBehavior>;
 
+  // I don't know what we could do with a component without a type, so we just throw an error
```
AwesomeStickz:
```suggestion
  // There's nothing to do with a component without a type, as it means an issue on Discord, so we just throw an error
```

packages/bot/src/transformers.ts
```diff
@@ -553,34 +545,159 @@ export function createTransformers<TProps extends TransformersDesiredProperties,
...
+ * Since the overloads are present only with partial transformers, this function requires the partial extra, for non-partial transformers the normal function call can be used
+ */
+export function callCustomizer<TInfo extends keyof TransformerInformations>(
+  customizer: TInfo,
+  bot: Bot,
+  payload: TransformerInformations[TInfo]['payload'] | Partial<TransformerInformations[TInfo]['payload']>,
+  transformed:
+    | SetupDesiredProps<TransformerInformations[TInfo]['transformed'], TransformersDesiredProperties, DesiredPropertiesBehavior>
+    | Partial<SetupDesiredProps<TransformerInformations[TInfo]['transformed'], TransformersDesiredProperties, DesiredPropertiesBehavior>>,
+  extra: TransformerInformations[typeof customizer]['extra'] & { partial: boolean },
```
AwesomeStickz:
```suggestion
  extra: TransformerInformations[TInfo]['extra'] & { partial: boolean },
```

packages/bot/src/transformers/component.ts
AwesomeStickz:
```suggestion
  // We assume that if we find 'values' it is the interaction response
```

AwesomeStickz:
```suggestion
  // We assume that if we find 'values' it is the interaction response
```

AwesomeStickz:
```suggestion
  // We assume that if we find 'values' it is the interaction response
```

AwesomeStickz:
```suggestion
  // We assume that if we find 'values' it is the interaction response
```

AwesomeStickz (changes requested):

packages/bot/src/handlers/invites/INVITE_CREATE.ts
Fleny113: I'd this should be undefined, given that we almost never use null

packages/bot/src/events.ts
Fleny113: See comment in INVITE_CREATE.ts

packages/bot/src/transformers/component.ts
Fleny113: no, DiscordRoleSelectInteractionResponseFromModal is in the if above

packages/bot/src/transformers/component.ts
Fleny113: same thing as above, the response is the if above ('values' in payload). TS can infer the type in that if, not in here because of partials, thats the reason for this type cast

packages/bot/src/handlers/invites/INVITE_CREATE.ts
AwesomeStickz: Good now?

Fleny113: i changed that myself

packages/bot/src/events.ts
AwesomeStickz: So `expiresAt?: number;`?

packages/bot/src/transformers/types.ts
```diff
@@ -1207,14 +1208,14 @@ export interface Invite {
   /** The maximum number of times the invite can be used */
   maxUses: number;
   /** The type of target for this voice channel invite */
-  targetType: number;
+  targetType?: TargetTypes;
   /** The target user for this invite */
   targetUser: User;
```
AwesomeStickz:
```suggestion
  targetUser?: User;
```

AwesomeStickz (dismissed):

packages/bot/src/transformers/interaction.ts
```diff
@@ -229,16 +222,19 @@ export function transformInteractionDataOption(bot: Bot, option: DiscordInteract
...
   if (payload.messages) {
     transformed.messages = new Collection(
       Object.entries(payload.messages).map(([key, value]) => {
-        // @ts-expect-error TODO: Deal with partials
-        const message = bot.transformers.message(bot, value, { shardId: extra?.shardId });
+        const message = bot.transformers.message(bot, value, { shardId: extra?.shardId, partial: true }) as SetupDesiredProps<
+          Message,
+          TransformersDesiredProperties,
+          DesiredPropertiesBehavior.RemoveKey
+        >;
```
TriForMine: `DiscordInteractionDataResolved.messages` permits partial Discord messages, but this cast widens the transformed value back to a complete message. A legal resolved entry can therefore be exposed as if required fields such as `id` and `author` exist when they do not.

This needs a coordinated change here and in `transformers/types.ts`. The collection member should become `messages?: Collection<bigint, Partial<Message>>` and this cast should remain partial. I compiled and tested that pair successfully.

packages/bot/src/transformers/interaction.ts
```diff
@@ -229,16 +222,19 @@ export function transformInteractionDataOption(bot: Bot, option: DiscordInteract
 
 export function transformInteractionDataResolved(
   bot: Bot,
-  payload: DiscordInteractionDataResolved,
-  extra?: { shardId?: number; guildId?: BigString },
-): TransformProperty<InteractionDataResolved, TransformersDesiredProperties, DesiredPropertiesBehavior.RemoveKey> {
-  const transformed: TransformProperty<InteractionDataResolved, TransformersDesiredProperties, DesiredPropertiesBehavior.RemoveKey> = {};
+  payload: Partial<DiscordInteractionDataResolved>,
```
TriForMine, line 224: Inside this partial resolved data transformer, resolved channels are still passed to the channel transformer without `{ partial: true }`. Their source type is a restricted channel shape, so channel customizers receive the complete branch and `extra.partial` is false. Please mark that channel call partial as well.

packages/bot/src/transformers/application.ts
```diff
@@ -1,10 +1,11 @@
-import { type DiscordApplication, DiscordApplicationIntegrationType, type DiscordUser } from '@discordeno/types';
+import { type DiscordApplication, DiscordApplicationIntegrationType } from '@discordeno/types';
 import { iconHashToBigInt } from '@discordeno/utils';
 import type { Bot } from '../bot.js';
+import { callCustomizer } from '../transformers.js';
 import { ToggleBitfield } from './toggles/ToggleBitfield.js';
 import type { Application } from './types.js';
 
-export function transformApplication(bot: Bot, payload: DiscordApplication, extra?: { shardId?: number }): Application {
+export function transformApplication(bot: Bot, payload: Partial<DiscordApplication>, extra?: { shardId?: number; partial?: boolean }) {
```
TriForMine, line 8: The mapped `bot.transformers.application` type hides two regressions in this directly exported implementation signature. Generated declarations infer `any` because the function returns `callCustomizer`, and direct callers can pass a partial payload while omitting `{ partial: true }`, which makes the customizer receive `partial: false`. Please publish overload declarations matching `TransformerFunction`, or give direct exports an equivalent conditional contract, and add a declaration test for direct imports.

packages/bot/src/transformers.ts
```diff
@@ -553,34 +545,159 @@ export function createTransformers<TProps extends TransformersDesiredProperties,
...
+        bot: Bot<TProps, TBehavior>,
+        payload: TPayload,
+        transformed: SetupDesiredProps<TTransformed, TProps, TBehavior>,
+        extra: TExtra & { partial: false },
+      ): any;
+      (
+        bot: Bot<TProps, TBehavior>,
+        payload: Partial<TPayload>,
+        transformed: Partial<SetupDesiredProps<TTransformed, TProps, TBehavior>>,
+        extra: TExtra & { partial: true },
```
TriForMine, line 640: These overloads do not preserve correlation when they contextually type an inline customizer. After checking `extra.partial`, TypeScript still treats `payload` and `transformed` as independent complete or partial unions. The complete branch still reports required properties as possibly undefined. A discriminated object or rest tuple union for the correlated arguments would allow the flag to narrow all three values. Please add a type test that exercises both branches.

packages/bot/src/transformers/component.ts
```diff
@@ -39,12 +39,22 @@ import {
...
 import type { Bot } from '../bot.js';
 import type { DesiredPropertiesBehavior, SetupDesiredProps, TransformersDesiredProperties } from '../desiredProperties.js';
+import { callCustomizer } from '../transformers.js';
 import { ToggleBitfield } from './toggles/ToggleBitfield.js';
 import type { Component, MediaGalleryItem, UnfurledMediaItem } from './types.js';
 
-export function transformComponent(bot: Bot, payload: DiscordMessageComponent | DiscordMessageComponentFromModalInteractionResponse): Component {
+export function transformComponent(
+  bot: Bot,
+  payload: Partial<DiscordMessageComponent | DiscordMessageComponentFromModalInteractionResponse>,
```
TriForMine, line 48: The runtime error below is intentional according to the merge request description, but this payload type still permits `{}` with `{ partial: true }`. The partial transformer contract needs a way to keep required discriminators such as `type` mandatory. A local intersection is not enough because the mapped `TransformerFunction` regenerates `Partial<TPayload>` and rejects the narrower implementation, so this likely belongs in the transformer metadata model.

packages/bot/src/transformers/message.ts
```diff
@@ -141,7 +142,7 @@ export const baseMessage: Message = {
   },
 };
 
-export function transformMessage(bot: Bot, payload: DiscordMessage, extra?: { shardId?: number }): Message {
+export function transformMessage(bot: Bot, payload: Partial<DiscordMessage>, extra?: { shardId?: number; partial?: boolean }) {
```
TriForMine, line 145: Nonblocking semantics question. Should a partial result preserve the distinction between a field that was not supplied and a field that was cleared. At present `{ id }` with `{ partial: true }` gains zeroed `bitfield` and `flags` plus `content: ""`, while explicit empty arrays are dropped by the length guards below. If partial results are intended to be mergeable patches, these defaults can overwrite existing state. If partial only means accepting an incomplete input, this behavior should be documented and tested.

TriForMine (changes requested):
This merge request was reviewed by an AI using GPT-5.6 Sol High.

I reviewed the implementation against the merge request description. I am not challenging the intentional removal of `DiscordInviteCreate` from the invite transformer, the removal of `isInviteWithMetadata`, or the deliberate runtime error for components without a type.

The inline suggestion is an apply ready change that I compiled and tested. Some other fixes require coordinated edits to lines that GitHub does not include in the current diff view, so those are explained in focused comments instead of unsafe partial suggestions.

The suggested edits were validated together in an isolated worktree. The declaration build, runtime build, bot source typecheck, unit tests, typecheck tests, Biome, and `git diff --check` all passed.

packages/bot/src/transformers/application.ts
Fleny113: You are not supposed to call transformers from the import. You are supposed to use bot.transformers which will be typed correctly

This is also ""enforced"" by the `Bot` type without generics being impossible to create via convential means (createBot)

packages/bot/src/transformers/component.ts
Fleny113: Because making the type enforced at the type level makes it very hard to maintain other parts of the codebase, including `@discordeno/types` which shouldn't be influenced by `@discordeno/bot` implementations. Discord documents them as partial, we use partial.

packages/bot/src/transformers/message.ts
Fleny113: semantics like this are not related. The execution side is (mostly) untouched by the PR, so our regular rules for transformers apply: If it doesn't exist, is undefined or null at runtime, it won't exist in the transformed object

packages/bot/src/events.ts
```diff
@@ -72,7 +72,22 @@ export type EventHandlers<TProps extends TransformersDesiredProperties, TBehavio
...
+    targetUser?: SetupDesiredProps<User, TProps, TBehavior>;
+    targetApplication?: Partial<SetupDesiredProps<Application, TProps, TBehavior>>;
+    temporary: boolean;
+    uses: number;
+    expiresAt?: number;
+    roleIds?: bigint[];
+  }) => unknown;
   inviteDelete: (payload: { channelId: bigint; guildId?: bigint; code: string }) => unknown;
   guildMemberAdd: (member: SetupDesiredProps<Member, TProps, TBehavior>, user: SetupDesiredProps<User, TProps, TBehavior>) => unknown;
   guildMemberRemove: (user: SetupDesiredProps<User, TProps, TBehavior>, guildId: bigint) => unknown;
```
TriForMine, line 93: The transformer now returns a partial member, so the public event type should expose a partial member as well.

```suggestion
  guildMemberRemove: (user: SetupDesiredProps<User, TProps, TBehavior>, guildId: bigint) => unknown;
  guildMemberUpdate: (member: Partial<SetupDesiredProps<Member, TProps, TBehavior>>, user: SetupDesiredProps<User, TProps, TBehavior>) => unknown;
```

packages/bot/src/transformers.ts
Fleny113: That is a typescript limitation, i can't do much about it. We still define the overloads this way to make it clear that this is the intention however

packages/bot/src/events.ts
Fleny113: This diff is wrong as applied as-is will duplicate guildMemberUpdate

AwesomeStickz (approved):

TriForMine (approved):

PR #4471, opened 2025-09-21
Updated interaction.respond and .deferEdit methods

mloetta:
Replaced `interaction.respond()` with `interaction.reply()` for cleaner code and replaced `interaction.deferEdit()` with `interaction.deferUpdate()` so it matches discord API

### DIFF
--- a/examples/advanced/src/commands/ping.ts
+++ b/examples/advanced/src/commands/ping.ts
@@ -10,6 +10,6 @@ createCommand({
 
     const embeds = createEmbeds().setTitle(`The bot ping is ${ping}ms`)
 
-    await interaction.respond({ embeds })
+    await interaction.reply({ embeds })
   },
 })
--- a/packages/bot/src/transformers/interaction.ts
+++ b/packages/bot/src/transformers/interaction.ts
@@ -36,7 +36,7 @@ export const baseInteraction: SetupDesiredProps<Interaction, CompleteDesiredProp
   // This allows typescript to still check for type errors on functions below
   ...(undefined as unknown as SetupDesiredProps<Interaction, CompleteDesiredProperties<{}, true>, DesiredPropertiesBehavior.RemoveKey>),
 
-  async respond(response, options) {
+  async reply(response, options) {
     let type = InteractionResponseTypes.ChannelMessageWithSource
 
     // If user provides a string, change it to a response object
@@ -55,7 +55,7 @@ export const baseInteraction: SetupDesiredProps<Interaction, CompleteDesiredProp
 
     // Modals cannot be chained
     if (this.type === InteractionTypes.ModalSubmit && type === InteractionResponseTypes.Modal)
-      throw new Error('Cannot respond to a modal interaction with another modal.')
+      throw new Error('Cannot reply to a modal interaction with another modal.')
 
     const result = await this.bot.helpers.sendInteractionResponse(
       this.id,
@@ -92,7 +92,7 @@ export const baseInteraction: SetupDesiredProps<Interaction, CompleteDesiredProp
 
     return await this.bot.helpers.editOriginalInteractionResponse(this.token, response)
   },
-  async deferEdit(options) {
+  async deferUpdate(options) {
     if (this.type === InteractionTypes.ApplicationCommandAutocomplete) throw new Error('Cannot edit an autocomplete interaction.')
     if (this.acknowledged) throw new Error('Cannot defer an already responded interaction.')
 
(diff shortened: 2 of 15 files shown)


zt3xdv:
aliases will work well

examples/advanced/src/commands/ping.ts
```diff
@@ -10,6 +10,6 @@ createCommand({
 
     const embeds = createEmbeds().setTitle(`The bot ping is ${ping}ms`)
 
-    await interaction.respond({ embeds })
```
Fleny113, line 13: Examples use v20, so this should say as it (this is valid for all examples file, I'm not going to comment on all to avoid some spam)

packages/bot/src/transformers/interaction.ts
```diff
@@ -36,7 +36,7 @@ export const baseInteraction: SetupDesiredProps<Interaction, CompleteDesiredProp
   // This allows typescript to still check for type errors on functions below
   ...(undefined as unknown as SetupDesiredProps<Interaction, CompleteDesiredProperties<{}, true>, DesiredPropertiesBehavior.RemoveKey>),
 
-  async respond(response, options) {
+  async reply(response, options) {
```
Fleny113, line 39: Respond is probably better in here since discord calls it ["Create Interaction Response"](https://discord.com/developers/docs/interactions/receiving-and-responding#create-interaction-response)

packages/bot/src/transformers/interaction.ts
```diff
@@ -92,7 +92,7 @@ export const baseInteraction: SetupDesiredProps<Interaction, CompleteDesiredProp
 
     return await this.bot.helpers.editOriginalInteractionResponse(this.token, response)
   },
-  async deferEdit(options) {
+  async deferUpdate(options) {
```
Fleny113, line 95: I have to be honest, checking the other methods idk if i like this too much.
While it's true that this way it matches what the interaction response type is (Deferred**Update**Message) it is inconsistent with `edit` above this, but calling that `update` doesn't sound too good to me as it does either an `UpdateMessage` (if there isn't a response) or ["Edit Original Interaction Response"](https://discord.com/developers/docs/interactions/receiving-and-responding#edit-original-interaction-response), so idk what would be best to call this

There is an argument to be made that this method should be splitted between the edit original & update message, same thing for respond, i will think about this a bit (@discordeno/team what do you think?)

packages/rest/src/types.ts
```diff
@@ -2558,7 +2558,7 @@ export interface RestManager {
   /**
    * Sends a response to an interaction.
    *
-   * @param interactionId - The ID of the interaction to respond to.
+   * @param interactionId - The ID of the interaction to reply to.
```
Fleny113, line 2561: This should stay "respond" as that is what discord calls it

packages/types/src/discord/interactions.ts
```diff
@@ -206,7 +206,7 @@ export interface DiscordMessageInteraction {
 export enum InteractionResponseTypes {
   /** ACK a `Ping` */
   Pong = 1,
-  /** Respond to an interaction with a message */
+  /** Reply to an interaction with a message */
```
Fleny113, line 209: This should say as it, this is directly from the discord docs
(same thing for `PremiumRequired` below, it should stay as is)

website/docs/examples/reactionroles.md
```diff
@@ -537,7 +537,7 @@ async execute(interaction, args: CommandArgs) {
       ],
     })
 
-    await interaction.respond(
+    await interaction.reply(
```
Fleny113, line 540: Docs use v21, so this should say as is.

When we do (if we manage to) versioned docs this will be in v22 docs

(this is valid for both docs files edited, not going to comment on each occasion to avoid spamming with comments this pr)

Fleny113 (changes requested):
I have to be honest, checking the other methods in the interaction prototype idk if i like this rename too much.

While it's true that deferEdit this way matches what the interaction response type is (`DeferredUpdateMessage`) it is inconsistent with edit above it, but calling that update doesn't sound too good to me as it does either an UpdateMessage (if there isn't a response) or ["Edit Original Interaction Response"](https://discord.com/developers/docs/interactions/receiving-and-responding#edit-original-interaction-response), so idk what would be best to call this

There is an argument to be made that this method should be splitted between the edit original & update message, same thing for respond, i will think about this a bit

@discordeno/team what do you think?

packages/bot/src/transformers/interaction.ts
lts20050703: I think it should be split. Optionally we can add a new function that do the same thing as the current `deferEdit`

PR #4556, opened 2025-11-02
fix(bot): make Interaction.guild a Partial<Guild>

Fleny113:


packages/bot/src/transformers/types.ts
```diff
@@ -947,7 +947,7 @@ export interface Interaction {
   /** The type of interaction */
   type: InteractionTypes
   /** Guild that the interaction was sent from */
-  guild: Guild
+  guild: Partial<Guild>
```
AwesomeStickz:
```suggestion
  guild?: Partial<Guild>
```

PR #4652, opened 2025-12-22
feat(user)!: enhance user tag getter

louiszn:
The current tag getter is currently for legacy username system, which uses `discriminator`. But for the new system, `discriminator` now only remains for discord bots.

For user, `discriminator` will be `0` and webhook/deleted account/system user will be `0000`. So I'd like to make it fallback to username instead of keeping `username#discriminator` form.


packages/bot/src/transformers/user.ts
```diff
@@ -11,7 +11,9 @@ export const baseUser: User = {
   ...(undefined as unknown as User),
 
   get tag() {
-    return `${this.username}#${this.discriminator}`
+    const { discriminator, username } = this
```
Fleny113: these is no need to destructure this

packages/bot/src/transformers/user.ts
```diff
@@ -11,7 +11,9 @@ export const baseUser: User = {
...
+    const isLegacy = discriminator === '0' || discriminator === '0000'
```
Fleny113: wouldn't #0000 be a valid discriminator?
> The new username system will rollout to users over time rather than all at once. The value of a single zero ("0") in the [discriminator field](https://discord.com/developers/docs/resources/user#user-object-user-structure) on a user will indicate that the user has been migrated to the new username system. Note that the discriminator for migrated users will not be 4-digits like a standard discriminator (it is "0", not "0000").
- [Discord API announcement related to the usernames](https://discord.com/developers/docs/change-log#unique-usernames-on-discord)

Looking around it seems webhook use 0000, was that the reason for this?

packages/bot/src/transformers/user.ts
louiszn: The condition looks too long for me so I destructed it. But surely I can change it back.

packages/bot/src/transformers/user.ts
louiszn: I'm kinda used to how discord.js tag getter works so I think the 0000 discriminator is kinda unnecessary. What do you think?

packages/bot/src/transformers/user.ts
Fleny113: even in this file we definitely have longer lines, i don't think adding the `this.` would add much

packages/bot/src/transformers/user.ts
Fleny113: I went to check discord.js, they do it for webhooks it seems from the commits/[prs](https://redirect.github.com/discordjs/discord.js/pull/11062)
I guess it's fine to keep 0000, we would need to add it to the default profile picture function as well but it is better to be made in a different pr from this

packages/bot/src/transformers/user.ts
louiszn: I changed to `this` keyword

Fleny113 (approved):

PR #4690, opened 2026-01-17
fix(bot)!: Split DiscordSelectComponent into specific select components

Fleny113:
And add missing resolved property from the responses, since this now correctly transformers the responses, resolved has to be added.
The reverse transformer does not transform resolved as it requires a bunch of others, such as a user, role, message reverse transformers etc.

> [!NOTE]
> This is based on #4689 and #4435
> - #4689 as it fixes the resolved bug which would be triggered by this changeset as well
> - #4435 as it makes the code less messy dealing with this

### DIFF
--- a/packages/bot/src/handlers/invites/INVITE_CREATE.ts
+++ b/packages/bot/src/handlers/invites/INVITE_CREATE.ts
@@ -6,5 +6,6 @@ export async function handleInviteCreate(bot: Bot, data: DiscordGatewayPayload, 
 
   const payload = data.d as DiscordInviteCreate;
 
+  // TODO: Add role_ids, the transformer should be kept for the Invite type, not for the gateway event
   bot.events.inviteCreate(bot.transformers.invite(bot, payload, { shardId }));
 }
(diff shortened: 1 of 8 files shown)


TriForMine (dismissed):

packages/bot/src/transformers/component.ts
```diff
@@ -211,6 +221,118 @@ function transformSelectMenuComponent(bot: Bot, payload: DiscordSelectMenuCompon
...
+
+function transformChannelSelectMenuComponent(bot: Bot, payload: DiscordChannelSelectComponent | DiscordChannelSelectInteractionResponseFromModal) {
+  const props = bot.transformers.desiredProperties.component;
+  const select = {} as SetupDesiredProps<Component, TransformersDesiredProperties, DesiredPropertiesBehavior>;
+
+  if (props.type && payload.type) select.type = payload.type;
+  if (props.id && payload.id) select.id = payload.id;
+  if (props.customId && payload.custom_id) select.customId = payload.custom_id;
+
+  // Check if this is the mentionable select response
```
AwesomeStickz:
```suggestion
  // Check if this is the channel select response
```

PR #4869, opened 2026-03-18
fix(types): make value of emoji property in DiscordButtonComponent and DiscordSelectOption partial

AwesomeStickz:
These are partial according to docs


AwesomeStickz:
> do not put a Partial without the partial transformer pr

This PR can wait until the partial transformer PR then

> also the docs do say "Partial", however in the description there is the list of the props you can expect

Are you referring to this?
image
If yes then it doesn't say anywhere that you can expect it unless I'm missing something

Fleny113:
yes i am referring to that
in that table it just lists the field, in others there is an asterisk, it is quite inconsistent but that's how i always interpret those things

PR #5085, opened 2026-06-15
feat(rest): add request timeout to manage stalled requests

NotDemonix:
Summary
- Adds a configurable `requestTimeout` option (default 30s) to `CreateRestManagerOptions`/`RestManager` that aborts a single request attempt if it stalls.
- A timed-out attempt is retried through the queue like other failures, up to `maxRetryCount`.
- Set `requestTimeout: 0` to disable and fall back to the runtime's default fetch timeouts.

### DIFF
--- a/packages/rest/src/manager.ts
+++ b/packages/rest/src/manager.ts
@@ -87,6 +87,13 @@ export const RATE_LIMIT_BUCKET_HEADER = 'x-ratelimit-bucket';
 export const RATE_LIMIT_LIMIT_HEADER = 'x-ratelimit-limit';
 export const RATE_LIMIT_SCOPE_HEADER = 'x-ratelimit-scope';
 
+/** Calls `.unref()` on a timer if the runtime's `setTimeout` returns an object supporting it (Node, Bun), so the timer doesn't keep the process alive. Browsers and Deno return a plain number and have no equivalent. */
+function unrefTimer(timer: unknown): void {
+  if (typeof timer === 'object' && timer !== null && 'unref' in timer && typeof (timer as { unref: unknown }).unref === 'function') {
+    (timer as { unref: () => void }).unref();
+  }
+}
+
 export function createRestManager(options: CreateRestManagerOptions): RestManager {
   const applicationId = options.applicationId ? BigInt(options.applicationId) : getBotIdFromToken(options.token);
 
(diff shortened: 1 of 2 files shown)


NotDemonix:
I've found this issue due to Tickety sometimes getting stuck on some guilds for actions like createChannel.
The only way to unblock this was to restart the dd rest proxy.

packages/rest/src/manager.ts
```diff
@@ -428,24 +429,63 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
...
+      const controller = new AbortController();
+      const timeoutId =
+        rest.requestTimeout > 0
+          ? setTimeout(() => {
+              timedOut = true;
+              controller.abort();
+            }, rest.requestTimeout).unref()
+          : undefined;
+
+      const request = new Request(url, { ...payload, signal: controller.signal });
```
Fleny113:
```suggestion
const request = new Request(url, { ...payload, signal: AbortSignal.timeout(rest.requestTimeout) });
```
I would guess that it is better to use the [abort signal timeout function](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static) and in the catch handle the DOM exception

packages/rest/src/manager.ts
```diff
@@ -87,6 +87,13 @@ export const RATE_LIMIT_BUCKET_HEADER = 'x-ratelimit-bucket';
 export const RATE_LIMIT_LIMIT_HEADER = 'x-ratelimit-limit';
 export const RATE_LIMIT_SCOPE_HEADER = 'x-ratelimit-scope';
 
+/** Calls `.unref()` on a timer if the runtime's `setTimeout` returns an object supporting it (Node, Bun), so the timer doesn't keep the process alive. Browsers and Deno return a plain number and have no equivalent. */
```
Fleny113: The issue with this is Deno, as it's the only one strictly keeping the browser behavior returning a number

However importing setTimeout from node:timers does make deno use it's deno compat making the timer an object

sample code:
```
import { setTimeout } from "node:timers"
// add .unref() and deno will terminate immediatly
setTimeout(() => { console.log("A") }, 10_000);
console.log("B");
```

apparently they also changed this in the latest versions, but i can't find when they did the change
edit: deno 2.8 apparently: https://deno.com/blog/v2.8#settimeout-and-setinterval which is from may so too new to rely on this i'd say

packages/rest/src/manager.ts
Fleny113, line 508: I don't understand the need for the try catch when unless i'm missing something the catch can do it by iself

Fleny113, line 697: Do you never retry on proxy timeout? 
This code does not have the same structure as the normal sendRequest, it does not have queues, it does not have retries

Fleny113 (approved):
LGTM

PR #5101, opened 2026-06-21
feat(rest): strip trailing slash from proxy baseUrl if present in createRestManager()

finaIcutpro:
this is a pretty simple pr. just gets rid of the trailing slash.

### DIFF
--- a/packages/rest/src/types.ts
+++ b/packages/rest/src/types.ts
@@ -169,7 +169,6 @@ export interface CreateRestManagerOptions {
   proxy?: {
     /**
      * The base url to connect to. If you create a proxy rest, that url would go here.
-     * IT SHOULD NOT END WITH A /
      * @default https://discord.com/api
      */
     baseUrl: string;
@@ -235,7 +234,6 @@ export interface RestManager {
   version: ApiVersions;
   /**
    * The base url to connect to. If you create a proxy rest, that url would go here.
-   * IT SHOULD NOT END WITH A /
    * @default https://discord.com/api
    */
   baseUrl: string;
--- a/packages/rest/src/types.ts
+++ b/packages/rest/src/types.ts
@@ -234,6 +234,7 @@ export interface RestManager {
   version: ApiVersions;
   /**
    * The base url to connect to. If you create a proxy rest, that url would go here.
+   * IT SHOULD NOT END WITH A /
    * @default https://discord.com/api
    */
   baseUrl: string;
--- a/packages/rest/tests/unit/manager.spec.ts
+++ b/packages/rest/tests/unit/manager.spec.ts
@@ -44,9 +44,9 @@ describe('[rest] manager', () => {
       expect(rest.isProxied).to.be.equal(true);
     });
 
-it('With an application id', () => {
-const rest = createRestManager({ ...options, applicationId: '130136895395987456' });
-expect(rest.applicationId).to.be.equal(130136895395987456n);
+    it('With an application id', () => {
+      const rest = createRestManager({ ...options, applicationId: '130136895395987456' });
+      expect(rest.applicationId).to.be.equal(130136895395987456n);
     });
   });
 
--- a/packages/rest/tests/unit/manager.spec.ts
+++ b/packages/rest/tests/unit/manager.spec.ts
@@ -39,15 +39,8 @@ describe('[rest] manager', () => {
     });
 
     it('Strips a trailing slash from proxy base urls', () => {
-      const rest = createRestManager({
-        token,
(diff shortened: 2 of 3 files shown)


packages/rest/tests/unit/manager.spec.ts
```diff
@@ -38,6 +38,19 @@ describe('[rest] manager', () => {
...
+        token,
+        proxy: {
+          baseUrl: 'http://microservices_bot-gateway:6001/',
+          authorization: token,
+        },
+      });
+
+      expect(rest.baseUrl).to.be.equal('http://microservices_bot-gateway:6001');
+      expect(rest.isProxied).to.be.equal(true);
+    });
```
AwesomeStickz, line 45:
```suggestion
    it('Strips a trailing slash from proxy base urls', () => {
      const rest = createRestManager({ ...options, proxy: { ...options.proxy, baseUrl: 'https://localhost:8000/' } });
      expect(rest.baseUrl).to.be.equal('https://localhost:8000');
      expect(rest.isProxied).to.be.equal(true);
    });
```

packages/rest/tests/unit/manager.spec.ts
```diff
@@ -38,6 +38,19 @@ describe('[rest] manager', () => {
...
+
     it('With an application id', () => {
       const subrest = createRestManager({ ...options, applicationId: '130136895395987456' });
       expect(subrest.applicationId).to.be.equal(130136895395987456n);
```
AwesomeStickz:
```suggestion
it('With an application id', () => {
const rest = createRestManager({ ...options, applicationId: '130136895395987456' });
expect(rest.applicationId).to.be.equal(130136895395987456n);
```

AwesomeStickz (changes requested):

packages/rest/tests/unit/manager.spec.ts
finaIcutpro: okay that's fine. the intention was to show how this works with docker swarms

AwesomeStickz:
image

Can you also remove this (only from `CreateRestManagerOptions` and not `RestManager`)

packages/rest/tests/unit/manager.spec.ts
AwesomeStickz: i just tried to keep it consistent with other tests but yeah it works with anything anyway

packages/rest/tests/unit/manager.spec.ts
```diff
@@ -38,9 +38,15 @@ describe('[rest] manager', () => {
...
-      expect(subrest.applicationId).to.be.equal(130136895395987456n);
+    it('Strips a trailing slash from proxy base urls', () => {
+      const rest = createRestManager({ ...options, proxy: { ...options.proxy, baseUrl: 'https://localhost:8000/' } });
+      expect(rest.baseUrl).to.be.equal('https://localhost:8000');
+      expect(rest.isProxied).to.be.equal(true);
+    });
+
+it('With an application id', () => {
+const rest = createRestManager({ ...options, applicationId: '130136895395987456' });
+expect(rest.applicationId).to.be.equal(130136895395987456n);
```
AwesomeStickz:
```suggestion
    it('With an application id', () => {
      const rest = createRestManager({ ...options, applicationId: '130136895395987456' });
      expect(rest.applicationId).to.be.equal(130136895395987456n);
```

I broke the indentation with my previous suggestion

lts20050703 (dismissed):
LGTM!

Does the trailing slash cause any issues in the first place?

AwesomeStickz:
> Does the trailing slash cause any issues in the first place?

Yes, rest manager expects to not have a trailing slash

As for my review, my comment went up because of changes after that but I still think this should be changed because it's no longer true for `CreateRestManagerOptions`:

> image
> Can you also remove this (only from `CreateRestManagerOptions` and not `RestManager`)

finaIcutpro:
should be good to merge now

packages/rest/src/types.ts
```diff
@@ -235,7 +234,6 @@ export interface RestManager {
   version: ApiVersions;
   /**
    * The base url to connect to. If you create a proxy rest, that url would go here.
```
AwesomeStickz, line 236:
```suggestion
   * The base url to connect to. If you create a proxy rest, that url would go here.
   * IT SHOULD NOT END WITH A /
```

This change only affects `CreateRestManagerOptions` and not `RestManager` so this should remain for `RestManager`

AwesomeStickz (approved):

lts20050703 (approved):

PR #5158, opened 2026-07-19
feat(rest): add signal to makeRequest and stop re-sending proxied requests by default

NotDemonix:
Follow-up to #5085.

While running the timeout changes in production I found an edge case in proxy mode: when a request to the proxy hits `requestTimeout`, only bot side of the connection gets aborted. The proxy keeps processing the request and it can still reach Discord. Resending it after the abort means non-idempotent requests get executed more than once, we had a single ticket action create several empty channels because the channel create kept being resent while the original was waiting out a rate limit on the proxy.

Retries around Discord-side slowness should only happen when talking to Discord directly, which the queue already handles when `isProxied` is false. In proxy mode the only thing that would be safe to resend is an attempt the proxy definitely never got, and fetch does not tell us that: a timeout only aborts our side of the connection, and a socket that dies once the request is on the wire may still have delivered it. So with this PR:

- nothing is re-sent in proxy mode by default, whatever the attempt failed on. The abort itself still protects the caller from stalled connections
- `proxy.retryRequests` opts back in, for setups where re-sending against the proxy is safe (an idempotency key the proxy tracks, for example). Bounded by `maxRetryCount` and `maxProxyRetryCount`, with a small backoff so a proxy that is down isn't hammered in a tight loop
- direct-to-Discord behavior is unchanged
- added unit tests for the proxied path and updated the `requestTimeout` docs to match

---

Update: this now also adds an optional `signal` to makeRequest.

It's the other half of the timeout problem. My rest runs as its own process like in the big bot example. When the bot side gives up on the http request, the request itself is often still sitting in the queue of the rest process, and there was no way to cancel it, so it would get executed later with nobody waiting for the result anymore. With a signal the rest process can tie the queued request to the incoming http request and drop it when the client disconnects or a deadline passes.

Aborting settles the caller right away rather than when the queue reaches it, and the request is taken back out of the queue, so it does not spend a rate limit slot or hold on to its payload until the queue resumes. The signal is handed to fetch as well, so an attempt that is already on the wire gets torn down instead of being read to the end.


packages/rest/src/manager.ts
```diff
@@ -652,19 +680,21 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
             body: options?.body,
           });
 
-          const result = await fetch(request).catch((fetchError) => {
+          const result = await fetch(request).catch(async (fetchError) => {
```
Fleny113, line 736: Why not let the fetch web api handle the AbortSignal? you can combine AbortSignals with [AbortSignal.any](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/any_static) and it's node 20+ so we can use it (bun and deno also support it)

packages/rest/tests/unit/manager.spec.ts
```diff
@@ -229,4 +232,147 @@ describe('[rest] manager', () => {
...
+      fetchStub = sinon.stub(globalThis, 'fetch');
+    });
+
+    afterEach(() => {
+      fetchStub.restore();
+    });
+
+    it('Rejects an already aborted request without sending it', async () => {
+      const rest = createRestManager({ token });
+      const controller = new AbortController();
```
Fleny113: AbortSignal.abort could be used instead if you want

Fleny113 (changes requested):
Sorry but if you don't resent the request, doesn't this mean that the bot will never receive the discord response?

NotDemonix:
> Sorry but if you don't resent the request, doesn't this mean that the bot will never receive the discord response?

Yes, but the response is lost either way. The timeout already killed the connection on our (bot) side, so resending is the only way to get a response back and that's what caused the duplicates in the first place. The proxy was still processing the original while we resent it.

The way I see it retries are split between the two sides. The bot should only retry when the proxy was never reached at all (down, restarting), because those requests were never processed so resending is safe. Discord side retries are the proxy's job since it's the one talking to Discord, and the rest manager already does that in direct mode, 429s wait out retry-after and go again, timeouts get requeued. And with the signal passed along, the proxy stops retrying as soon as the bot hangs up, so nothing gets retried with nobody waiting for the result.

Nirn works the same way btw, I checked the source. It never resends on its own timeout either, it just returns 408 and lets the client decide if resending is safe, since only the caller knows if a request is idempotent. If waiting out rate limits behind the proxy is expected, requestTimeout just needs to be high enough (or 0 to disable it).

packages/rest/src/manager.ts
```diff
@@ -426,6 +426,19 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
     },
 
     async sendRequest(options) {
+      // The request may have been aborted while it was waiting in the queue; never dispatch it in that case.
```
Fleny113: if you want, this code is unnecessary, fetch will check the aborted before sending any http request.

packages/rest/src/types.ts
```diff
@@ -3289,6 +3294,17 @@ export interface CreateRequestBodyOptions {
   unauthorized?: boolean;
   reason?: string;
   files?: FileContent[];
+  /**
+   * An `AbortSignal` to cancel the request.
+   *
+   * @remarks
+   * Aborting rejects a request that is still waiting in the queue right away and guarantees it is never sent.
+   * An attempt that is already in flight gets its connection cancelled as well, though that cannot recall it:
+   * Discord (or the proxy) may have received it and will still process it, only the response is discarded.
```
Fleny113:
```suggestion
   * Discord (or the proxy) may have received it and may still process it, only the response is discarded.
```
You are not sure, if a proxy implements an AbortSignal it will drop the request for example

Fleny113:
Beside the failing tests now fixed

I think the signal is fine and can be kept, if you want in a separate pr
Beside that: if someone wants to implement to code such as a nonce for each requests towards the proxy they should be able to to keep the retrying, otherwise not

so i'd propone adding a "retryProxyRequests: boolean" param which defaults to false and can be enabled

packages/rest/src/manager.ts
```diff
@@ -640,31 +660,61 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
...
+            if (options?.signal?.aborted) throw fetchError;
+
+            if (retryCount < rest.maxRetryCount) {
+              // The proxy could not be reached, so the request was never processed and is safe to retry.
+              if (!isTimeoutError(fetchError)) {
+                rest.logger.debug(`request to proxy ${url} failed before reaching it, retrying.`, fetchError);
+                // Small backoff so a proxy that is down or restarting isn't hammered in a tight loop.
+                await delay(250);
+                return undefined;
+              }
```
AwesomeStickz: Considering that `rest.maxRetryCount` is `Infinity` by default, this can cause a mem leak if your requests keep failing for other reasons for extended period of time, for example your proxy is down and it's unable to connect etc. This would not be ideal

packages/rest/src/manager.ts
```diff
@@ -640,31 +660,61 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
...
+        // rate limit) and it can still reach Discord. Re-sending it would execute non-idempotent requests twice
+        // (e.g. duplicate channel creates). Only attempts that failed to reach the proxy at all (connection
+        // refused/reset) are retried, since those were never processed — unless the user opted into timeout
+        // retries via `proxy.retryOnTimeout` because their setup deduplicates re-sent requests. Retrying on
+        // Discord errors is the job of the queue when talking to Discord directly (`isProxied` is false); a proxy
+        // is expected to handle its own retrying.
         for (let retryCount = 0; ; retryCount++) {
-          // Give the request to the proxy a hard deadline too via `AbortSignal.timeout` (omitted when disabled).
+          // The request may have been aborted while waiting out the retry backoff; never send it in that case.
+          if (options?.signal?.aborted) throw options.signal.reason;
```
AwesomeStickz: Do we want to throw an error like this and not follow our error object format? Same for line 695 and 714

examples/bigbot/src/rest/index.ts
```diff
@@ -20,9 +20,18 @@ app.all('/*', async (req, res) => {
   const hasBody = req.method !== 'GET' && req.method !== 'DELETE';
   const body = hasBody ? (isMultipart ? await parseMultiformBody(req.body) : req.body) : undefined;
 
+  // Tie the request to the connection: when the client aborts, the request is dropped from the queue instead
+  // of being executed with nobody waiting for the result anymore.
+  // https://fastify.dev/docs/latest/Guides/Detecting-When-Clients-Abort/
+  const controller = new AbortController();
+  req.raw.on('close', () => {
+    if (req.raw.aborted) controller.abort();
```
AwesomeStickz, line 28:
```suggestion
    if (req.raw.destroyed) controller.abort();
```

`aborted` is deprecated in favor of `destroyed`

AwesomeStickz:
Also please update the PR title

examples/bigbot/src/rest/index.ts
Fleny113: From the fastify page referenced:
> Whilst the aborted property has been deprecated, destroyed is not a suitable replacement as the Node.js documentation suggests. A request can be destroyed for various reasons, such as when the server closes the connection. The aborted property is still the most reliable way to detect when a client intentionally aborts a request.

AwesomeStickz: Interesting, I just noticed it from vsc and didn't read fastify's documentation

packages/rest/src/manager.ts
NotDemonix: Connection retries now have their own cap (3, or maxRetryCount if it's lower) so a proxy that stays down fails the request instead of keeping it alive forever. Also added a test for it.

packages/rest/src/manager.ts
NotDemonix: Fixed. The message/cause building from the queued path is now a helper both paths use, so proxied requests throw the same error shape with status 999 and the original error in `cause.errorObject`. The non-ok response in proxy mode goes through it too, it previously set a cause but left the message empty.

packages/rest/src/manager.ts
```diff
@@ -103,6 +113,50 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
     },
   };
 
+  // Fills in the error given to the user with the message and cause built from the rejection. The error itself is
+  // created at the call site because of how stack traces get calculated.
+  const fillRequestError = (error: Error, reason: RestRequestRejection): Error => {
```
Fleny113: Any specific reason this has to be in the createRestManager function?

NotDemonix: No, it only needed baseErrorPrototype and that one doesn't depend on the manager either. Both are at module level now.

packages/rest/src/manager.ts
NotDemonix: Made it configurable instead of a fixed number: `rest.maxProxyConnectionRetryCount`, 3 by default, and still bound by `maxRetryCount` when that one is lower.

packages/rest/src/manager.ts
NotDemonix: Moved it onto the manager as `rest.createRequestError`, next to `createRequestBody`, so it follows the same pattern as the other members and stays overridable. `baseErrorPrototype` is back where it was, untouched.

packages/rest/src/manager.ts
NotDemonix: Dropped the extra counter, `maxRetryCount` is the only one again. Connection failures towards the proxy are no longer re-sent at all, they reject right away with the same error a failed fetch gives when talking to Discord directly, so there is no loop left to bound. The only re-sending left in proxy mode is the opt in `proxy.retryOnTimeout`.

packages/rest/src/manager.ts
NotDemonix: Every failure in proxy mode now goes through `rest.createRequestError` and reuses the exact strings from the direct path: same message for a timeout, a failed fetch and an abort, same `{ status, statusText, body }` for a non ok response, and `errorObject` on all of them. The non ok response in proxy mode also gets a message now, it used to set a cause and leave the message empty.

packages/rest/src/manager.ts
```diff
@@ -699,51 +798,25 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
...
           },
           runThroughQueue: options?.runThroughQueue,
         };
 
+        // Reject a queued request as soon as its signal aborts instead of when the queue reaches it, so a caller enforcing a deadline gets its
+        // answer right away. Fetch refusing to send an already aborted request is what guarantees the stale queue entry never goes out.
+        if (signal?.aborted) {
+          abortListener();
+          return;
+        }
```
Fleny113, line 842: just to be sure, this is to handle if the signal is already aborted when we create the request payload (the payload object above) right?

Correct me if i am wrong but doesn't this either belong in processRequest to avoid queuing an aborted signal (so the retryRequest also does this check) or can be avoided and let the request to to sendRequest where it will be aborted by fetch?

packages/rest/src/manager.ts
```diff
@@ -640,13 +700,24 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
...
-        // connection doesn't permanently fail the request (and can never hang the caller forever).
+        // A timed out attempt is not re-sent by default: the timeout only aborts our side of the connection, the proxy keeps processing the
+        // request (it may simply be queued behind a rate limit) and it can still reach Discord, so re-sending it would execute non-idempotent
+        // requests twice. Retrying is the proxy's job, it is the one talking to Discord.
         for (let retryCount = 0; ; retryCount++) {
-          // Give the request to the proxy a hard deadline too via `AbortSignal.timeout` (omitted when disabled).
+          // The request may have been aborted while the previous attempt was timing out, never send it in that case.
+          if (options?.signal?.aborted) {
+            throw rest.createRequestError(error, { ok: false, status: 999, error: 'The request was aborted.', errorObject: options.signal.reason });
+          }
```
Fleny113, line 720: This is not necessary given the fetch call below
The only reason i can think of keeping this is avoid calling the events (request + requestError) and creating the signals array

Fleny113 (dismissed):
I will wait for stickz to review this as well, the comments are nit more than anything

packages/rest/src/manager.ts
NotDemonix: Yes. If it goes through anyway it still fails, just later, once the queue gets to it, which can be after a rate limit wait. The idea is that a caller who already gave up doesn't keep waiting. processRequest isn't needed: sendRequest rejects instead of retrying when the signal is aborted, and fetch won't send an aborted request anyway. Can move it there if you prefer.

packages/rest/src/manager.ts
NotDemonix: Yeah, just the events. Without it we'd fire request + requestError for a request that never actually gets sent. Everything else is the same, fetch throws the same error. Can remove it if you want.

AwesomeStickz (dismissed):
Is the removal of handling connection errors intentional?

NotDemonix:
> Is the removal of handling connection errors intentional?

Was, but thinking about it again I'd rather have it back. I sometimes restart the proxy without the bot which could cause the bot not running a request just because the proxy was down for a split second.
Only thing is a failed fetch doesn't always mean we didn't connect. If the socket drops mid request it may have already gone to Discord, so resending brings the duplicates back. So it only retries when it failed while connecting (refused, dns, connect timeout), anything else rejects right away.

packages/rest/src/manager.ts
```diff
@@ -678,20 +798,26 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
...
           return result.status !== 204 ? (typeof body === 'string' ? JSON.parse(body) : body) : undefined;
         }
       }
 
       return await new Promise(async (resolve, reject) => {
+        const signal = options?.signal;
+        const abortListener = () => {
+          // Reject right away instead of waiting for the queue to reach the request. The signal is attached to the fetch itself as well, so an
+          // attempt that is already in flight gets its connection torn down too.
+          payload.reject({
```
TriForMine, line 810: **[P1] Remove cancelled requests from the queue instead of only settling the caller**

This rejects the caller, but it does not remove the request from `Queue.waiting` or `Queue.pending`. When a route is unavailable, every aborted call leaves a waiter whose continuation retains the payload; once the queue resumes, stale requests also pass through rate-limit accounting and call `fetch` with an already-aborted signal. I reproduced 25 cancelled and settled requests leaving all 25 entries in `Queue.waiting`.

Please make cancellation remove the queued work—or have the queue skip aborted entries before waiting and rate-limit accounting—so cancellation actually releases its resources and cannot delay live requests.

_AI-written review comment — model: GPT-5.6 Sol._

packages/rest/src/manager.ts
```diff
@@ -678,20 +798,26 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
...
           if (!result.ok) {
-            error.cause = Object.assign(Object.create(baseErrorPrototype), {
-              ok: false,
-              status: result.status,
-              body,
-            });
-
-            throw error;
+            throw rest.createRequestError(error, { ok: false, status: result.status, statusText: result.statusText, body });
           }
```
TriForMine, line 799: **[P1] Reject when the signal aborts while consuming the proxy response body**

`fetch` can resolve as soon as the response headers arrive. If the caller aborts while `result.json()` or `result.text()` is still reading, that read rejects, but the preceding `.catch(() => null)` swallows the abort. Unlike the queued path, the proxy path has no outer abort listener, so `makeRequest` then resolves successfully with `null`; I reproduced that behavior with a streaming response.

Re-check the caller signal after reading the body so the documented cancellation contract is preserved:

```suggestion
          if (options?.signal?.aborted) {
            throw rest.createRequestError(error, {
              ok: false,
              status: 999,
              error: 'The request was aborted.',
              errorObject: options.signal.reason,
            });
          }

          if (!result.ok) {
            throw rest.createRequestError(error, { ok: false, status: result.status, statusText: result.statusText, body });
          }
```

_AI-written review comment — model: GPT-5.6 Sol._

packages/rest/src/manager.ts
```diff
@@ -642,31 +703,90 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
...
 
-        // Retry on timeout up to `maxRetryCount`, mirroring the queued `sendRequest` path, so a stalled proxy
-        // connection doesn't permanently fail the request (and can never hang the caller forever).
-        for (let retryCount = 0; ; retryCount++) {
-          // Give the request to the proxy a hard deadline too via `AbortSignal.timeout` (omitted when disabled).
+        // A timed out attempt is not re-sent by default: the timeout only aborts our side of the connection, the proxy keeps processing the
+        // request (it may simply be queued behind a rate limit) and it can still reach Discord, so re-sending it would execute non-idempotent
+        // requests twice. Retrying that is the proxy's job, it is the one talking to Discord. Attempts that never made it onto the wire are
+        // still re-sent, those never reached the proxy at all.
+        let timeoutRetryCount = 0;
```
TriForMine: **[P1] Keep a total retry count as well as the connection-specific count**

The timeout and connection counters are each independently bounded by `maxRetryCount`, so one request can exceed the documented global retry cap by alternating failure types. For example, with `maxRetryCount = 1` and `retryOnTimeout = true`, I reproduced `TimeoutError -> connect error -> success`, which performs two retries.

Please track a total retry count shared by both branches while retaining `connectionRetryCount` for the narrower `maxProxyConnectionRetryCount` limit, and add a mixed-failure test for the global cap.

_AI-written review comment — model: GPT-5.6 Sol._

TriForMine (changes requested):
AI-authored code review — model: GPT-5.6 Sol.

Requesting changes for three correctness issues: cancelled requests remain retained by the rate-limit queue, a proxied request can resolve successfully after being aborted during response-body consumption, and mixed proxy failures can exceed the documented global retry cap.

packages/rest/src/manager.ts
NotDemonix: Fixed, cancellation takes the request back out of the queue now instead of only settling the caller. `waitUntilRequestAvailable` takes the signal and splices the waiter out when it fires, `Queue.makeRequest` stops before pushing to `pending` if the signal went off while it was waiting for a slot, and `processPending` drops an entry whose caller gave up while it sat there, before the interval wait and before `remaining--`. So a cancelled request no longer spends a rate limit slot, never reaches fetch, and its payload is not kept alive until the queue resumes.

`processWaiting` calls `cleanup` when its loop ends too. Without it a queue whose waiters were all cancelled never moves anything to `pending`, so nothing would schedule its deletion and it would sit in `rest.queues` forever.

The existing queue abort test asserts one fetch call instead of two now, plus an empty `waiting` and `pending`, and there is a new one for a request that gets cancelled after it already made it into `pending`.

packages/rest/src/manager.ts
NotDemonix: Fixed. The body read is checked against the caller's signal now and throws the same aborted error the other abort paths in proxy mode throw.

One difference from your patch: it only throws when the read itself failed. A read that finished is still handed back, because the queued path does the same, `resolve` removes the abort listener so a result that arrived before the signal fired wins. Throwing on the signal alone would discard a response we already have in full whenever the abort lands a moment after it, which is not that unlikely given aborts are usually deadlines.

New test hands back a response whose body stream errors when the signal fires and asserts the rejection instead of a `null` resolve.

The same hole exists for `requestTimeout` in both modes: if the deadline fires while the body is being read, the read rejects, the `null` fallback kicks in and the caller gets a successful empty response rather than a timeout. It is a separate case so I left it alone here, happy to cover it in this PR if you want it.

packages/rest/src/manager.ts
NotDemonix: Gone, there is one counter again.

The two of them existed because a failure to connect was re-sent on a budget of its own, and that whole distinction is out now. Telling a failure to connect apart from a connection that died carrying the request meant reading syscalls, error codes and error text from three runtimes, and it could never be more than a guess: fetch does not tell you which phase a failure happened in, and guessing wrong re-sends a request the proxy already has, which is the thing this PR is trying to stop.

So re-sending in proxy mode sits behind `proxy.retryRequests` now and covers anything that failed without producing a response, timeouts included. That leaves a single `retryCount`, bounded by `maxRetryCount` and `maxProxyRetryCount`, so there are no longer two budgets to alternate between.

Kept your mixed failure case as a test: with `maxRetryCount = 1` a timeout followed by a connect error stops after two fetches.

NotDemonix:
Dropped the runtime error code reading, it was reaching too far. Telling a failure to connect apart from a connection that died carrying the request meant going off syscalls, error codes and error text across Node, Bun and Deno, and it could never be better than a guess, since fetch does not tell you which phase a failure happened in. Guessing wrong re-sends a request the proxy already has, which is the thing this PR is trying to stop.

So `proxy.retryOnTimeout` is now `proxy.retryRequests` and covers any attempt that failed without producing a response, timeouts and dropped connections alike. Same safety condition for both, so one switch instead of two, and `isConnectError` and everything under it is gone.

Default stays false, which is what main does today for connection failures anyway, so nobody loses behaviour they have now. `retryRequests` because the library has no business assuming how you made a re-send safe, a nonce the proxy tracks is one way but it is your code either way.

That also takes care of the third review comment on its own: with one flag there is a single `retryCount` again, bounded by `maxRetryCount` and `maxProxyRetryCount`, so there are no longer two budgets to alternate between.

packages/rest/src/manager.ts
```diff
@@ -2029,6 +2123,9 @@ enum HttpResponseCode {
   TooManyRequests = 429,
 }
 
+/** How much longer to wait before each further re-send to a proxy, in milliseconds. The nth attempt waits n times this. */
+const proxyRetryDelayStep = 250;
```
Fleny113: Can you either add this to the rest manager or put it at the top of the file?

Fleny113 (dismissed):
beside the nit, LGTM

packages/rest/src/manager.ts
NotDemonix: Moved it onto the manager as `rest.proxyRetryDelayStep`, next to `maxProxyRetryCount`, and that one's docs point at it now instead of repeating the 250ms.

The tests set it to 0 as well, so the two that re-send no longer wait out a real backoff.

packages/rest/tests/unit/manager.spec.ts
```diff
@@ -229,4 +232,50 @@ describe('[rest] manager', () => {
...
+    it('Retries when the proxy cannot be reached', async () => {
+      fetchStub.onFirstCall().rejects(new TypeError('fetch failed'));
+      fetchStub
+        .onSecondCall()
+        .resolves(new Response(JSON.stringify({ url: 'wss://gateway.discord.gg' }), { headers: { 'Content-Type': 'application/json' } }));
+
+      const result = await rest.makeRequest('GET', '/gateway/bot');
+      expect(result).to.be.deep.equal({ url: 'wss://gateway.discord.gg' });
+      expect(fetchStub.callCount).to.be.equal(2);
+    });
```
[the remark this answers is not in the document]
NotDemonix: The backoff is a manager field now (`rest.proxyRetryDelayStep`), so the proxy tests set it to 0 and none of them wait for real anymore. The two that did were at 261ms and 773ms, the whole file is down to about 120ms.

packages/rest/tests/unit/manager.spec.ts
```diff
@@ -1,10 +1,15 @@
...
+import chaiAsPromised from 'chai-as-promised';
 import { afterEach, beforeEach, describe, it } from 'mocha';
 import sinon from 'sinon';
 import { createRestManager } from '../../src/manager.js';
 import type { RestManager } from '../../src/types.js';
 import { fakeToken as token } from '../constants.js';
 
+chaiUse(chaiAsPromised);
+
+/** Builds an error in the shape Node.js gives a failed syscall. */
```
Fleny113:
```suggestion
```

Fleny113 (dismissed):
LGTM (fix the comment tho)

Fleny113:
fix formatting

Fleny113 (approved):
LGTM

TriForMine (approved):

AwesomeStickz (approved):

PR #5165, opened 2026-07-22
feat!: Remove fzstd

Fleny113:
Given node v20 EOL, we can now drop fzstd in favor of node:zlib's zstd implementation


AwesomeStickz:
Should this be considered breaking?

AwesomeStickz (approved):

PR #5223, opened 2026-07-24
build(dev): Remove dotenv, use process.loadEnvFile instead

Fleny113:


AwesomeStickz (approved):

packages/bot/tests/e2e/constants.ts
```diff
@@ -1,6 +1,8 @@
-import dotenv from 'dotenv';
-
-dotenv.config({ path: '../../.env' });
+try {
+  process.loadEnvFile('../../.env');
+} catch {
+  // ignore
```
AwesomeStickz, line 4: Not that this code breaks as of now, but wouldn't it be better to throw error here as well so a future change doesn't cause unintended issues?

Fleny113: dotenv doesn't throw + it would break CI as in there the env are passed from the CI job not a .env file

AwesomeStickz: I was just thinking it wouldn't make sense to catch and ignore, we could either just not wrap it around try catch or throw an error? Yes, it would break CI but isn't that what it's supposed to if errors?

Fleny113: sorry but i don't understand 

dotenv configure doesn't throw
node loadEnvFiles throws if it doesn't find the path

in CI creating the file is a pain I'd say

the alternative is adding a fs.exists check to avoid the try catch

PR #5229, opened 2026-07-25
fix: check for the GUILD_MEMBERS intent when requesting members

### LINKED ISSUE #5226: The zero-intents configuration bypasses validation that correctly rejects an unrelated nonzero intent configuration
## Summary

requestMembers guards its missing-GuildMembers validation with `gateway.intents &&`. Consequently the default/no-intents value 0 bypasses the validator and sends a two-member request, while the same request with an unrelated nonzero Guilds intent throws the promised error.

Validation source commit: `4a7f58a623d5b3ec264f52e856607bdd1cb92ba9`.

## Code path

- `packages/gateway/src/manager.ts:29-35`: Gateway intent configuration defaults to numeric zero
- `packages/gateway/src/manager.ts:609-613`: Public option documents intents as a calculated bitfield with default 0
- `packages/types/src/discord/gateway.ts:24-56`: GatewayIntents assigns separate nonzero Guilds and GuildMembers bits
- `packages/gateway/src/manager.ts:450-507`: requestMembers validator and fulfillment send
- `packages/gateway/src/manager.ts:804-828`: Public requestMembers intent requirements
- Repository-local validation artifact: Minimized zero-versus-nonzero intent replay

## Steps to reproduce

This report is based on a source-control-flow validation against the commit above.

1. Check out the source commit listed in this report.
2. Inspect the code path and contract anchors listed above.
3. Exercise the described boundary/state path: The zero-intents configuration bypasses validation that correctly rejects an unrelated nonzero intent configuration.
4. Compare the observed behavior with the expected contract below.

Validation note: The replay stubs only the transport send and executes the complete manager validator. With `intents=0`, `requestMembers('1',{limit:2})` resolves and sends. With the unrelated nonzero `GatewayIntents.Guilds`, the identical request throws `Cannot fetch more then 1 member without the GUILD_MEMBERS intent`. Assertions require this inconsistent pair.

## Expected behavior

The implementation should preserve the repository contract described by the documentation, public type, parser, serializer, or sibling implementation anchors above.

## Actual behavior

requestMembers guards its missing-GuildMembers validation with `gateway.intents &&`. Consequently the default/no-intents value 0 bypasses the validator and sends a two-member request, while the same request with an unrelated nonzero Guilds intent throws the promised error.

## Existing coverage

I did not find an existing issue or PR title that appears to cover this exact root cause in the pre-submission check.

## Suggested fix

RequestMembers GuildMembers-intent validation

## Suggested tests

- Add a regression test for: The zero-intents configuration bypasses validation that correctly rejects an unrelated nonzero intent configuration.
- Include the boundary value or state transition described above so the old behavior fails before the fix.

---
Submitted with Codex.

Fleny113:


AwesomeStickz (approved):

packages/gateway/src/manager.ts
```diff
@@ -450,7 +450,7 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
     async requestMembers(guildId, options) {
       const shardId = gateway.calculateShardId(guildId);
 
-      if (gateway.intents && (!options?.limit || options.limit > 1) && !(gateway.intents & GatewayIntents.GuildMembers))
+      if ((!options?.limit || options.limit > 1) && (gateway.intents & GatewayIntents.GuildMembers) === 0)
```
AwesomeStickz, line 453: Why not `!(gateway.intents & GatewayIntents.GuildMembers)`?

Fleny113: checking for 0 feels more intentional to me

might be a me thing tho

AwesomeStickz: Fair, just wanted to know

PR #5241, opened 2026-07-27
ci: Improve caching for Docusaurus build

Fleny113:
Update the CI for the website to optimize it a bit:
- Cache separately the root deps and the website deps
- Don't build the .js files for the packages, typedoc only needs the .d.ts files
- Move the creation of the second actions/cache backed turbo remote cache to remove the need to manually specify the turbo server
- Don't install root deps on website typecheck: it's not needed

From the CI in this pr it looks like that yarn install is the longest step, so not much we can do there

### DIFF
--- a/.github/workflows/site.yml
+++ b/.github/workflows/site.yml
@@ -20,26 +20,30 @@ jobs:
         with:
           node-version: 22
           cache: yarn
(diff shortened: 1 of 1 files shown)


AwesomeStickz (approved):
🏅

PR #5242, opened 2026-07-29
Setup actionlint to detect GitHub Actions workflow issues

chernojagne:
This adds [actionlint](https://github.com/rhysd/actionlint) to catch GitHub Actions workflow issues — invalid expressions, unknown action inputs, and type mismatches — before they reach CI.

**What's added:**

- `.github/actionlint.yaml` — config with ShellCheck enabled and four common GHA false-positive rules silenced (SC2086, SC2129, SC2155, SC1091)
- `.github/workflows/actionlint.yml` — dedicated workflow that runs actionlint on push and PR when workflow files or the config change

The five existing workflows include nested expressions, matrix strategies, shell scripts, and third-party actions — all patterns actionlint is designed to validate.

---

This change was generated by [Autar](https://autar.ai) and reviewed by a human before submission.

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,0 +1,14 @@
+# actionlint configuration
+# https://github.com/rhysd/actionlint/blob/main/docs/config.md
+self-hosted-runner:
+  labels: []
+
+# ShellCheck rules disabled inside run: blocks.
+# Each `run:` block is checked through ShellCheck; suppress noisy rules here.
+paths:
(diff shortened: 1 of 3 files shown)


Fleny113:
We have run `actionlint` locally and addressed it warnings in #5243
however we'd rather avoid having it run in CI every time for the time being

Also considering it would always report an error has it hasn't been updated to consider valid the "queue: max" option in the concurrency options (see https://github.com/orgs/community/discussions/12835#discussioncomment-16602100)

PR #5249, opened 2026-07-31
fix(rest): re-send a request when a gateway in front of the api answered

NotDemonix:
Split out of #5158, which is about not re-sending requests that may already have run. This is the other side of it: there is a case we currently do not re-send where we safely could.

`sendRequest` only retries a 429. Every other error status rejects right away, server errors included, so a bot gets a failed call whenever Discord has a wobble.

Discord gives a 502 its own row in the [http response codes table](https://docs.discord.com/developers/topics/opcodes-and-status-codes), separate from the generic 5xx one, and it reads "There was not a gateway available to process your request. Wait a bit and retry." That sentence is already in the library, it is the message we build for a 502 in `sendRequest`, and then nothing waits or retries.

A 502 comes from a gateway sitting in front of the api rather than from the api itself. The request never reached anything that could act on it, so re-sending it cannot execute it twice. A 503 is not in Discord's table but is the same shape, the server is not taking requests, so it is re-sent too.

500 and 504 are deliberately left out. Both mean the api did get the request: a 500 is it failing while handling it, and a 504 is it not answering in time, so either may have carried the request out. Re-sending those is how you end up with two channels for one call, which is the thing #5158 exists to prevent.

Discord sends no `retry-after` with a server error, so it backs off a little further on each attempt. The counter is the one the rate limit path already uses, and there is a small cap on top of it because `maxRetryCount` is Infinity by default and an outage would otherwise keep every request alive for as long as it lasts. Happy to promote that cap to an option if you would rather have it configurable, I left it as a constant to avoid adding a knob for something that has a sensible answer.

Tests cover all four statuses and the retry cap.

### DIFF
--- a/packages/rest/src/manager.ts
+++ b/packages/rest/src/manager.ts
@@ -573,6 +573,22 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
         rest.logger.debug(`Request to ${url} failed.`);
 
         if (response.status !== HttpResponseCode.TooManyRequests) {
+          // A gateway in front of the api answered instead of the api itself, so the request never got to run and re-sending it cannot
+          // execute anything twice. `retryCount` is shared with the rate limit path below, and the extra cap keeps a Discord outage from
+          // keeping every request alive, since `maxRetryCount` is Infinity by default.
+          if (
+            retryableServerErrors.has(response.status) &&
+            options.retryCount < rest.maxRetryCount &&
+            options.retryCount < maxServerErrorRetryCount
+          ) {
+            options.retryCount += 1;
+            rest.logger.debug(`Request to ${url} got a ${response.status}, retrying.`);
+            // Discord sends no retry-after with these, so back off a little further on every attempt.
+            await delay(serverErrorRetryDelay * options.retryCount);
+
+            return await options.retryRequest?.(options);
+          }
+
           options.reject({ ok: false, status: response.status, statusText: response.statusText, body });
           return;
         }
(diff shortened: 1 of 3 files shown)


packages/rest/src/manager.ts
```diff
@@ -2029,6 +2045,25 @@ enum HttpResponseCode {
...
+ * either could create a second channel for one call.
+ */
+const retryableServerErrors = new Set([502, 503]);
+
+/** How many times a request is re-sent on one of {@link retryableServerErrors}, on top of `rest.maxRetryCount`. */
+const maxServerErrorRetryCount = 3;
+
+/** How much longer to wait before each further attempt, in milliseconds. Discord sends no `retry-after` with a server error. */
+const serverErrorRetryDelay = 1000;
+
```
Fleny113: for `maxServerErrorRetryCount` and `serverErrorRetryDelay` could be good to have them in the rest manager as options.
For `retryableServerErrors` put them at the top of the file (or if you want make it configurable too... if someone for some reason ever needs to)

Fleny113 (dismissed):
beside the vars at the bottom of manager.ts, LGTM

packages/rest/src/manager.ts
NotDemonix: Done. Both are on the manager now, next to `maxRetryCount`, as `maxServerErrorRetryCount` and `serverErrorRetryDelayStep`, and `retryableServerErrors` moved up to the constants at the top of the file as `RETRYABLE_SERVER_ERRORS`.

Renamed the delay to `...Step` because the nth attempt waits n times it, same naming as the proxy one in #5158.

Making them settable paid off in the tests too, the 502 and 503 cases set the step to 0 instead of really sitting out a 1s backoff each, and `maxServerErrorRetryCount` has a test of its own now.

Fleny113 (dismissed):
LGTM
I am going to wait Stickz review however

PR #5274, opened 2026-08-05
fix(rest): hand the rate limit slot back when an attempt gets no response

NotDemonix:
Follow-up to #5158.

A queue spends a rate limit slot before it sends, and only ever gets one back from the rate limit headers of a response, or from the timer `processPending` arms when it runs out of them, which needs an `interval` that only such a response teaches it. A queue that has not had a response yet has neither, so an attempt that ends without one leaves the slot spent for good. Nothing is ever sent again, so no response can arrive to restore it, and `isQueueClearable` keeps saying no, so the queue is not deleted either. Everything routed through it after that waits forever.

An abort and a `requestTimeout` are both exactly that kind of attempt, so the two things #5085 and #5158 added are what turned this from something you would rarely hit into something you hit routinely. The counting itself is older than both.

Queues are keyed by token plus the route with its major parameter kept, which is what makes this look so odd in production: one aborted `POST /guilds/{id}/channels` strands ticket creation in that one guild forever while every other guild carries on. The bot side only sees its own `requestTimeout` fire over and over until it runs out of re-sends, and the rest process logs nothing at all, because the request never gets far enough to fail.

The timeout path deadlocks outright on top of that. `handleTimeout` re-sends through `retryRequest` -> `processRequest` -> `Queue.makeRequest` -> `waitUntilRequestAvailable`, which is reached from inside the `processPending` loop that is awaiting `sendRequest`, so the re-send waits on the very slot the attempt it is replacing is still holding. That request never settles, not even to reject once `maxRetryCount` is exhausted.

So with this PR:

- the slot goes back in the fetch catch, right next to the `invalidBucket` bookkeeping that is already there. That covers aborts, timeouts and failed connections in one place, and it happens before a timeout is re-sent, which is what the deadlock needs
- nothing changes for an attempt that did get a response. `processHeaders` stays the only thing that applies what Discord actually said, and a queue that has not learned a bucket yet still sends one request at a time while it waits to find out
- `Queue.handleFailedRequest` is the counterpart to `handleCompletedRequest`, so the two cases sit next to each other
- two tests, one per failure mode. Without the fix the first leaves `remaining` at 0 and the second never settles at all

Reproduced against a local rest process before and after: a guild that could not create a channel again for as long as requests kept arriving now recovers immediately.


Fleny113 (approved):
LGTM. i fear that a short timeout could mess this up, however there really isn't much you can do about it i think

I am going to wait for stickz to review this as well

NotDemonix:
> LGTM. i fear that a short timeout could mess this up, however there really isn't much you can do about it i think
> 
> I am going to wait for stickz to review this as well

Works perfectly fine in prod for Tickety - no more stuck queues so seems to be fine

packages/rest/tests/unit/manager.spec.ts
```diff
@@ -526,4 +526,61 @@ describe('[rest] manager', () => {
       expect(error.message).to.be.equal('[999] The request was aborted.');
     });
   });
+
+  // A queue spends a rate limit slot before it sends, and the only thing that hands one back is the rate limit headers of a response, plus the
+  // timer that needs an `interval` those same headers teach it. A queue that has not had a response yet has neither, so an attempt ending
+  // without one used to leave the slot spent for good and nothing routed through that queue was ever sent again.
+  describe('rest.makeRequest when an attempt ends without a response', () => {
```

Fleny113 (changes requested):

packages/rest/tests/unit/manager.spec.ts
NotDemonix: Removed. The two source changes in `manager.ts` and `queue.ts` are unchanged.

PR #5319, opened 2026-08-15
fix: Don't create a new bucket on HELLO 

Fleny113:
This fixes also the issue where integration tests took a long time to exit despite finishing all the tests: this is because the bucket created in the constructor of the shard had to naturally clear which takes 1 minute.

> [!WARNING]
> Since this does changes to the shard code i will test this for a while, if someone is available to test this it is appreciated


Fleny113:
This has been stable in my own bot, so marking as ready for review

PR #5366, opened 2026-08-28
fix(bot): stop message toggles from overwriting each other

NotDemonix:
`baseMessage` keeps `tts`, `mentionEveryone` and `pinned` in one `ToggleBitfield`, using the bit **values** 1, 2 and 3 (`message.ts:128`, `:85`, `:93`). But `ToggleBitfield.contains` is an any-bit test:

```ts
contains(bits: number): boolean {
  return Boolean(this.bitfield & bits);
}
```

So 3 is not a third flag — `0b11` is `tts | mentionEveryone`. All three are set straight from the gateway payload (`message.ts:194`, `:231`, `:251`), so this happens on ordinary `MESSAGE_CREATE`:

| payload from Discord | `.tts` | `.mentionEveryone` | `.pinned` |
| --- | --- | --- | --- |
| `{ tts: true }` | `true` | `false` | **`true`** |
| `{ mention_everyone: true }` | `false` | `true` | **`true`** |
| `{ pinned: true }` | **`true`** | **`true`** | `true` |

and setting `pinned = false` clears the other two, because `remove(3)` unsets both bits.

`pinned` needs the next free bit, `4`.

---

### DIFF
--- a/packages/bot/src/transformers/message.ts
+++ b/packages/bot/src/transformers/message.ts
@@ -90,12 +90,12 @@ export const baseMessage: Message = {
     else this.bitfield.remove(2);
   },
   get pinned() {
-    return this.bitfield?.contains(3) ?? false;
+    return this.bitfield?.contains(4) ?? false;
   },
   set pinned(value: boolean) {
     if (!this.bitfield) return;
-    if (value) this.bitfield.add(3);
-    else this.bitfield.remove(3);
+    if (value) this.bitfield.add(4);
+    else this.bitfield.remove(4);
   },
   get sourceMessageDeleted() {
     return this.flags?.contains(MessageFlags.SourceMessageDeleted) ?? false;


packages/bot/tests/unit/transformers/message.spec.ts
```diff
@@ -0,0 +1,38 @@
+import type { DiscordMessage } from '@discordeno/types';
```
Fleny113: we don't need a test to check that we don't overlap tts, metion_everyone and pinned...

Also "bot = {} as unknown as Bot" is an horrible idea

Fleny113 (changes requested):

packages/bot/tests/unit/transformers/message.spec.ts
NotDemonix: Dropped, along with the `{} as unknown as Bot` cast. The bit change stands on its own.

PR #5367, opened 2026-08-28
fix(bot): do not recurse forever on a sub command with no options

NotDemonix:
`commandOptionsParser` recurses into sub commands with `option.options`:

```ts
if (!options) options = interaction.data.options ?? [];      // :17
...
case ApplicationCommandOptionTypes.SubCommand:
  args[option.name] = commandOptionsParser(interaction, option.options);   // :25
```

Discord omits the `options` key entirely for a sub command invoked with no arguments, and `transformInteractionDataOption` copies that absence straight through (`interaction.ts:210-216`, no `?? []`). So the recursive call gets `undefined`, line 17 re-seeds `options` from the **root** `interaction.data.options`, the loop reaches the same sub command again, and recurses with `undefined` again. There is no base case and no depth bound.

Feeding the real exported function a `/config reset` shaped payload:

```
SubCommand, options key ABSENT       -> THREW RangeError: Maximum call stack size exceeded
SubCommand, options: []              -> {"reset":{}}
SubCommandGroup, options key ABSENT  -> THREW RangeError: Maximum call stack size exceeded
SubCommand with a nested string opt  -> {"set":{"k":"v"}}
```

`[]` is truthy so it terminates correctly — only an absent key triggers it. Thrown from inside the async `INTERACTION_CREATE` handler, it surfaces as an unhandled rejection rather than a per-interaction failure.

`option.options ?? []` fixes it and keeps line 17's fallback reserved for the genuine top-level call.

Also in here: `InteractionResolvedDataUser.member` was declared required, but `resolved.members` is absent in DMs, so the parser hands back `undefined` under a non-optional type. Made optional.

---

### DIFF
--- a/packages/bot/src/commandOptionsParser.ts
+++ b/packages/bot/src/commandOptionsParser.ts
@@ -22,7 +22,7 @@ export function commandOptionsParser<
     switch (option.type) {
       case ApplicationCommandOptionTypes.SubCommandGroup:
       case ApplicationCommandOptionTypes.SubCommand:
-        args[option.name] = commandOptionsParser(interaction, option.options) as InteractionResolvedData<TProps, TBehavior>;
+        args[option.name] = commandOptionsParser(interaction, option.options ?? []) as InteractionResolvedData<TProps, TBehavior>;
         break;
       case ApplicationCommandOptionTypes.Channel:
         args[option.name] = interaction.data.resolved?.channels?.get(BigInt(option.value!)) as InteractionResolvedData<TProps, TBehavior>;
@@ -70,7 +70,7 @@ export type InteractionResolvedData<TProps extends TransformersDesiredProperties
 
 export interface InteractionResolvedDataUser<TProps extends TransformersDesiredProperties, TBehavior extends DesiredPropertiesBehavior> {
   user: SetupDesiredProps<User, TProps, TBehavior>;
-  member: InteractionResolvedDataMember<TProps, TBehavior>;
+  member?: InteractionResolvedDataMember<TProps, TBehavior>;
 }
 
 export type InteractionResolvedDataChannel<TProps extends TransformersDesiredProperties, TBehavior extends DesiredPropertiesBehavior> = Pick<


packages/bot/src/commandOptionsParser.ts
```diff
@@ -70,7 +70,7 @@ export type InteractionResolvedData<TProps extends TransformersDesiredProperties
 
 export interface InteractionResolvedDataUser<TProps extends TransformersDesiredProperties, TBehavior extends DesiredPropertiesBehavior> {
   user: SetupDesiredProps<User, TProps, TBehavior>;
-  member: InteractionResolvedDataMember<TProps, TBehavior>;
+  member?: InteractionResolvedDataMember<TProps, TBehavior>;
```
Fleny113, line 73: unrelated fix, but ok

packages/bot/tests/unit/commandOptionsParser.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+        type: ApplicationCommandOptionTypes.SubCommand,
+        name: 'set',
+        options: [{ type: ApplicationCommandOptionTypes.String, name: 'key', value: 'value' }] as InteractionDataOption[],
+      },
+    ]);
+
+    expect(commandOptionsParser<Props, Behavior>(interaction)).to.be.deep.equal({ set: { key: 'value' } });
+  });
+
+  it('types the resolved member of a user option as optional', () => {
```
Fleny113: this makes makes NO sense, you are asserting that an empty object doesn't have `member`
and you are asserting on typescript to make this valid in the processes, proving that TS is already catching this, no need for a unit test

packages/bot/tests/unit/commandOptionsParser.spec.ts
```diff
@@ -0,0 +1,47 @@
+import { ApplicationCommandOptionTypes } from '@discordeno/types';
```
Fleny113: These tests are really not needed, fixing it is fine, but it's not something that needs to be tested for.

Fleny113 (changes requested):

packages/bot/tests/unit/commandOptionsParser.spec.ts
NotDemonix: Agreed, that one was asserting the type system. Dropped.

packages/bot/tests/unit/commandOptionsParser.spec.ts
NotDemonix: Dropped the whole file. The two source lines are unchanged.

PR #5368, opened 2026-08-28
fix(bot): route THREAD_MEMBER_UPDATE to its own handler

NotDemonix:
Three handler-layer mistakes.

**`THREAD_MEMBER_UPDATE` is wired to the plural handler** (`handlers.ts:67`):

```ts
THREAD_MEMBER_UPDATE:  _options.THREAD_MEMBERS_UPDATE ?? handlers.handleThreadMembersUpdate,
THREAD_MEMBERS_UPDATE: _options.THREAD_MEMBERS_UPDATE ?? handlers.handleThreadMembersUpdate,
```

Both the override key and the default are the plural ones. `handleThreadMemberUpdate` exists (`handlers/channels/THREAD_MEMBER_UPDATE.ts`) and `threadMemberUpdate` is declared in `events.ts:44`, but neither can ever fire — and a user passing `THREAD_MEMBER_UPDATE` in `options.handlers` is silently ignored while their `THREAD_MEMBERS_UPDATE` override gets called for both events.

**`READY` skips the bot id assignment when no listener is registered.** `handleReady` returns early on `if (!bot.events.ready) return`, and the `bot.id` / `bot.applicationId` assignment sat *after* that guard, so a bot with no `ready` listener never learned its own id. Moved above the guard — which also means `bot.id` is now populated *during* the `ready` event rather than just after it.

**`THREAD_LIST_SYNC` hand-built its thread members** instead of calling `bot.transformers.threadMember`, so desired properties and any customizer were bypassed for that batch only.

---

### DIFF
--- a/packages/bot/src/handlers/channels/THREAD_LIST_SYNC.ts
+++ b/packages/bot/src/handlers/channels/THREAD_LIST_SYNC.ts
@@ -12,11 +12,6 @@ export async function handleThreadListSync(bot: Bot, data: DiscordGatewayPayload
     guildId,
     channelIds: payload.channel_ids?.map((id) => bot.transformers.snowflake(id)),
     threads: payload.threads.map((thread) => bot.transformers.channel(bot, thread, { guildId })),
-    members: payload.members.map((member) => ({
-      id: member.id ? bot.transformers.snowflake(member.id) : undefined,
-      userId: member.user_id ? bot.transformers.snowflake(member.user_id) : undefined,
-      joinTimestamp: Date.parse(member.join_timestamp),
-      flags: member.flags,
-    })),
+    members: payload.members.map((member) => bot.transformers.threadMember(bot, member, { guildId: payload.guild_id })),
   });
 }
--- a/packages/bot/src/handlers/misc/READY.ts
+++ b/packages/bot/src/handlers/misc/READY.ts
@@ -2,9 +2,13 @@ import type { DiscordGatewayPayload, DiscordReady } from '@discordeno/types';
 import type { Bot } from '../../bot.js';
 
 export async function handleReady(bot: Bot, data: DiscordGatewayPayload, shardId: number): Promise<void> {
+  const payload = data.d as DiscordReady;
+
+  bot.id = bot.transformers.snowflake(payload.user.id);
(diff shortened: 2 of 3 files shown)


packages/bot/src/handlers/misc/READY.ts
```diff
@@ -18,7 +22,4 @@ export async function handleReady(bot: Bot, data: DiscordGatewayPayload, shardId
     },
     payload,
   );
-
-  bot.id = bot.transformers.snowflake(payload.user.id);
-  bot.applicationId = bot.transformers.snowflake(payload.application.id);
```
Fleny113, line 23: unrelated fix, but ok sure, the return does prevent id and application id from updating

packages/bot/src/handlers/channels/THREAD_LIST_SYNC.ts
```diff
@@ -12,11 +12,6 @@ export async function handleThreadListSync(bot: Bot, data: DiscordGatewayPayload
     guildId,
     channelIds: payload.channel_ids?.map((id) => bot.transformers.snowflake(id)),
     threads: payload.threads.map((thread) => bot.transformers.channel(bot, thread, { guildId })),
-    members: payload.members.map((member) => ({
-      id: member.id ? bot.transformers.snowflake(member.id) : undefined,
-      userId: member.user_id ? bot.transformers.snowflake(member.user_id) : undefined,
-      joinTimestamp: Date.parse(member.join_timestamp),
-      flags: member.flags,
-    })),
+    members: payload.members.map((member) => bot.transformers.threadMember(bot, member, { guildId: payload.guild_id })),
```
Fleny113, line 15: unrelated fix but ok

packages/bot/tests/unit/handlers.spec.ts
```diff
@@ -0,0 +1,112 @@
+import type { DiscordGatewayPayload, GatewayDispatchEventNames } from '@discordeno/types';
```
Fleny113: remove this test

Fleny113 (changes requested):

packages/bot/tests/unit/handlers.spec.ts
NotDemonix: Removed.

Issue #123, opened 2020-10-01
[Docs] Add A Guide on Visual Studio Code


Skillz4Killz:
**Is your feature request related to a problem? Please describe.**
Our current guide assumes that the user has an understanding and has already installed and setup VSC. However, a lot of bot developers are just starting to want to learn and this would be a useful part of the guide.

**Describe the solution you'd like**
- What is VSC?
- Installing VSC
- Opening VSC
- Opening A Terminal
- Moving To A Folder

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

chroventer:
Instead of creating a guide about the entire environment setup, we can have a getting started page which would have curated redirect links to installation page for Deno, setting up a project (Visual Studio Code Docs), etc.
What are your thoughts on it?

Skillz4Killz:
That is a great idea but ideally id also like to see links to youtube tutorials on them in that case of how to set each thing up. A lot of developers that makes bots are doing this for the first time, seeing it in detail whether written on a guide or just watching a youtube video installing and setting it up would be perfect solution.

chroventer:
On it!

Suyashtnt:
Ive added this exact thing in my docs for discordeno helper, I'll make a PR that adds it to this docs

ayntee:
Merged into #211

PR #5369, opened 2026-08-28
fix(bot): correct member, guild and message transformers

NotDemonix:
Three independent one-liners.

**`reverse.member` wrote `"[object Object]"` for permissions.** `permissions` is a `ToggleBitfieldBigint`, and `_payload.permissions?.toString()` hits `Object.prototype.toString`. The sibling flag fields already use `toJSON()`, which returns `this.bitfield.toString()`. Round-tripping a member through `transformers.member` → `transformers.reverse.member` produced `permissions: "[object Object]"`.

**`guild.features` never reported `SOUNDBOARD`.** `GuildToggle.soundboard` exists (`1n << 25n`) and the bit *is* set from `GuildFeatures.Soundboard` (`toggles/guild.ts:158`), but the `features` getter filters by membership in `guildFeatureNames`, and `'soundboard'` was missing from that list. So the toggle read `true` while the feature was absent from the array. (Position in the list carries no meaning — it is a membership filter, not a bit index.)

**`mentionedChannelIds` returned duplicates.** It concatenates the ids Discord sends in `mention_channels` with the ids it scrapes out of `content`, and any channel appearing in both was listed twice. Now deduplicated.

The scan is also gated behind `payload.mention_channels?.length`, so it never runs for an ordinary message. I had widened that condition here; per review that is a separate call about whether the content scan should exist at all, so this now leaves the condition alone and only fixes the duplicates.

---

### DIFF
--- a/packages/bot/src/transformers/message.ts
+++ b/packages/bot/src/transformers/message.ts
@@ -194,13 +194,15 @@ export function transformMessage(bot: Bot, payload: Partial<DiscordMessage>, ext
   if (payload.mention_everyone) message.mentionEveryone = true;
   if (props.mentionedChannelIds && payload.mention_channels?.length) {
     message.mentionedChannelIds = [
-      // Keep any ids tht discord sends
-      ...(payload.mention_channels ?? []).map((m) => bot.transformers.snowflake(m.id)),
-      // Add any other ids that can be validated in a channel mention format
-      ...(payload.content?.match(CHANNEL_MENTION_REGEX) ?? []).map((text) =>
-        // converts the <#123> into 123
-        bot.transformers.snowflake(text.substring(2, text.length - 1)),
-      ),
+      ...new Set([
+        // Keep any ids tht discord sends
+        ...(payload.mention_channels ?? []).map((m) => bot.transformers.snowflake(m.id)),
+        // Add any other ids that can be validated in a channel mention format
+        ...(payload.content?.match(CHANNEL_MENTION_REGEX) ?? []).map((text) =>
+          // converts the <#123> into 123
+          bot.transformers.snowflake(text.substring(2, text.length - 1)),
+        ),
+      ]),
     ];
   }
   if (props.mentionedRoleIds && payload.mention_roles?.length)
--- a/packages/bot/src/transformers/reverse/member.ts
+++ b/packages/bot/src/transformers/reverse/member.ts
@@ -33,7 +33,7 @@ export function transformMemberToDiscordMember(bot: Bot, payload: typeof bot.tra
     joined_at: _payload.joinedAt ? new Date(_payload.joinedAt).toISOString() : null,
     premium_since: _payload.premiumSince ? new Date(_payload.premiumSince).toISOString() : undefined,
     avatar: _payload.avatar ? iconBigintToHash(_payload.avatar) : undefined,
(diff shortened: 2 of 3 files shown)


packages/bot/tests/unit/transformers/reverseMember.spec.ts
```diff
@@ -0,0 +1,18 @@
+import type { DiscordMember } from '@discordeno/types';
```
Fleny113: no.

packages/bot/tests/unit/transformers/messageMentionedChannelIds.spec.ts
```diff
@@ -0,0 +1,28 @@
+import type { DiscordMessage } from '@discordeno/types';
```
Fleny113: no.

packages/bot/tests/unit/transformers/guildToggles.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { type DiscordGuild, GuildFeatures } from '@discordeno/types';
```
Fleny113: no.

packages/bot/src/transformers/message.ts
```diff
@@ -192,15 +192,17 @@ export function transformMessage(bot: Bot, payload: Partial<DiscordMessage>, ext
...
-      ),
+      ...new Set([
+        // Keep any ids tht discord sends
+        ...(payload.mention_channels ?? []).map((m) => bot.transformers.snowflake(m.id)),
+        // Add any other ids that can be validated in a channel mention format
+        ...(payload.content?.match(CHANNEL_MENTION_REGEX) ?? []).map((text) =>
+          // converts the <#123> into 123
+          bot.transformers.snowflake(text.substring(2, text.length - 1)),
+        ),
+      ]),
```
Fleny113, line 205: the fix is valid i'd say, however looking at this we provably don't want to do this actually

Fleny113 (changes requested):

packages/bot/tests/unit/transformers/reverseMember.spec.ts
NotDemonix: Dropped.

packages/bot/tests/unit/transformers/messageMentionedChannelIds.spec.ts
NotDemonix: Dropped.

packages/bot/tests/unit/transformers/guildToggles.spec.ts
NotDemonix: Dropped.

packages/bot/src/transformers/message.ts
NotDemonix: Scoped it down to just the duplicates, which is the part you called valid.

I had also widened the condition so the content scan ran on a message with `content` but no `mention_channels` — which is every ordinary message, since Discord only populates `mention_channels` for crossposts. That made a nearly dead branch live, and that is the behaviour you are questioning, so I have reverted it. The condition is back to `payload.mention_channels?.length`.

Whether scraping `<#123>` out of the content should be there at all is your call. Happy to strip it entirely in a follow-up if you want it gone — just say so.

PR #5370, opened 2026-08-28
fix(bot): keep resolved members and roles on select menus

NotDemonix:
`transformComponent` never had a `guildId` to hand to `interactionDataResolved`, so `transformInteractionDataResolved` could not build the `members` and `roles` collections — both need a guild id to construct their entries.

The visible effect: a user, role or mentionable select **submitted through a modal** comes back with `resolved.users` populated but `resolved.members` and `resolved.roles` missing. The same select sent as a message component is unaffected, because that path reaches `interactionDataResolved` through `transformInteraction`, which does have the guild id.

This adds an optional `ComponentTransformerExtra { guildId?: BigString }` and threads it from `transformInteraction` down through the container/action-row/section/label recursion to the four select transformers. Additive only — `extra` was already optional, and every internal call site passes it along.

---

### DIFF
--- a/packages/bot/src/transformers.ts
+++ b/packages/bot/src/transformers.ts
@@ -101,7 +101,12 @@ import { transformAutoModerationActionExecution } from './transformers/automodAc
 import { transformAutoModerationRule } from './transformers/automodRule.js';
 import { transformAvatarDecorationData } from './transformers/avatarDecorationData.js';
 import { transformChannel, transformForumTag } from './transformers/channel.js';
-import { transformComponent, transformMediaGalleryItem, transformUnfurledMediaItem } from './transformers/component.js';
+import {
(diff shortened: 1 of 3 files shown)


packages/bot/tests/unit/transformers/component.spec.ts
```diff
@@ -0,0 +1,46 @@
+import { type DiscordInteraction, InteractionTypes, MessageComponentTypes } from '@discordeno/types';
```
Fleny113: no.

Fleny113 (changes requested):

packages/bot/tests/unit/transformers/component.spec.ts
NotDemonix: Dropped.

PR #5371, opened 2026-08-28
fix(rest): correct routes and methods that do not match the API

NotDemonix:
Seven places where the request we build is not the request Discord documents. Each is independent.

**Routes** (`routes.ts`)

| route | built | should be |
| --- | --- | --- |
| `soundboard.sendSound` | `/channels/{id}` | `/channels/{id}/send-soundboard-sound` |
| `channels.reactions.message` | `?type=1after=852…` | `?type=1&after=852…` |
| `guilds.events.event` | `/guilds/1/scheduled-events/2with_user_count=true` | `…/2?with_user_count=true` |
| `guilds.roles.memberCounts` | `/guilds/${guildId}/…` — no `encode()` | `/guilds/${encode(guildId)}/…` |

`sendSound` posts to a path with no `POST` handler. The missing `&` only bites when both `type` and `after` are given. The missing `?` corrupts the event id segment — note the sibling `events.events` five lines up does have it. `memberCounts` is the only route in the file that skips `encode()`.

**Helpers** (`manager.ts`)

- `executeWebhook` passed `body: options` but not `files`, so webhook attachments were dropped and the request went out as `application/json`. `sendMessage` one line away already does `{ body, files: body.files }`.
- `modifyGuildSoundboardSound` used `rest.post`; Discord documents `PATCH /guilds/{guild.id}/soundboard-sounds/{sound.id}`.
- `getReactions` was the only one of the five reaction methods not calling `processReactionString`, so it rejected the `<:name:id>` form its siblings accept. `processReactionString` only rewrites strings starting with `<:` / `<a:`, so unicode and `name:id` inputs are untouched.

---

### DIFF
--- a/packages/rest/src/manager.ts
+++ b/packages/rest/src/manager.ts
@@ -1271,6 +1271,7 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
     async executeWebhook(webhookId, token, options) {
       return await rest.post<DiscordMessage>(rest.routes.webhooks.webhook(webhookId, token, options), {
         body: options,
+        files: options.files,
         unauthorized: true,
       });
     },
@@ -1615,6 +1616,8 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
     },
 
     async getReactions(channelId, messageId, reaction, options) {
+      reaction = processReactionString(reaction);
+
       return await rest.get<DiscordUser[]>(rest.routes.channels.reactions.message(channelId, messageId, reaction, options));
     },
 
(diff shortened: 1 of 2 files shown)


packages/rest/tests/unit/routes.spec.ts
```diff
@@ -0,0 +1,32 @@
+import { expect } from 'chai';
```
Fleny113: no.

packages/rest/tests/unit/manager.spec.ts
```diff
@@ -526,4 +526,45 @@ describe('[rest] manager', () => {
       expect(error.message).to.be.equal('[999] The request was aborted.');
     });
   });
+
+  describe('rest helper methods', () => {
```

Fleny113 (changes requested):

packages/rest/tests/unit/routes.spec.ts
NotDemonix: Dropped, so this no longer creates the file #5293 also adds.

packages/rest/tests/unit/manager.spec.ts

PR #5372, opened 2026-08-28
fix(rest): keep a queue alive when a request throws

NotDemonix:
A queue spends a rate limit slot before it sends. `sendRequest` hands that slot back from the rate limit headers of a response — so an attempt that **throws** instead of getting a response leaves the slot spent for good, the queue can never send again, and every later request on that route waits forever.

Reproducing it needs nothing exotic: an audit-log `reason` containing a lone surrogate (`'cleanup \ud83d'`) makes `encodeURIComponent` throw inside `createRequestBody` — after the queue has spent its slot, and outside everything `sendRequest` guards.

Three fixes, one theme:

- **`queue.ts:161`** — the `catch` around `sendRequest` only rejected the request. It now also hands back the rate limit slot and the `invalidBucket` slot the attempt spent. Before: a second, valid request on the same route never reached `fetch` and never settled; the mocha process would not even exit.
- **`manager.ts:811`** — `new Promise(async (resolve, reject) => …)`. A promise returned from an executor is dropped, so a throw out of `processRequest` left the caller hanging *and* raised an unhandled rejection. The executor is no longer `async`; the rejection is routed to `payload.reject`.
- **`manager.ts:670`** — the `runThroughQueue: false` branch never called `invalidBucket.waitUntilRequestAvailable()`, but `sendRequest` reports the request as completed either way. Every such request drove `activeRequests` one lower; it read `-1` after a single ordinary call, steadily raising the invalid-request ceiling the bucket exists to enforce.

**Relationship to #5274:** complementary, not overlapping. #5274 covers *`fetch` rejected* (inside `sendRequest`'s fetch catch, which returns normally and never reaches this one); this covers *`sendRequest` threw*. I deliberately did not add a `handleFailedRequest` method or touch the fetch catch. If #5274 lands first, this branch's inline `this.remaining++` could call `handleFailedRequest()` instead — a one-line follow-up.

---

### DIFF
--- a/packages/rest/src/manager.ts
+++ b/packages/rest/src/manager.ts
@@ -667,7 +667,17 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
       const url = rest.simplifyUrl(request.route, request.method);
 
       if (request.runThroughQueue === false) {
-        await rest.sendRequest(request);
+        // This never reaches a queue, so this is the only place the invalid request bucket can be told about the request. `sendRequest` reports
+        // it as completed either way, and without this that report has nothing to match it: the bucket's count of active requests drifts one
+        // lower with every one of these, which raises the invalid request ceiling it exists to enforce.
+        await rest.invalidBucket.waitUntilRequestAvailable();
+
+        try {
+          await rest.sendRequest(request);
+        } catch (error) {
+          rest.invalidBucket.handleCompletedRequest(999, false);
+          throw error;
+        }
 
         return;
       }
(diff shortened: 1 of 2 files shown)


packages/rest/tests/unit/manager.spec.ts
```diff
@@ -526,4 +526,53 @@ describe('[rest] manager', () => {
       expect(error.message).to.be.equal('[999] The request was aborted.');
     });
   });
+
+  // A request that throws instead of failing with a response used to leave the state it was given behind: the queue kept the rate limit slot it
+  // spent, the invalid request bucket kept one of its own, and a caller that skipped the queue was never told anything at all.
+  describe('rest.makeRequest when a request throws', () => {
```

Fleny113 (changes requested):

packages/rest/tests/unit/manager.spec.ts

PR #5374, opened 2026-08-28
fix(rest): respect the global rate limit and keep queue keys stable

NotDemonix:
Three rate limit accounting bugs. This is the subtlest of the batch — please read the two flags at the bottom.

**`checkRateLimits` ignored the global limit when a shorter per-url one existed.** It returned on the first match:

```ts
if (ratelimited && now < ratelimited.resetTimestamp) return ratelimited.resetTimestamp - now;
if (global && now < global.resetTimestamp) return global.resetTimestamp - now;
```

With a 6541ms url wait and a 9849ms global wait it returned 6541, and `processPending` dispatched 3.3s inside the active global window. Now returns the longer of the two.

**The queue key for a message delete depends on wall-clock time.** `simplifyUrl` suffixes a `DELETE /channels/x/messages/y` with `message-delete`, `message-delete-10s` or `message-delete-2w` based on how old the message is *right now*. A request enqueued under `…-10s` whose response arrives 11 seconds later gets its rate limit credit applied to `…-2w` — a key nobody owns. The queue's `remaining` then stays `0` instead of taking the header's `4`; `remaining: 0` with `interval: 0` is the permanent wedge. The simplified url is now computed once, stored on `SendRequestOptions`, and reused.

**`updateTokenQueues`' merge branch deleted the live queue.** After moving the old queue's work onto the queue that already holds the new key, it calls `queue.cleanup()` on the *old* instance — whose `identifier` has already been rewritten, so the key it deletes is the merged queue's. `rest.queues.get(key)` returned `undefined` afterwards. `cleanup` now only deletes a key the manager actually holds *this* queue under. A request parked on the detached old queue was also never sent (fetch count 2 instead of 3); `Queue.makeRequest` now resolves the owning queue from `rest.queues` after waiting, and `updateTokenQueues` starts the merged queue's `processWaiting`/`processPending`, which nothing did before.

**Two things to flag.**

1. This updates an existing test, which is the one test change left in this PR. `'Will return URL time first if before resetTimestamp'` asserted the old precedence (6541 with a 9849ms global wait set). It is renamed to `'Will return the global time when it is the longer of the two'` and its expectation changed to 9849, because the behaviour it pinned is the behaviour being fixed. `checkRateLimits` is public API, so its return value genuinely changes for a caller holding both limits at once.
2. `SendRequestOptions` gains one optional field, `simplifiedUrl?: string`. Additive, but it is a public interface.

Deliberately left out: turning `processPending`'s single `if (urlResetIn) await delay(urlResetIn)` into a re-checking loop. `Math.max` alone fixes the reported defect; the loop changes queue timing more broadly and is a maintainer call. Does not touch the lines #5274 touches.

---

### DIFF
--- a/packages/rest/src/manager.ts
+++ b/packages/rest/src/manager.ts
@@ -148,15 +148,13 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
       const global = rest.rateLimitedPaths.get('global');
       const now = Date.now();
 
-      if (ratelimited && now < ratelimited.resetTimestamp) {
-        return ratelimited.resetTimestamp - now;
-      }
+      // Whichever of the two lasts longer is the one that decides when this request may go out, a shorter per url limit does not make it
+      // alright to send while the global one is still running.
+      const urlWait = ratelimited && now < ratelimited.resetTimestamp ? ratelimited.resetTimestamp - now : 0;
+      const globalWait = global && now < global.resetTimestamp ? global.resetTimestamp - now : 0;
+      const wait = Math.max(urlWait, globalWait);
 
-      if (global && now < global.resetTimestamp) {
-        return global.resetTimestamp - now;
-      }
-
-      return false;
+      return wait > 0 ? wait : false;
     },
 
     async updateTokenQueues(oldToken, newToken) {
@@ -206,6 +204,10 @@ export function createRestManager(options: CreateRestManagerOptions): RestManage
 
           queue.waiting = [];
           queue.pending = [];
+
+          // The merged queue is not processing anything of its own, so nothing would ever look at what it was just handed.
+          newQueue.processWaiting();
+          newQueue.processPending();
 
           queue.cleanup();
         } else {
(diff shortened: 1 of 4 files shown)


NotDemonix:
The red X on **Unit Tests** here is the codecov upload step, not the tests:

```
error -- Report creating failed: {"message":"Token required because branch is protected"}
##[error]Process completed with exit code 1.
```

The suite itself is green in that same job — `Tasks: 12 successful, 12 total`, every package passing. It looks like the usual fork-PR codecov token flake (my other PRs in this batch uploaded fine). I do not have permission to re-run the job, so it needs a maintainer re-run or a push to clear it.

NotDemonix:
Dropped the added `manager.spec.ts` block here too, for the same reason it was rejected on #5371-#5373.

The one test change left is the **update** to the existing `checkRateLimits` case, which is not optional: that test asserted the shorter url wait was returned while a longer global limit was still running, so it pins the exact behaviour this PR fixes and fails against it. It is renamed and its expectation flipped to 9849. Details in the body.

PR #5375, opened 2026-08-28
fix(gateway): settle connect() and honour close() during the handshake

NotDemonix:
Five defects in the shard socket lifecycle. They share the same code, so they are here together; each is separable in review.

**`connect()` never settles when the handshake fails.** The promise resolved only from `socket.onopen`. A socket that closes before it ever opened — unreachable url, rejected handshake, a `close()` during `CONNECTING` — leaves it pending forever, and with it `identify()` and `spawnShards()`. Against an unreachable url the shard self-healed and reached READY while the original `connect()` call stayed pending (mocha timed out at 8s). Now settled from the close handler too.

**`close()` is a silent no-op while the socket is `CONNECTING`.** The guard was `readyState !== WebSocket.OPEN`. So `shutdown()` on a shard mid-handshake returned without closing anything, the handshake completed, and the shard identified and heartbeated *after* it had been shut down. `CONNECTING` is now closed as well, which aborts the handshake. This is also the root cause of `gateway.shutdown()` leaving mid-reconnect shards alive.

**`handleClose()` threw out of `socket.onclose`.** On 4004/4010/4013/4014 it did `throw new Error(close.reason || …)` from a fire-and-forget handler — nothing can catch it. `handleClose({code: 4014})` produced `Error: Disallowed intent(s) at DiscordenoShard.handleClose (Shard.ts:419)` with no reachable handler. Now logs at `fatal` and fires `events.disconnected`.

**Payloads parked while offline never settle.** `offlineSendQueue` is drained on READY/RESUMED only, so a `send()` parked before a permanent shutdown hung forever. The queue is now drained on the terminal paths too, and those parked promises reject.

**A payload could be written before IDENTIFY.** `checkOffline` gated on socket `readyState`, not session state, so a payload sent from the `connected` event reached the server ahead of the Identify — Discord answers that with close 4003. Recorded server op order was `[RequestGuildMembers, Identify]`; it is now `[Identify, RequestGuildMembers]`.

**Behaviour changes worth a look:**
- `send()`/`sendPayload()` can now **reject** when the shard is permanently offline. Previously they hung, so nothing could have depended on the old behaviour, but it is observable.
- `handleClose` no longer throws on unrecoverable codes. I did not add a new public event for this — that felt like a maintainer's call.
- The `goingOffline` check moved above the switch so it applies to every close code, which meant splitting `Shutdown`/`Resharded` from `ReIdentifying`/`ResumeClosingOldConnection`.
- `connect()` ignores duplicate close events: Node's built-in `WebSocket` dispatches close twice for a closed `CONNECTING` socket (code 0, then 1002), which otherwise made the shard reconnect straight after a shutdown.
- Additive API only: `send()`/`checkOffline()` gain an optional trailing parameter; `canSend()` and `drainOfflineSendQueue()` are new.

I did **not** swap `shard.close(code, reason)` for `shard.shutdown()` in `gateway.shutdown()` as one reading suggested — `Shard.shutdown()` hardcodes `ShardSocketCloseCodes.Shutdown` and would discard the `TestingFinished`/`Resharded` codes the manager passes.

Five integration tests added; all five fail on `main` (three time out, one throws, one asserts the wrong op order).

**A failed handshake is only reported as an error on Node 22.** The first version of this PR settled `connect()` from the close handler alone. That is enough on Node 24, but Node 22's `WebSocket` fires *only* `error` when the opening handshake fails — no close event ever arrives — so `connect()` and `close()` both waited forever and the integration job ran for six hours before CI cancelled it. The socket is now finished from whichever of the two events arrives first, deduplicated, and a handshake failure is reported as `ShardSocketCloseCodes.Shutdown`: the rejection tells the caller the connection failed, so a close code that reconnects on its own would take that decision out of its hands.

Cleanup for the servers those tests start lives in an `afterEach` rather than at the end of each test, because mocha does not abort the body of a test that times out — a test left awaiting forever would otherwise keep its listener open and the runner would never exit, turning any future regression here into another six hour job instead of a red X.

Overlap: #5292 rewrites `handleMessage`/`decompressPacket` and my `socket.onmessage` catch wrapper sits in that region — expect a textual conflict there.

---

### DIFF
--- a/packages/gateway/src/Shard.ts
+++ b/packages/gateway/src/Shard.ts
@@ -63,12 +63,12 @@ export class DiscordenoShard {
   /** zlib inflate/zstd buffer. */
   inflateBuffer: Uint8Array | null = null;
   /**
-   * A function that will be called once the socket is closed and handleClose() has finished updating internal states.
+   * A function that will be called once the socket is closed and the internal states have been updated.
    *
    * @private
    * This is for internal purposes only, and subject to breaking changes.
    */
-  resolveAfterClose?: (close: CloseEvent) => void;
+  resolveAfterClose?: () => void;
 
   constructor(options: ShardCreateOptions) {
     this.id = options.id;
@@ -111,28 +111,13 @@ export class DiscordenoShard {
     return safeRequests < 0 ? 0 : safeRequests;
   }
 
-  /**
-   * Check whether the shard is allowed to write a payload to the socket right now.
-   *
-   * @param bypassSessionCheck - Whether the payload is part of the handshake itself and as such may be sent before the session is established.
-   */
-  canSend(bypassSessionCheck: boolean = false): boolean {
-    return bypassSessionCheck ? this.isOpen() : this.state === ShardState.Connected;
-  }
+  async checkOffline(highPriority: boolean): Promise<void> {
+    if (this.isOpen()) return;
 
-  async checkOffline(highPriority: boolean, bypassSessionCheck: boolean = false): Promise<void> {
-    if (this.canSend(bypassSessionCheck)) return;
-
-    return await new Promise<void>((resolve, reject) => {
-      // The queue is also drained when the shard goes offline for good, in which case the payload can not be sent anymore.
-      const onDrain = () => {
-        if (this.canSend(bypassSessionCheck)) resolve();
-        else reject(new Error(`[Shard] Shard #${this.id} is offline, the payload has been dropped.`));
-      };
-
+    return await new Promise<void>((resolve) => {
       // Higher priority requests get added at the beginning of the array.
-      if (highPriority) this.offlineSendQueue.unshift(onDrain);
-      else this.offlineSendQueue.push(onDrain);
+      if (highPriority) this.offlineSendQueue.unshift(resolve);
+      else this.offlineSendQueue.push(resolve);
     });
   }
 
@@ -140,24 +125,24 @@ export class DiscordenoShard {
   async close(code: number, reason: string): Promise<void> {
     this.logger.debug(`[Shard] Request for Shard #${this.id} to close the socket with code ${code}.`);
 
-    const socket = this.socket;
-
-    // A socket which is still connecting has to be closed as well, otherwise the shard would finish the handshake and identify itself after it has
-    // been asked to close. Closing it aborts the handshake, which is reported with a close code of its own.
-    if (!socket || (socket.readyState !== WebSocket.OPEN && socket.readyState !== WebSocket.CONNECTING)) {
-      this.logger.debug(`[Shard] Shard #${this.id}'s ready state is ${socket?.readyState}, Unable to close.`);
+    // A socket which is still connecting has to be closed as well, otherwise the shard finishes the handshake and identifies itself after it has
+    // been asked to close.
+    if (this.socket?.readyState !== WebSocket.OPEN && this.socket?.readyState !== WebSocket.CONNECTING) {
+      this.logger.debug(`[Shard] Shard #${this.id}'s ready state is ${this.socket?.readyState}, Unable to close.`);
       return;
     }
 
+    // Closing a socket which is still connecting aborts the handshake, which is reported with a close code chosen by the runtime rather than the
+    // one asked for here, so the closure is marked as intentional to keep handleClose() from taking it for an unexpected one and reconnecting.
     this.goingOffline =
-      socket.readyState === WebSocket.CONNECTING || code === GatewayCloseEventCodes.NormalClosure || code === GatewayCloseEventCodes.GoingAway;
+      this.socket.readyState === WebSocket.CONNECTING || code === GatewayCloseEventCodes.NormalClosure || code === GatewayCloseEventCodes.GoingAway;
 
     // This has to be created before the actual call to socket.close as for example Bun calls socket.onclose immediately on the .close() call instead of waiting for the connection to end
-    const promise = new Promise((resolve) => {
+    const promise = new Promise<void>((resolve) => {
       this.resolveAfterClose = resolve;
     });
 
-    socket.close(code, reason);
+    this.socket.close(code, reason);
 
     this.logger.debug(`[Shard] Waiting for Shard #${this.id} to close the socket with code ${code}.`);
 
@@ -235,46 +220,42 @@ export class DiscordenoShard {
     // By default WebSocket will give us a Blob, this changes it so that it gives us an ArrayBuffer
     socket.binaryType = 'arraybuffer';
 
-    // The handlers below are called by the socket, so nothing is able to catch the errors of the promises they return.
-    socket.onmessage = (messageEvent) => {
-      this.handleMessage(messageEvent).catch((error) => this.logger.error(`[Shard] Shard #${this.id} failed to handle a message.`, error));
-    };
+    socket.onmessage = (messageEvent) => this.handleMessage(messageEvent);
 
     return await new Promise((resolve, reject) => {
-      // The socket may be closed before it ever opened, for example when the handshake fails or when the shard is closed while it is connecting.
-      // In those cases `onopen` is never called, so the connection request has to be settled from the close handler instead.
-      let socketOpened = false;
-      let socketFinished = false;
-
-      // A socket is only finished once, but the runtimes disagree on how a failed opening handshake is reported: some send a close event, some
-      // report it more than once, and the WebSocket of Node 22 only ever fires the error. Whichever arrives first ends the socket here, so a
-      // handshake that never completes cannot leave `connect()` and `close()` both waiting on an event which is not coming.
-      const finish = (closeEvent: CloseEvent) => {
-        if (socketFinished) return;
-        socketFinished = true;
-
-        if (!socketOpened) {
-          reject(new Error(`[Shard] Shard #${this.id} closed the socket with code ${closeEvent.code} before it was connected.`));
-        }
-
-        this.handleClose(closeEvent).catch((error) => this.logger.error(`[Shard] Shard #${this.id} failed to handle the socket closure.`, error));
-      };
+      // Whether the socket reached the open state, and whether the shard has already been torn down for it. A handshake which never completes is
+      // not reported the same way everywhere: it can be an error followed by a close event, more than one close event when the handshake is
+      // aborted, or only an error, so both handlers below go through the same guard.
+      let opened = false;
+      let closed = false;
 
       socket.onerror = (event) => {
         this.handleError(event);
 
-        // An error on a socket which did open is followed by a close event, so only a failed handshake is finished from here. It is reported as
-        // a shutdown because the caller is told the connection failed by the rejection above and decides itself whether to try again, which a
-        // close code that reconnects on its own would take out of its hands.
-        if (!socketOpened) {
-          finish({ code: ShardSocketCloseCodes.Shutdown, reason: 'The socket failed to connect.' } as CloseEvent);
-        }
+        // An error on a socket which is open is followed by a close event which takes care of the rest.
+        if (opened) return;
+
+        // onopen is never called for a handshake which failed, so the connection request is settled here. Doing so a second time is a no-op,
+        // which is what happens when the runtime reports the failure as an error and a close event both.
+        reject(new Error(`[Shard] Shard #${this.id} could not connect, the socket errored before the handshake completed.`));
+
+        // Not every runtime fires a close event for a handshake which never completed, so the shard is put offline here when the close handler
+        // has not already done it.
+        if (closed) return;
+        closed = true;
+
+        this.handleFailedHandshake();
       };
 
-      socket.onclose = (closeEvent) => finish(closeEvent);
+      socket.onclose = (closeEvent) => {
+        if (closed) return;
+        closed = true;
+
+        void this.handleClose(closeEvent);
+      };
 
       socket.onopen = () => {
-        socketOpened = true;
+        opened = true;
 
         // Only set the shard to `Unidentified` state if the connection request does not come from an identify or resume action.
         if (![ShardState.Identifying, ShardState.Resuming].includes(this.state)) {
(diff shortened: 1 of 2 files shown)


packages/gateway/src/Shard.ts
```diff
@@ -111,33 +111,53 @@ export class DiscordenoShard {
...
+    // A socket which is still connecting has to be closed as well, otherwise the shard would finish the handshake and identify itself after it has
+    // been asked to close. Closing it aborts the handshake, which is reported with a close code of its own.
+    if (!socket || (socket.readyState !== WebSocket.OPEN && socket.readyState !== WebSocket.CONNECTING)) {
+      this.logger.debug(`[Shard] Shard #${this.id}'s ready state is ${socket?.readyState}, Unable to close.`);
       return;
     }
 
-    this.goingOffline = code === GatewayCloseEventCodes.NormalClosure || code === GatewayCloseEventCodes.GoingAway;
+    this.goingOffline =
+      socket.readyState === WebSocket.CONNECTING || code === GatewayCloseEventCodes.NormalClosure || code === GatewayCloseEventCodes.GoingAway;
```
Fleny113: why.

packages/gateway/src/Shard.ts
```diff
@@ -111,33 +111,53 @@ export class DiscordenoShard {
...
 
     // This has to be created before the actual call to socket.close as for example Bun calls socket.onclose immediately on the .close() call instead of waiting for the connection to end
     const promise = new Promise((resolve) => {
       this.resolveAfterClose = resolve;
     });
 
-    this.socket.close(code, reason);
+    socket.close(code, reason);
```
Fleny113: why does this need to be aliased?

packages/gateway/src/Shard.ts
```diff
@@ -216,11 +236,32 @@ export class DiscordenoShard {
...
+    // The handlers below are called by the socket, so nothing is able to catch the errors of the promises they return.
+    socket.onmessage = (messageEvent) => {
+      this.handleMessage(messageEvent).catch((error) => this.logger.error(`[Shard] Shard #${this.id} failed to handle a message.`, error));
+    };
+
+    return await new Promise((resolve, reject) => {
+      // The socket may be closed before it ever opened, for example when the handshake fails or when the shard is closed while it is connecting.
+      // In those cases `onopen` is never called, so the connection request has to be settled from the close handler instead.
+      let socketOpened = false;
+      let socketClosed = false;
```
Fleny113: Even if the WS hasn't opened, close will be called, as well as error since it failed to open the request

NotDemonix:
The Integration Tests job on the previous head sat for six hours and was cancelled. It never produced a line of output — not even turbo's build logs — so it hung before the suite ran rather than in a test; the suite passes locally in under a second and mocha exits cleanly.

I can't re-run jobs on this repo, so I've pushed to retrigger it. The one change is to the first lifecycle test, which dialed port 1 and relied on the connection being refused; whether an unused port is refused or silently dropped depends on the machine, so it now uses a local server that drops the connection during the handshake. Deterministic everywhere, and it closes the server in a `finally`.

packages/gateway/src/Shard.ts
```diff
@@ -216,11 +236,32 @@ export class DiscordenoShard {
...
+
+      socket.onclose = (closeEvent) => {
+        // A socket can only be closed once, however aborting a handshake is reported more than once by some WebSocket implementations.
```
Fleny113: may i ask what implementation we are talking about?

packages/gateway/src/Shard.ts
```diff
@@ -216,11 +236,32 @@ export class DiscordenoShard {
...
+        if (socketClosed) return;
+        socketClosed = true;
+
+        if (!socketOpened) {
+          reject(new Error(`[Shard] Shard #${this.id} closed the socket with code ${closeEvent.code} before it was connected.`));
```
Fleny113: it will fire error, wiring that to the reject is fine, hjacking other events is not

packages/gateway/src/Shard.ts
```diff
@@ -350,23 +393,40 @@ export class DiscordenoShard {
...
+  drainOfflineSendQueue(): void {
+    if (this.offlineSendQueue.length === 0) return;
+
+    this.logger.debug(`[Shard] Dropping ${this.offlineSendQueue.length} queued payloads of Shard #${this.id} as it went offline.`);
+
+    for (const resolve of this.offlineSendQueue) resolve();
+    // Setting the length to 0 will delete the elements in it
+    this.offlineSendQueue.length = 0;
+  }
+
```
Fleny113: what

packages/gateway/src/Shard.ts
```diff
@@ -388,17 +448,37 @@ export class DiscordenoShard {
...
+    // The shard has been closed on purpose, so it stays disconnected no matter which close code the socket reported.
+    if (this.goingOffline) {
+      this.state = ShardState.Disconnected;
+      this.events.disconnected?.(this);
+
+      this.goingOffline = false;
+      this.drainOfflineSendQueue();
+
+      return;
+    }
```
Fleny113: see the "why." on where goingOffline is assigned: why.

packages/gateway/src/Shard.ts
```diff
@@ -416,7 +496,14 @@ export class DiscordenoShard {
...
 
-        throw new Error(close.reason || 'Discord gave no reason! GG! You broke Discord!');
+        this.drainOfflineSendQueue();
+
+        // This runs in the socket's close handler, throwing here would result in an uncaught exception which no caller is able to handle.
+        this.logger.fatal(
+          `[Shard] Shard #${this.id} closed with the unrecoverable code ${close.code}: ${close.reason || 'Discord gave no reason! GG! You broke Discord!'}`,
+        );
+
+        return;
```
Fleny113: what

packages/gateway/src/Shard.ts
```diff
@@ -429,21 +516,10 @@ export class DiscordenoShard {
...
-          this.events.disconnected?.(this);
-
-          this.goingOffline = false;
-
-          return;
-        }
-
-        // Otherwise, we want the shard to go through the default case where it gets resumed, as it might be an unexpected closure from Discord or Cloudflare for example, so we don't use break / return here.
-      }
+      case GatewayCloseEventCodes.GoingAway:
```
Fleny113: what.

packages/gateway/tests/integration/lifecycle.spec.ts
```diff
@@ -0,0 +1,179 @@
...
+    resolve = _resolve;
+    reject = _reject;
+  });
+
+  return {
+    promise,
+    resolve,
+    reject,
+  };
+}
```
Fleny113: do not duplicate code that already exists
(since this is done by an AI) we do not need to do "isRecord" but with Promise.withResolvers

packages/gateway/tests/integration/lifecycle.spec.ts
```diff
@@ -0,0 +1,179 @@
...
+      totalShards: 1,
+      url,
+      version: 10,
+    },
+    events: {},
+    logger: { debug: () => {}, info: () => {}, warn: () => {}, error: () => {}, fatal: () => {} },
+  });
+}
+
+function createGatewayManagerWithPort(port: number): GatewayManager {
```
Fleny113: why is this different from the other one

packages/gateway/tests/integration/lifecycle.spec.ts
```diff
@@ -0,0 +1,179 @@
...
+      transportCompression: null,
+      intents: Intents.Guilds,
+      properties: { os: 'linux', browser: 'Discordeno', device: 'Discordeno' },
+      token: '',
+      totalShards: 1,
+      url,
+      version: 10,
+    },
+    events: {},
+    logger: { debug: () => {}, info: () => {}, warn: () => {}, error: () => {}, fatal: () => {} },
```
Fleny113: why are we nulling the logger?

Fleny113 (changes requested):
what.

i am genuinely confused on what the * is going on for most of the code edited

NotDemonix:
Found and fixed the cause of the six hour job. It was a real bug in this PR, not CI.

Node 22's `WebSocket` fires **only** `error` when the opening handshake fails — no close event follows. Settling `connect()` from the close handler alone therefore worked on my Node 24 machine and never settled on CI's Node 22, so `connect()` and `close()` both waited forever. The socket is now finished from whichever of `error`/`close` arrives first.

Two of the tests here were sitting on exactly that, which is why the runner never exited. Their server cleanup has also moved into an `afterEach`, since mocha does not abort the body of a test that times out — so a regression here now fails in seconds instead of hanging the job.

Integration Tests is green on Node 22. The red X on Unit Tests is the codecov upload step; `yarn test:unit` passed in that same job.

packages/gateway/src/Shard.ts
NotDemonix: Because closing a socket that is still connecting aborts the handshake, and the close code that comes back is the runtime's, not the one passed in here.

Measured on Node 24, `close(4000, ...)` on a `CONNECTING` socket fires `close(0)`, `error`, `close(1002)` — the code the shard asked for never appears. Without the flag, `handleClose` sees 1002, falls through to the default case and resumes, so the shard reconnects immediately after being told to shut down.

`goingOffline` is the existing "this closure was on purpose" signal, so that is what I used rather than inventing a second one. The comment now says this instead of leaving you to guess.

packages/gateway/src/Shard.ts
NotDemonix: It doesn't. The alias was only there because I wasn't sure TS would narrow `this.socket` after a compound `readyState` check. It does, so the alias is gone and it's `this.socket.close(code, reason)` again.

packages/gateway/src/Shard.ts
NotDemonix: On most runtimes yes — on the one CI runs, no. Node 22's built-in `WebSocket` fires only `error` when the opening handshake fails; no close event follows. That is what left the Integration Tests job hanging for six hours on the previous head.

For what it's worth, on Node 24 all three cases do fire both (probe output):

```
handshake dropped        : [error, close(1002)]
close() while CONNECTING : [close(0), error, close(1002)]
connection refused       : [error, close(1002)]
```

So the reject is wired to `error`, which every runtime fires, exactly as you asked below.

packages/gateway/src/Shard.ts
NotDemonix: Node's built-in `WebSocket` on Node 22, which is what CI runs and what `connect()` uses. The comment names it explicitly now rather than saying "some implementations".

packages/gateway/src/Shard.ts
NotDemonix: Done. `reject` is wired to `onerror` only, and `onclose` is back to just calling `handleClose` — no fabricated close event anywhere.

The one thing the error handler still does beyond rejecting is put the shard offline through a small `handleFailedHandshake()`, and only when the close handler hasn't already run. That's needed because on Node 22 there is no close event for a failed handshake, so otherwise `this.socket` is never cleared and a `close()` waiting on the socket to go away never returns.

packages/gateway/src/Shard.ts
NotDemonix: Out of scope for this PR — removed. `canSend`, `bypassSessionCheck`, the offline queue rejection and `drainOfflineSendQueue` are all gone, and `checkOffline`/`send` are byte-for-byte what they were. Happy to raise it separately if you want that behaviour.

packages/gateway/src/Shard.ts
NotDemonix: Same reason as the assignment: an aborted handshake is reported with 1002/1006, not `NormalClosure` or `GoingAway`, so the check cannot stay inside that case or it never runs for the case this PR is about. The comment now explains why it sits above the switch.

packages/gateway/src/Shard.ts
NotDemonix: Out of scope — reverted, the `throw` is back exactly as it was.

packages/gateway/src/Shard.ts
NotDemonix: The NOTE is back verbatim. I'd deleted it because the explanation it pointed at lived inside the block body that went away; there's now one added line saying why those two codes fall through, and nothing else changed here.

packages/gateway/tests/integration/lifecycle.spec.ts
NotDemonix: Right, `createGatewayManagerWithPort` and the `promiseWithResolvers` polyfill were straight copies out of `connection.spec.ts`. Both are gone — the two tests that needed them are gone too, so the file is down to `createShard` and a small `listen` helper.

packages/gateway/tests/integration/lifecycle.spec.ts
NotDemonix: Gone with the helper.

packages/gateway/tests/integration/lifecycle.spec.ts
NotDemonix: No good reason, it was only to keep the test output quiet. Removed, the shard uses the default logger now.

NotDemonix:
Rebuilt this one from `main` and re-applied only what the title claims, so the diff is a lot smaller now.

**Kept** — the two actual fixes:

- `connect()` settles when the opening handshake fails. `reject` is wired to `onerror` only, as you asked; `onclose` just calls `handleClose` again, no synthesised close event.
- `close()` acts on a `CONNECTING` socket, and `goingOffline` keeps the aborted handshake from being read as an unexpected closure and resumed.

The only new surface is `handleFailedHandshake()`, which puts the shard offline when the error handler is the only thing that fires. Node 22's built-in `WebSocket` sends no close event for a handshake that never completed, and that is what hung the Integration Tests job for six hours earlier.

**Dropped** — everything you marked with a "what", none of it belongs under this title:

- `canSend` / `bypassSessionCheck` / the offline queue rejection / `drainOfflineSendQueue`
- the `throw` → `logger.fatal` change in the unrecoverable close codes
- the `Shutdown`/`Resharded` case split

Say the word on any of those and I'll open them separately.

**Tests**: chai instead of `node:assert`, the default logger, no copies of `connection.spec.ts`'s helpers, and only the two tests that cover what is left. Teardown moved to `afterEach` so a regression fails in seconds rather than hanging the runner.

PR #5376, opened 2026-08-28
fix(gateway): stop background async work from crashing the process

NotDemonix:
Three floating promises. The first one takes the process down.

**The resharding check can kill the bot.** `spawnShards` registers:

```ts
gateway.resharding.checkIntervalId = setInterval(async () => {
  const reshardingInfo = await gateway.resharding.checkIfReshardingIsNeeded();
  if (reshardingInfo.needed && reshardingInfo.info) await gateway.resharding.reshard(reshardingInfo.info);
}, gateway.resharding.checkInterval);
```

`setInterval` discards the promise and there is no `try`/`catch`. Resharding is on by default (`manager.ts:62`) with an 8h interval, and `createBot` defaults `getSessionInfo` to `bot.rest.getGatewayBot()` (`bot.ts:86-90`) — a live REST call. So eight hours in, one 500 from `GET /gateway/bot` becomes an `unhandledRejection`, which on Node 22 (the repo's engine floor) terminates the process and drops every connected shard. In the test a rejecting `getSessionInfo` took the process down and no error was ever logged. It now logs `[Resharding] The resharding check failed, it will be retried at the next interval.` and keeps running.

**`editBotStatus` sent nothing before resolving.**

```ts
await Promise.all([...gateway.shards.values()].map(async (shard) => {
  gateway.editShardStatus(shard.id, data);   // not returned
}));
```

The inner callback returns `undefined`, so `Promise.all` awaited nothing — `await gateway.editBotStatus(...)` came back with zero presence updates sent (recorded `[]` instead of `[0, 1]`), and any per-shard send failure became an unhandled rejection.

**The zlib write promise is dropped on the partial-frame path.** `decompressPacket` returns early when a frame does not end with the flush marker, abandoning the `writePromise` it just created. A write error then has no handler. Now caught and logged.

Left alone deliberately: a `Z_DATA_ERROR` never invokes the per-write callback at all, so the *awaited* path can still hang. That needs a one-shot `'error'` listener per write — a separate defect with separate risk.

Three tests; all three fail on `main`.

Overlap: #5292 rewrites `decompressPacket`, so the one-line `.catch()` there will likely conflict textually.

---


packages/gateway/src/manager.ts
```diff
@@ -312,10 +312,17 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
           return;
         }
 
-        gateway.resharding.checkIntervalId = setInterval(async () => {
-          const reshardingInfo = await gateway.resharding.checkIfReshardingIsNeeded();
-
-          if (reshardingInfo.needed && reshardingInfo.info) await gateway.resharding.reshard(reshardingInfo.info);
+        gateway.resharding.checkIntervalId = setInterval(() => {
+          // The promise of the callback is not awaited by the timer, so it has to handle its own errors, otherwise a single failed check terminates the process.
+          void (async () => {
```
Fleny113: given neither `setInterval(async () => {}, ...)` and `void (async () => {})` await, there is NO point in an inline function here

packages/gateway/src/Shard.ts
```diff
@@ -494,9 +494,10 @@ export class DiscordenoShard {
         return null;
       }
 
+      // The promise is abandoned when a partial frame returns early, so it needs to handle its own errors.
       const writePromise = new Promise<void>((resolve, reject) => {
         this.inflate!.write(compressedData, 'binary', (error) => (error ? reject(error) : resolve()));
-      });
+      }).catch((error) => this.logger.error('[Shard] There was an error writing to the decompression stream', error));
```
Fleny113: this catch will make it so if the inflate fails to write, the `await` below will not throw, causing it to clear the buffer and decode incomplete/invalid JSON data

packages/gateway/tests/integration/background.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+    });
+    // The shards are irrelevant for this test, so they are not actually identified.
+    gateway.tellWorkerToIdentify = async () => {};
+
+    await gateway.spawnShards();
+    await delay(100);
+
+    clearInterval(gateway.resharding.checkIntervalId);
+
+    assert.ok(errors.some((error) => error.includes('[Resharding]')));
```
Fleny113: why are we using node:assert if EVERY OTHER TEST IN THE CODEBASE uses chai

packages/gateway/tests/integration/background.spec.ts
```diff
@@ -0,0 +1,100 @@
+import assert from 'node:assert/strict';
```
Fleny113: no.

Fleny113 (changes requested):

packages/gateway/src/manager.ts
NotDemonix: You're right, the IIFE bought nothing. It's a plain `setInterval(async () => { ... })` with the try/catch inside now.

packages/gateway/src/Shard.ts
NotDemonix: Good catch, that was a real bug in the fix. The catch is now attached only on the branch where the promise is actually abandoned (the partial-frame early return); the awaited path is untouched, so a failed write still throws before anything decodes the buffer.

packages/gateway/tests/integration/background.spec.ts
NotDemonix: File removed.

packages/gateway/tests/integration/background.spec.ts
NotDemonix: Removed.

PR #5377, opened 2026-08-28
fix(gateway): fix member and soundboard requests, presence and resharding

NotDemonix:
Five fixes in `manager.ts`, each independent.

**`requestSoundboardSounds` throws on a bigint guild id.** It groups `BigString` ids into `Map<number, BigString[]>` and forwards them unconverted, so `JSON.stringify` in `Shard.send` throws `TypeError: Do not know how to serialize a BigInt` — after a bucket token has already been spent. Ids are now stringified when grouped.

**`editShardStatus` discarded `since` and `afk`.** The type forces the caller to supply both, then the payload hardcoded `since: null, afk: false`. Passing `{ since: 1700000000000, afk: true, … }` put `since: null, afk: false` on the wire. Now `data.since ?? null` / `data.afk ?? false`.

**The `GUILD_MEMBERS` guard ran too early and was too broad.** It fired before `userIds` was folded into `limit`:

```ts
if ((!options?.limit || options.limit > 1) && (gateway.intents & GatewayIntents.GuildMembers) === 0) throw …
```

so `requestMembers(guildId, { userIds: [1n], limit: 0 })` was rejected without the privileged intent. Discord only requires `GUILD_MEMBERS` for the **entire** member list (`query: ''`, `limit: 0`) — a `user_ids` or prefix-query lookup does not need it. The guard now fires only when there are no `userIds`, no `query` and no `limit`. A test asserts `requestMembers(guildId)` with no options still rejects. Message changed to `Cannot fetch the entire member list without the GUILD_MEMBERS intent`; grep shows no other reference to the old string.

**`requestMembers` leaked pending entries and stranded callers.** A failed `sendPayload` left the pending entry in place forever (members that will never arrive), and two calls sharing a caller-supplied `nonce` silently overwrote the first entry — the first caller's promise never settled. The send is now wrapped so a failure deletes the entry, and a duplicate nonce rejects with `A member request with the nonce '…' is already pending.` instead of stranding the earlier call.

**`reshard()` used a stale `maxConcurrency`.** It never refreshed `gateway.connection`, so `prepareBuckets` rebuilt identify buckets from the session start limit captured at process start. With a freshly fetched `maxConcurrency` of 16 it still prepared 1 bucket. Now assigns `info.sessionStartLimit` before using it.

Seven tests; six fail on `main` and the seventh is the guard rail for the narrowed intent check.

Not included: the `reshard()` total-shards routing window. It is real — `totalShards`/`firstShardId`/`lastShardId` are reassigned at the top of `reshard()` while `gateway.shards` still holds the old shards, so `calculateShardId` misroutes for the whole preparation window — but every fix changes when the swap becomes visible, and that is a maintainer's design call, not mine.

---

### DIFF
--- a/packages/gateway/src/manager.ts
+++ b/packages/gateway/src/manager.ts
@@ -113,6 +113,8 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
       },
       async reshard(info) {
         gateway.logger.info(`[Resharding] Starting the reshard process. Previous total shards: ${gateway.totalShards}`);
+        // The information which has just been fetched holds the current session start limit, which the total shards and the buckets are based on.
+        gateway.connection.sessionStartLimit = info.sessionStartLimit;
         // Set values on gateway
         gateway.totalShards = info.shards;
         // Handles preparing mid sized bots for LBS
@@ -439,8 +441,8 @@ export function createGatewayManager(options: CreateGatewayManagerOptions): Gate
       await gateway.sendPayload(shardId, {
         op: GatewayOpcodes.PresenceUpdate,
         d: {
-          since: null,
-          afk: false,
+          since: data.since,
+          afk: data.afk,
           activities: data.activities,
           status: data.status,
         },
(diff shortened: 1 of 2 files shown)


NotDemonix:
The red X on **Unit Tests** here is the codecov upload step, not the tests:

```
error -- Report creating failed: {"message":"Token required because branch is protected"}
##[error]Process completed with exit code 1.
```

The suite itself is green in that same job — `Tasks: 12 successful, 12 total`, every package passing. It looks like the usual fork-PR codecov token flake (my other PRs in this batch uploaded fine). I do not have permission to re-run the job, so it needs a maintainer re-run or a push to clear it.

PR #5378, opened 2026-08-28
fix(utils): correct sticker and guild tag badge CDN urls

NotDemonix:
**`stickerUrl()` returns a url the CDN 404s.** It routes through `formatImageUrl`, whose animated check is `url.includes('/a_')` — a sticker path is a bare id, so it never matches and the extension always defaults to `.webp`:

```
stickerUrl('1228092333061443654')
-> https://cdn.discordapp.com/stickers/1228092333061443654.webp?size=128
```

Probing the live CDN with that url:

```
404 application/xml   .../stickers/1228092333061443654.webp?size=128
200 image/png         .../stickers/1228092333061443654.png?size=128
200 application/json  .../stickers/749054660769218631.json
```

[Image Formatting](https://discord.com/developers/docs/reference#image-formatting) documents the sticker path as `stickers/{sticker_id}.png` with formats PNG / Lottie / GIF, GIFs served from `media.discordapp.net`, and the `size` parameter *ignored* for stickers. The extension is now picked from the format type (`png`, `json` for Lottie, `gif` for the media host) instead of the `/a_` heuristic.

**`guildTagBadgeUrl()` printed the hash in base 10.** The badge hash is a bigint and was interpolated directly:

```
before: .../guild-tag-badges/785384884197392384/4034407661299384404326332419647968090039.webp?size=128
after:  .../guild-tag-badges/785384884197392384/db26a6fb924c985f66b79364cf5797b7.webp?size=128
```

Its `undefined` check was also widened to falsy, so an empty hash no longer produces `.../{guildId}/.webp`.

**Worth noting:** the string `stickerUrl()` returns changes shape — different extension, and no `?size=` (Discord ignores it here). Nothing that worked before stops working, since the old `.webp` url 404s, but a snapshot test would see a diff. Might deserve a changeset note.

---

### DIFF
--- a/packages/utils/src/images.ts
+++ b/packages/utils/src/images.ts
@@ -319,16 +319,16 @@ export function stickerPackBannerUrl(bannerAssetId: BigString | undefined, optio
  * @param stickerId - The ID of the sticker to get the icon of
  * @param options - The parameters for the building of the URL.
  * @returns The link to the resource or `undefined`.
+ *
+ * @remarks
+ * Stickers are only available as `png`, `json` (Lottie) and `gif`, and the size parameter is ignored by Discord for stickers.
  */
 export function stickerUrl(stickerId: BigString | number, options?: ImageOptions & { type?: StickerFormatTypes }): string | undefined {
   if (!stickerId) return;
 
-  const url =
-    options?.type === StickerFormatTypes.Gif
-      ? `https://media.discordapp.net/stickers/${stickerId}`
-      : `https://cdn.discordapp.com/stickers/${stickerId}`;
+  if (options?.type === StickerFormatTypes.Gif) return `https://media.discordapp.net/stickers/${stickerId}.${options.format ?? 'gif'}`;
 
-  return formatImageUrl(url, options?.size ?? 128, options?.format);
+  return `https://cdn.discordapp.com/stickers/${stickerId}.${options?.format ?? (options?.type === StickerFormatTypes.Lottie ? 'json' : 'png')}`;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/utils/src/images.ts
```diff
@@ -376,7 +376,11 @@ export function roleIconUrl(roleId: BigString, iconHash: BigString | undefined,
  * @returns The link to the resource or `undefined` if no badge has been set.
  */
 export function guildTagBadgeUrl(guildId: BigString, badgeHash: BigString | undefined, options?: ImageOptions): string | undefined {
-  if (badgeHash === undefined) return undefined;
+  if (!badgeHash) return undefined;
```
Fleny113: why.

packages/utils/src/images.ts
```diff
@@ -319,16 +319,16 @@ export function stickerPackBannerUrl(bannerAssetId: BigString | undefined, optio
...
+ * Stickers are only available as `png`, `json` (Lottie) and `gif`, and the size parameter is ignored by Discord for stickers.
  */
 export function stickerUrl(stickerId: BigString | number, options?: ImageOptions & { type?: StickerFormatTypes }): string | undefined {
   if (!stickerId) return;
 
-  const url =
-    options?.type === StickerFormatTypes.Gif
-      ? `https://media.discordapp.net/stickers/${stickerId}`
-      : `https://cdn.discordapp.com/stickers/${stickerId}`;
+  if (options?.type === StickerFormatTypes.Gif) return `https://media.discordapp.net/stickers/${stickerId}.${options.format ?? 'gif'}`;
```
Fleny113:
```suggestion
  if (options?.type === StickerFormatTypes.Gif) return `https://media.discordapp.net/stickers/${stickerId}.gif`;
```

packages/utils/src/images.ts
```diff
@@ -319,16 +319,16 @@ export function stickerPackBannerUrl(bannerAssetId: BigString | undefined, optio
...
 
-  return formatImageUrl(url, options?.size ?? 128, options?.format);
+  return `https://cdn.discordapp.com/stickers/${stickerId}.${options?.format ?? (options?.type === StickerFormatTypes.Lottie ? 'json' : 'png')}`;
```
Fleny113: keep formatImageUrl, specify the format if not provided and size it not used so it's not an issue if we keep it

Fleny113 (changes requested):

packages/utils/src/images.ts
NotDemonix: No good reason beyond matching the truthiness check the sibling functions use. Reverted to `badgeHash === undefined`; the bigint-to-hash conversion is the only change left in this function.

packages/utils/src/images.ts
NotDemonix: Applied.

packages/utils/src/images.ts
NotDemonix: Done, `formatImageUrl` is back with the format passed in when it isn't given (`json` for Lottie, `png` otherwise). Sticker urls therefore carry `?size=128` again, which the tests now expect.

PR #5379, opened 2026-08-28
fix(utils): allow updates at maxSize and keep sweeper options intact

NotDemonix:
Three `Collection` bugs, each with a test that fails on `main`.

**`set()` refuses to update a key it already holds once at `maxSize`.** The guard only looked at `size`:

```ts
new Collection([['foo', 'bar'], ['me', 'you']], { maxSize: 2 }).set('foo', 'baz')
// get('foo') === 'bar'
```

An update cannot grow the collection, so the cap is still honoured — it just should not block writes to existing keys. The value is silently dropped today, with no error.

**`changeSweeperInterval` / `changeSweeperFilter` drop `bot`.** They rebuild the options object from `interval` and `filter` only, so the filter's third argument became `undefined` on the next sweep — throwing `Cannot read properties of undefined (reading 'cache')` out of `setInterval`, i.e. as an uncaught exception.

**`startSweeper` stores the caller's options object by reference and mutates it.** Two Collections given one config shared a single `intervalId` field, so the second `startSweeper` overwrote the first's timer id. After both `stopSweeper()` calls the first collection's timer was still running and still sweeping. It now stores a copy.

**Behaviour changes:** a Collection at `maxSize` now accepts `set()` on an existing key; `collection.sweeper` is a copy rather than the caller's object, so mutating your own options object to reconfigure a live sweeper no longer has an effect (undocumented, and it never worked reliably). discordeno itself never constructs a `maxSize` Collection or a sweeper, so nothing in-repo is affected.

---

### DIFF
--- a/packages/utils/src/Collection.ts
+++ b/packages/utils/src/Collection.ts
@@ -22,10 +22,12 @@ export class Collection<K, V> extends Map<K, V> {
   startSweeper(options: CollectionSweeper<K, V>): NodeJS.Timeout {
     if (this.sweeper?.intervalId) clearInterval(this.sweeper.intervalId);
 
-    this.sweeper = options;
+    const sweeper = { ...options };
+
+    this.sweeper = sweeper;
     this.sweeper.intervalId = setInterval(() => {
       this.forEach((value, key) => {
-        if (!this.sweeper?.filter(value, key, options.bot)) return;
+        if (!this.sweeper?.filter(value, key, sweeper.bot)) return;
 
         this.delete(key);
         return key;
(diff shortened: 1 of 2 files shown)


packages/utils/src/Collection.ts
```diff
@@ -36,25 +38,27 @@ export class Collection<K, V> extends Map<K, V> {
   }
 
   stopSweeper(): void {
-    return clearInterval(this.sweeper?.intervalId);
+    clearInterval(this.sweeper?.intervalId);
+
+    if (this.sweeper) this.sweeper.intervalId = undefined;
```
Fleny113: guarding this function with if(!this.sweeper) it clener

Fleny113 (changes requested):

packages/utils/src/Collection.ts
NotDemonix: Done, it's an early return now.

PR #5380, opened 2026-08-28
fix(utils): fix leaky bucket spin, embed field limit and urlToBase64

NotDemonix:
Three unrelated fixes in `utils`, each with a test.

**`LeakyBucket.processQueue()` spins the event loop when `max <= 0`.** With `max: 0`, `remaining` is 0 and neither branch inside `while (this.queue.length)` applies — so the loop never awaits and never drains. `acquire()` wedges the process; the repro exited 124 under `timeout 15`. `max` is now clamped to at least 1, with an `else` branch as a safety net.

> The regression test cannot hang CI: mocha's timeout runs on sinon's fake clock and would never fire during a synchronous spin. Instead of calling `acquire()`, the test swaps `bucket.queue` for a Proxy that throws once `length` has been read 100 times — before the fix that surfaces as a rejected `processQueue()` promise (bounded, instant); after the fix `length` is read once and the loop awaits.

**`EmbedsBuilder.validate()` checked field values against 4096.** Discord's limit is 1024, and the message the builder itself throws already said 1024. A 2000-character field value passed validation and the API then rejected the message with `embeds.0.fields.0.value: Must be 1024 or fewer in length`.

**`urlToBase64` encoded error responses as image data.** There was no `res.ok` check, so a 404 came back as `data:image/png;base64,eyJtZXNzYWdlIjoiNDA0…` — a base64-encoded JSON error document presented as an image. The mime type was also derived by slicing the url, so a url of the exact shape `images.ts` produces yielded the malformed label `data:image/webp?size=128;base64,…`. It now uses the response's own `Content-Type` and **rejects** on a non-2xx.

That last one is a deliberate behaviour change: callers that were silently passing a bad data uri to Discord will now see an exception. Note `rest` no longer calls this helper — upstream 8a66a76b2 removed those call sites and the docs now tell users to call it themselves — so this is squarely a public-helper fix.

---

### DIFF
--- a/packages/utils/src/bucket.ts
+++ b/packages/utils/src/bucket.ts
@@ -96,11 +96,6 @@ export class LeakyBucket implements LeakyBucketOptions {
           await delay(1000);
         }
       }
-
-      // Nothing can be processed and no refill is scheduled, wait for the next interval instead of spinning.
-      else {
-        await delay(this.refillInterval);
-      }
     }
 
     // Loop has ended mark false so it can restart later when needed
--- a/packages/utils/src/bucket.ts
+++ b/packages/utils/src/bucket.ts
@@ -20,7 +20,7 @@ export class LeakyBucket implements LeakyBucketOptions {
   logger: Pick<typeof logger, 'debug' | 'info' | 'warn' | 'error' | 'fatal'>;
 
   constructor(options?: LeakyBucketOptions) {
-    this.max = options?.max ?? 1;
+    this.max = Math.max(1, options?.max ?? 1);
     this.refillAmount = options?.refillAmount ? (options.refillAmount > this.max ? this.max : options.refillAmount) : 1;
     this.refillInterval = options?.refillInterval ?? 5000;
     this.logger = options?.logger ?? logger;
@@ -95,6 +95,11 @@ export class LeakyBucket implements LeakyBucketOptions {
           this.logger.debug(`[LeakyBucket] Delaying execution of leaky bucket requests for 1000ms`);
           await delay(1000);
         }
+      }
+
+      // Nothing can be processed and no refill is scheduled, wait for the next interval instead of spinning.
+      else {
+        await delay(this.refillInterval);
       }
     }
 
--- a/packages/utils/src/urlToBase64.ts
+++ b/packages/utils/src/urlToBase64.ts
@@ -7,9 +7,7 @@ export async function urlToBase64(url: string): Promise<string> {
   if (!response.ok) throw new Error(`Failed to fetch ${url}: ${response.status} ${response.statusText}`);
 
   const imageStr = encode(await response.arrayBuffer());
-  const contentType = response.headers.get('content-type')?.split(';')[0];
-  const pathname = new URL(url).pathname;
-  const type = pathname.includes('.') ? pathname.substring(pathname.lastIndexOf('.') + 1) : 'png';
+  const type = url.substring(url.lastIndexOf('.') + 1);
 
-  return `data:${contentType || `image/${type}`};base64,${imageStr}`;
+  return `data:image/${type};base64,${imageStr}`;
 }
--- a/packages/utils/src/urlToBase64.ts
+++ b/packages/utils/src/urlToBase64.ts
@@ -2,8 +2,14 @@ import { encode } from './base64.js';
 
 /** Converts a url to base 64. Useful for example, uploading/creating server emojis. */
 export async function urlToBase64(url: string): Promise<string> {
-  const buffer = await fetch(url).then(async (res) => await res.arrayBuffer());
-  const imageStr = encode(buffer);
-  const type = url.substring(url.lastIndexOf('.') + 1);
-  return `data:image/${type};base64,${imageStr}`;
+  const response = await fetch(url);
+
+  if (!response.ok) throw new Error(`Failed to fetch ${url}: ${response.status} ${response.statusText}`);
+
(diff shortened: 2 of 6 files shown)


packages/utils/src/urlToBase64.ts
```diff
@@ -2,8 +2,14 @@ import { encode } from './base64.js';
...
+  const response = await fetch(url);
+
+  if (!response.ok) throw new Error(`Failed to fetch ${url}: ${response.status} ${response.statusText}`);
+
+  const imageStr = encode(await response.arrayBuffer());
+  const contentType = response.headers.get('content-type')?.split(';')[0];
+  const pathname = new URL(url).pathname;
+  const type = pathname.includes('.') ? pathname.substring(pathname.lastIndexOf('.') + 1) : 'png';
+
+  return `data:${contentType || `image/${type}`};base64,${imageStr}`;
```
Fleny113: I would rather not change the behavior of this function as it uses the provided path and not the content-type

packages/utils/src/bucket.ts
```diff
@@ -96,6 +96,11 @@ export class LeakyBucket implements LeakyBucketOptions {
           await delay(1000);
         }
       }
+
+      // Nothing can be processed and no refill is scheduled, wait for the next interval instead of spinning.
+      else {
+        await delay(this.refillInterval);
+      }
```
Fleny113: I'm not sure we want this. This should be an impossible case, as `remaining` is falsy (0) AND `refillsAt` is falsy (undefined)

as for this condition to happen it would need to:
- Never had a remaining at all (covered by the fix above with Math.max), or
- Someone messed up with the internals, or
- `refillsAt` is undefined where it shouldn't

packages/utils/tests/bucket.spec.ts
```diff
@@ -183,6 +183,39 @@ describe('bucket.ts', () => {
...
+      let error: unknown;
+      void bucket.processQueue().catch((err) => {
+        error = err;
+      });
+
+      await Promise.resolve();
+
+      expect(error).to.equal(undefined);
+      expect(reads).to.be.lessThan(100);
+    });
```
Fleny113: remove this given editing the bucket internals is something that should be only done when knowing what you are doing and it's a skill issue if you spin-lock yourself

packages/utils/tests/urlToBase64.spec.ts
```diff
@@ -19,12 +19,47 @@ describe('urlToBase64.ts', () => {
       const mockArrayBuffer = new ArrayBuffer(8);
 
       fetchStub.resolves({
+        ok: true,
+        headers: new Headers({ 'content-type': 'image/png' }),
         arrayBuffer: () => Promise.resolve(mockArrayBuffer),
       });
```
Fleny113: See #5318

packages/utils/tests/urlToBase64.spec.ts
```diff
@@ -19,12 +19,47 @@ describe('urlToBase64.ts', () => {
...
     });
+
+    it('Will throw on a non ok response', async () => {
+      fetchStub.resolves({
+        ok: false,
+        status: 404,
+        statusText: 'Not Found',
+        headers: new Headers({ 'content-type': 'application/json' }),
+        arrayBuffer: () => Promise.resolve(new ArrayBuffer(8)),
+      });
```
Fleny113: See #5318: use a Response object, do not fake it

packages/utils/tests/urlToBase64.spec.ts
```diff
@@ -19,12 +19,47 @@ describe('urlToBase64.ts', () => {
...
+      fetchStub.resolves({
+        ok: true,
+        headers: new Headers(),
+        arrayBuffer: () => Promise.resolve(new ArrayBuffer(8)),
+      });
+
+      const url = await urlToBase64('https://cdn.discordapp.com/avatars/785384884197392384/1234.webp?size=128');
+
+      expect(url).equal('data:image/webp;base64,AAAAAAAAAAA=');
+    });
```
Fleny113: given i don't think we want to change the behavior of this function this doesn't fit

Fleny113 (changes requested):

packages/utils/src/urlToBase64.ts
NotDemonix: Reverted. The type comes from the provided path again, exactly as before; the only thing left in the function is the `!response.ok` throw.

packages/utils/src/bucket.ts
NotDemonix: Fair, dropped. The `Math.max(1, ...)` in the constructor is the only change left, which covers the one reachable way in.

packages/utils/tests/bucket.spec.ts
NotDemonix: Removed.

packages/utils/tests/urlToBase64.spec.ts
NotDemonix: Now a real `Response`.

packages/utils/tests/urlToBase64.spec.ts
NotDemonix: Same here, `new Response(null, { status: 404, statusText: 'Not Found' })`.

packages/utils/tests/urlToBase64.spec.ts
NotDemonix: Removed along with the behaviour change it was covering.

